PR #556, opened 2020-07-09
Remove ExpectedSystemExit rule

stefanbirkner:
This rule is no used since 8b0e061740702453ad232d7f892b133054a32a4c.


bamthomas:
thank you @stefanbirkner !!

PR #961, opened 2022-11-22
Revert "Feat/upgrade es client"

mvanzalu:
Reverts ICIJ/datashare#959


mvanzalu:
Opensearch is compatible with version > 7.10

PR #991, opened 2023-01-18
fix: embedded mode redis error

mvanzalu:
Following the merge of this [PR](https://github.com/ICIJ/datashare/pull/989) a bug has been introduced, when launching Datashare in Embedded mode, it tries to connect to Redis (through the TaskManager), even if it's not specified, which is resulting to an error.

In this fix we moved the creation of TaskManager in CommonMode in order to solve this.

**Important Note** : for `busType` `queueType` `batchQueueType`, we set the default values to MEMORY so we have to take this into account in the deployment scripts.
`


datashare-app/src/main/java/org/icij/datashare/mode/CommonMode.java
```diff
@@ -210,6 +221,10 @@ Routes addExtensionConfiguration(Routes routes) {
         return routes;
     }
 
+    protected boolean hasRedisProperty() {
+        return propertiesProvider.getProperties().contains(QueueType.REDIS.name());
```
pirhoo, line 225: What's the value of `QueueType.REDIS.name()` ?

pirhoo (approved):

Issue #993, opened 2023-01-19
bug: Display all projects in detailed batch search view


caro3801:
**Describe the bug**
Multi-project batch search details show only the first of all the projects in detailed view of the batch search

**Expected behavior**
The detailed view should show the all the projects

PR #994, opened 2023-01-20
fix: display project list in detailed view

caro3801:
Related to #993 

The batchSearch without queries: 
before fix: based on a single project
after fix: get the BS projects from the BS and merge projects into a comma separated string of projects

The batchSearch with queries: not changed because they are already merging the projects into a comma separated string of projects with the "mergeBatchSearches". Eg with 2 projects (p1,p2) and 2 queries (q1, q2), the merge will create 4 rows p1q1, p1q2,p2q1, p2q2 and then retrieve all the different project values from these rows.


pirhoo (approved):
Looks good to me but I wouldn't mind a second-review from someone with more experience in Java! @bthomas @mvanzalu

mvanzalu (approved):

Issue #1012, opened 2023-02-08
Broken tree on Windows


mvanzalu:
**Describe the bug**

In the indexation page, the folder tree isn't properly displayed.
![TreeBugWin](https://user-images.githubusercontent.com/35509100/217496229-f21a4dee-cd90-49ca-bc3f-db32dfaadb0e.png)
As a result, it's impossible to go back to the parent folder.
Update 14/02, also broken in Batch Search Form and any other path explorer
![BSform Tree](https://user-images.githubusercontent.com/35509100/218734366-0a928ea9-c4c0-4ed7-bfe6-84e71552d93f.png)

**To Reproduce**
1. Go to 'Analyze your documents' in "Tasks" tab
2. Click on 'Add Document'
3. Delve into the folder tree
4. It's impossible to go back up the tree

**Expected behavior**
We are supposed to have a tree like this.
![TreebugWinExpected](https://user-images.githubusercontent.com/35509100/217498817-0c5e4461-29cb-492a-81b7-b1b41516d492.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome
 - Version: 11.0.7

Soliine:
To do: size and numbers of documents

PR #1021, opened 2023-02-15
feat: adds path separator in settings endpoint PATH #1012

mvanzalu:


datashare-app/src/main/java/org/icij/datashare/web/RootResource.java
```diff
@@ -72,7 +74,9 @@ public String getRoot(Context context) throws IOException {
      */
     @Get("settings")
     public Map<String, Object> getPublicSettings() {
-        return propertiesProvider.getFilteredProperties(".*Address.*", ".*Secret.*", ".*Url.*", ".*Key.*");
+        Map<String, Object> filteredProperties = propertiesProvider.getFilteredProperties(".*Address.*", ".*Secret.*", ".*Url.*", ".*Key.*");
+        filteredProperties.put("pathSeparator", File.separator);
```
mvanzalu, line 78: Just to be sure, should it belong to the properties or it ok if it's just computed ?

pirhoo: Better if it's computed!

pirhoo (approved):

PR #1066, opened 2023-05-17
fix: add missing nested fields for content_translated

pirhoo:
This PR use a different index settings to search into nested attributes of the content_translated field when doing a search without specifying the field.

Documentation about flattened field names: https://www.elastic.co/guide/en/elasticsearch/reference/current/nested.html


marieglr (approved):
Demo was very clear this morning :+1:

Issue #1092, opened 2023-06-13
feat: make BatchDownload folder configurable


mvanzalu:
**Is your feature request related to a problem? Please describe.**
As it is it's not possible to configure the download directory for the Batch Downloads, we would like to specify it with `--batchDownloadDir` argument

PR #1096, opened 2023-06-20
feat: make batch download folder configurable

mvanzalu:
Related to issue #1092


datashare-app/src/main/java/org/icij/datashare/tasks/BatchDownloadLoop.java
```diff
@@ -19,7 +19,8 @@
 import static org.icij.datashare.tasks.BatchDownloadRunner.NULL_BATCH_DOWNLOAD;
 
 public class BatchDownloadLoop {
-    private final Path DOWNLOAD_DIR = Paths.get(System.getProperty("user.dir")).resolve("app/tmp");
+    private final String DEFAULT_DIR = Paths.get(System.getProperty("user.dir")).resolve("app/tmp").toString();
```
bamthomas, line 22: why String instead of Path?

mvanzalu: To be compatible with `propertiesProvider.get` which is an `Optional<String>` used in the constructor below

Issue #1106, opened 2023-07-03
bug: datashare fails to send mail when smtp URL contains user/password parameters


bamthomas:
**Describe the bug**

When using a url of the form `smtp://user:password@host` for `--batchDownloadEncrypt` datashare is not sending the password mail.

**To Reproduce**
1. Launch datashare with the parameter --batchDownloadEncrypt that contains a user/pass url
2. Create a batch download (download button after a search)
3. See the logs 

```
2023-07-03 13:55:39,608 [pool-2-thread-1] ERROR MailSender - Failed to send mail : hostmail=sandbox.smtp.mailtrap.io, port=2525, Exception=530 5.7.1 Authentication required                                       
                                                                                                                                                                                                                   
2023-07-03 13:55:39,610 [pool-2-thread-1] ERROR BatchDownloadRunner - failed to send mail password                                                                                                                 
org.icij.datashare.com.mail.MailException: com.sun.mail.smtp.SMTPSendFailedException: 530 5.7.1 Authentication required                                                                                            
                                                                                                                                                                                                                   
        at org.icij.datashare.com.mail.MailSender.send(MailSender.java:44)                                                                                                                                         
        at org.icij.datashare.tasks.BatchDownloadRunner$ZipperWithPassword.close(BatchDownloadRunner.java:228)                                                                                                     
        at org.icij.datashare.tasks.BatchDownloadRunner.call(BatchDownloadRunner.java:119)                                                                                                                         
        at org.icij.datashare.tasks.BatchDownloadRunner.call(BatchDownloadRunner.java:49)                                                                                                                          
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)                                                                                                                                      
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)                                                                                                                       
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)                                                                                                                                      
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)                                                                                                               
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)                                                                                                               
        at java.base/java.lang.Thread.run(Thread.java:829)                                                                                                                                                         
Caused by: com.sun.mail.smtp.SMTPSendFailedException: 530 5.7.1 Authentication required                                                                                                                            
                                                                                                                                                                                                                   
[...]

PR #1114, opened 2023-07-03
feat: add user/pass support for MailSender  #1106

bamthomas:
See the #1106 issue

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/com/mail/MailSender.java
+++ b/datashare-app/src/main/java/org/icij/datashare/com/mail/MailSender.java
@@ -1,6 +1,7 @@
 package org.icij.datashare.com.mail;
 
 import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.Properties;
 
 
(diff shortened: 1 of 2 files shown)


pirhoo (approved):

datashare-app/src/main/java/org/icij/datashare/com/mail/MailSender.java
```diff
@@ -23,15 +21,29 @@ public class MailSender {
...
 
     public MailSender(String host, int port) {
+        this(host, port, null, null);
+    }
+
+    public MailSender(URI uri) {
+        this(
+            uri.getHost(),
+            uri.getPort(),
+            uri.getUserInfo().split(":")[0],
```
caro3801, line 36: Do we need to check if it's empty or it will never happen or it is checked by the lib? (empty username or password)

datashare-app/src/main/java/org/icij/datashare/com/mail/MailSender.java
```diff
@@ -68,11 +80,16 @@ private Message createMessage(Mail donneesEmail) throws MessagingException, Addr
         return message;
     }
 
-    private synchronized static Session getMailSession(String hostTransportMail, int portTransportMail) {
+    private synchronized static Session getMailSession(String hostTransportMail, int portTransportMail, String user, String password) {
```
caro3801, line 83: I did not know about the `synchronized` keyword. What is async here?

caro3801 (approved):

datashare-app/src/main/java/org/icij/datashare/com/mail/MailSender.java
bamthomas: it is in case it is used in a multithreaded context : this makes it thread safe.

Issue #1071, opened 2023-05-31
feat: exclude or include tagged documents in Batch Search results


Soliine:
**Is your feature request related to a problem? Please describe.**
Miguel and Delphine asked to be able to exclude or include tagged documents in the batch search form. One can do it manually in the CSV at the moment but it requires to add it in each query of the CSV. 

**Describe the solution you'd like**
<img width="400" alt="Screenshot 2023-05-31 at 13 53 14" src="https://github.com/ICIJ/datashare/assets/17233829/e7bf3810-6771-4512-bede-663e92256deb">
I suggest that we add in the new batch search form, in "Advanced filters", two lines, on the same model as "File types":
- Only run on document(s) tagged with:
- Do not run on document(s) tagged with:
Autocompletion of the tags of the projects selected above if possible.

PR #1119, opened 2023-07-07
feat: adds batch search tag filter

mvanzalu:
This PR is related to [this issue](https://github.com/ICIJ/datashare/issues/1071).

This PR also modifies the liquibase schema by adding a `tags` column into `batch_search` table


pirhoo (approved):
Interesting! 

I was convinced we were saving the ElasticSearch query body instead. This might be a nice improve in the future (so we can add much more filters to the BS).

mvanzalu:
I'm closing this PR, we will go with the query body solution.

Issue #1093, opened 2023-06-16
bug: get BatchSearchId returns 500 when it does not exist


caro3801:
**Describe the bug**
# BatchSearch 

when querying a batch search that does not exists datashare return a 500:
```bash
2023-06-16 08:09:07,160 [Worker: RequestDispatcher: Thread-44] ERROR Fluent - Unexpected error:       │
java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0                               │
        at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:64)               │
        at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:70)     │
        at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:248)               │
        at java.base/java.util.Objects.checkIndex(Objects.java:372)                                   │
        at java.base/java.util.ArrayList.get(ArrayList.java:459)                                      │
        at org.icij.datashare.db.JooqBatchSearchRepository.get(JooqBatchSearchRepository.java:202) 
```

This issue comes from `JooqBatchSearchRepository.java` in [the two methods `get`](https://github.com/ICIJ/datashare/blob/master/datashare-db/src/main/java/org/icij/datashare/db/JooqBatchSearchRepository.java#L186-L203) when the `get(0)` is called but does not exists. 

```java
vient du  :
@Override
    public BatchSearch get(User user, String batchId, boolean withQueries) {
        if (withQueries){
            return get(user, batchId);
        }
        return createBatchSearchWithoutQueriesSelectStatement(DSL.using(dataSource, dialect)).
                where(BATCH_SEARCH.UUID.eq(batchId)).groupBy(BATCH_SEARCH.UUID).having(count().eq(
                        selectCount().from(BATCH_SEARCH_PROJECT)
                                .where(BATCH_SEARCH_PROJECT.SEARCH_UUID.eq(BATCH_SEARCH.UUID))))
                .fetch().stream().map(this::createBatchSearchWithoutQueries).collect(toList()).get(0);
    }
```
Expected:
The backend should handle this request by returning 404 
Suggestion:
Maybe change the get(0) by a findFirst() like [here](https://github.com/ICIJ/datashare/blob/cfc6a4444b2bb1252a51b854d6d1b503a9bc9ad5/datashare-db/src/main/java/org/icij/datashare/db/JooqBatchSearchRepository.java#L141).

PR #1122, opened 2023-07-10
fix: unknown batch search wrong status

pirhoo:
## PR description

This tackles the issue of unknown BS returning a 500 as described in https://github.com/ICIJ/datashare/issues/1093.

## Changes

The error handling is done at two level:

* In the `BatchSearchResources` to ensure the correct code and error message is shown
* In the `BatchSearchRepository` to ensure get a BS from the db is done with a safe accessor

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/web/BatchSearchResource.java
+++ b/datashare-app/src/main/java/org/icij/datashare/web/BatchSearchResource.java
@@ -36,7 +36,6 @@ public class BatchSearchResource {
     private final BatchSearchRepository batchSearchRepository;
     private final BlockingQueue<String> batchSearchQueue;
     private final PropertiesProvider propertiesProvider;
-    private final PayloadFormatter payloadFormatter;
     private final int MAX_BATCH_SIZE = 60000;
 
     @Inject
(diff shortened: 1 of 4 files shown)


datashare-app/src/main/java/org/icij/datashare/web/BatchSearchResource.java
```diff
@@ -37,13 +39,15 @@ public class BatchSearchResource {
...
     private final PropertiesProvider propertiesProvider;
+    private final PayloadFormatter payloadFormatter;
     private final int MAX_BATCH_SIZE = 60000;
 
     @Inject
     public BatchSearchResource(final BatchSearchRepository batchSearchRepository, BlockingQueue<String> batchSearchQueue, PropertiesProvider propertiesProvider) {
         this.batchSearchRepository = batchSearchRepository;
         this.batchSearchQueue = batchSearchQueue;
         this.propertiesProvider = propertiesProvider;
+        this.payloadFormatter = new PayloadFormatter();
```
mvanzalu, line 50: This one can be removed

mvanzalu (approved):

datashare-app/src/main/java/org/icij/datashare/web/BatchSearchResource.java
pirhoo: Good catch! I need to rebase this branch to use the static method instead.

PR #1123, opened 2023-07-11
fix: use of TLS/auth for mail sender see #1106

bamthomas:
Second PR to fix the use of mailsender with authenticated user.

It comes with  :
* fix for the logger that was not properly used
* a new parameter for TLS connection (on 465)
* a new debug parameter to see exactly why the mail sending is failing
* extracted a `shouldAuth()` méthod that centralize the criteria for knowing if we sould auth or not

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/com/mail/MailSender.java
+++ b/datashare-app/src/main/java/org/icij/datashare/com/mail/MailSender.java
@@ -1,17 +1,13 @@
 package org.icij.datashare.com.mail;
 
-import java.net.URI;
-import java.net.URISyntaxException;
-import java.util.Properties;
-
-
 import com.sun.mail.smtp.SMTPMessage;
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.mail.*;
-import javax.mail.internet.AddressException;
 import javax.mail.internet.InternetAddress;
+import java.net.URI;
+import java.util.Properties;
 
 import static java.util.Optional.ofNullable;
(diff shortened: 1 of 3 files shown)


datashare-app/src/main/java/org/icij/datashare/com/mail/MailSender.java
```diff
@@ -82,16 +97,21 @@ private Message createMessage(Mail donneesEmail) throws MessagingException, Addr
...
+    private synchronized static Session getMailSession(MailSender mailSender) {
         Properties properties = new Properties();
         properties.setProperty("mail.smtp.class", "com.sun.mail.smtp.SMTPTransport");
-        properties.setProperty("mail.smtp.port", String.valueOf(portTransportMail));
-        properties.setProperty("mail.smtp.host", hostTransportMail);
-        return user == null ? Session.getDefaultInstance(properties): Session.getInstance(properties, new Authenticator() {
+        properties.setProperty("mail.smtp.port", String.valueOf(mailSender.port));
+        properties.setProperty("mail.smtp.host", mailSender.host);
+        properties.setProperty("mail.smtp.auth", String.valueOf(mailSender.shouldAuth()));
+        properties.setProperty("mail.smtp.ssl.enable", String.valueOf(mailSender.tls));
```
pirhoo, line 106: Shouldn't we have some sort of validation to ensure that String value of boolean are either "true" or "false"?

bamthomas: except if I don't understand everything, java boolean type is guarantying that the value cannot be something else than true/false ?

bamthomas: plus there are UT's on auth.

PR #1126, opened 2023-07-12
feat: swagger openapi

bamthomas:
This PR is adding an openapi endpoint that returns the specification file in json.


datashare-app/src/main/java/org/icij/datashare/web/OpenApiResource.java
```diff
@@ -0,0 +1,22 @@
...
+import net.codestory.http.payload.Payload;
+import org.icij.swagger.FluentReader;
+
+import static org.icij.swagger.ClassUtils.findAllClassesUsingClassLoader;
+
+@Singleton
+@Prefix("/api/openapi")
+public class OpenApiResource {
+    @Get()
+    public Payload get() {
```
pirhoo, line 18: Shouldn't we annotation this endpoint as well?

bamthomas: haha yes. I'm adding it.

pirhoo (approved):

PR #1128, opened 2023-07-13
feat: allow multi projects in local

pirhoo:
## PR description

This PR ensures that users are granted access to every projects in LOCAL mode.

## Changes 

* Upgrade to Datashare API 11.0.1 in order to use `user.setProjects` ;
* In LocalUserFilter, build a user with access to all projects ;
* Adapt all the test to mock `getProjects` from the `JooqRepository` ;
* Refactor `IndexResource` with `PayloadFormatter` utils class ;
* Add `IndexAccessVerifier` utils class.

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/utils/IndexAccessVerifier.java
+++ b/datashare-app/src/main/java/org/icij/datashare/utils/IndexAccessVerifier.java
@@ -28,21 +28,32 @@ public class IndexAccessVerifier {
 
     public String checkPath(String path, Context context) {
         String[] pathParts = path.split("/");
-        if(pathParts.length < 2){
+        if (pathParts.length < 2) {
             throw new IllegalArgumentException(String.format("Invalid path: '%s'", path));
         }
-        if ("_search".equals(pathParts[0]) && "scroll".equals(pathParts[1])) {
+        if (isSearchScrollPath(path)) {
             return getUrlString(context, path);
         }
         String[] indexes = this.checkIndices(pathParts[0]).split(",");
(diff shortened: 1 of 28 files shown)


datashare-app/src/main/java/org/icij/datashare/utils/PayloadFormatter.java
```diff
@@ -8,7 +8,16 @@
 public class PayloadFormatter {
```
mvanzalu, line 13: Maybe this class can be `static` so that we can use it without instantiate it in the other classes

datashare-app/src/main/java/org/icij/datashare/utils/IndexAccessVerifier.java
```diff
@@ -0,0 +1,58 @@
...
+import org.icij.datashare.session.DatashareUser;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static java.lang.String.join;
+import static java.util.Arrays.stream;
+import static java.util.stream.Collectors.toList;
+
+public class IndexAccessVerifier {
```
mvanzalu, line 15: This one can be `static` too

datashare-app/src/main/java/org/icij/datashare/utils/PayloadFormatter.java
pirhoo: Agree, thanks. Changing now.

datashare-app/src/main/java/org/icij/datashare/utils/IndexAccessVerifier.java
pirhoo: Same, I'm changing :)

mvanzalu (approved):

PR #1162, opened 2023-08-08
feat: limit download for embedded docs

pirhoo:
## PR description

This add a new `--embeddedDocumentDownloadMaxSize` option which prevent from downloading embedded files when there root is too big. The default value for this option is `1G`. It works both from the API endpoint to download document and from the batch download runner (https://github.com/ICIJ/datashare/issues/1161).


caro3801 (commented):
Nice feature

caro3801 (approved):

PR #1168, opened 2023-08-30
Feat/amqp init

bamthomas:
This PR brings new utility classes for using AMQP on top of the rabbitMQ java library.

The goal of these classes are to provide a facility layer to avoid misusing AMQP connections/channels and also give a representation of configuration, queues and settings.

The use of classes can be seen in tests.


mvanzalu (approved):

datashare-app/src/main/java/org/icij/datashare/com/bus/amqp/Event.java
```diff
@@ -0,0 +1,42 @@
...
+
+import java.io.IOException;
+import java.io.Serializable;
+import java.util.Date;
+
+/**
+ * Base event for AMQP.
+ * It provides creation timestamp and ttl that can be used for error handling and republishing/DLQ strategies.
+ */
+public class Event implements Serializable {
```
ClemDoum, line 16: I'm wondering about the `Event` naming, I was thinking about something more generic like `Message`, which could be describe as task to perform, a task related event or anything else.

bamthomas:
thanks @ClemDoum 
done here eba92e30a802c3a8ab4a609ab8787fd699040408

PR #1172, opened 2023-08-30
refactor: pull deserialize method in Consumer

bamthomas:
This is a simplification for implementers of the `AbstractConsumer` they won't have to define the `deserialize(jsonBody)` function.
Plus removing uneeded `TimeoutException`.

So no need to define a new class for the consumer, the type inference will be enough.

```java
    // declaration of specific Event/Saver
    static class TestEvent extends Event {
        public final String field;
        @JsonCreator
        public TestEvent(@JsonProperty("field") String field) {
            this.field = field;
        }
    }

    static class TestEventSaver extends EventSaver<TestEvent> {
        @Override
        public void save(TestEvent event) {
           // do something with event
        }
    }
    
    // then somewhere (main or other classes)
    Consumer<TestEvent, TestEventSaver> consumer = new Consumer<>(new TestEventSaver(), AmqpQueue.EVENT, TestEvent.class);
    consumer.consumeEvents();
```

ds_amqp drawio(8) drawio(1)


mvanzalu (approved):

ClemDoum:
yay, much simpler this way

PR #1183, opened 2023-09-15
feature: support installation of plugins with extensions dependencies

ClemDoum:
# PR description

This PR adds the ability for `Plugin` to depend on a list of `Extension`.

The dependency was added on this direction because if a `Plugin` depends on an `Extension` the opposite it not always true and the extension might be usable standalone (even if a plugin can run on the top of it).

I also allowed a plugin to depend on several extensions (this is debatable but a bit more futureproof).

# Changes
### Added
- added an optional `public final List<String> extensions`  to the `Plugin` class

### Changed
- added a `private final ExtensionService extensionService;` to the `PluginService` which nows check if the installed plugin depends on extensions and installs them thanks to its `extensionService` if needed

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/Plugin.java
+++ b/datashare-app/src/main/java/org/icij/datashare/Plugin.java
@@ -34,7 +34,7 @@ public class Plugin extends Extension {
                   @JsonProperty("extensions") List<String> extensions
     ){
         super(id, name, version, description, url, homepage, Type.PLUGIN);
-        this.extensions = extensions == null ? List.of() : extensions;
+        this.extensions = ofNullable(extensions).orElse(List.of());
     }
 
     public Plugin(String id, String name, String version, String description, URL url, URL homepage){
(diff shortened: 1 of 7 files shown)


datashare-app/src/main/java/org/icij/datashare/Plugin.java
```diff
@@ -22,19 +22,28 @@
...
                   @JsonProperty("name") String name,
                   @JsonProperty("version") String version,
                   @JsonProperty("description") String description,
                   @JsonProperty("url") URL url,
-                  @JsonProperty("homepage") URL homepage){
+                  @JsonProperty("homepage") URL homepage,
+                  @JsonProperty("extensions") List<String> extensions
+    ){
         super(id, name, version, description, url, homepage, Type.PLUGIN);
+        this.extensions = extensions == null ? List.of() : extensions;
```
pirhoo, line 37: We might use an Opionalable instead:

```suggestion
        this.extensions = ofNullable(extensions).orElse(List.of());
```

pirhoo (approved):

ClemDoum:
Addresses #1183

Issue #1191, opened 2023-09-25
feat: add a block list for tika metadata fields


pirhoo:
The list must allow wildcard definition such as `metadata.tika_metadata_mboxparser_*`

Issue #39, opened 2019-01-15
Go offline at runtime


Soliine:
When user is offline, NLP models can't work without being blocked when they try to check if there are new models available.

bamthomas:
![offline_online](https://user-images.githubusercontent.com/551723/51481159-862ced00-1d93-11e9-8cb3-4a6504f501eb.jpg)

bamthomas:
next to the button "start a new task"

PR #1200, opened 2023-09-29
feat: tika metadata block list

pirhoo:
## PR description

This PR implements a block list for tika metadata to avoid adding necessary data to the index (see #1191). The change is done through Extract which support now a new `MetadataBlockList` in the `MetadataTransform` class (see PR [#39](https://github.com/ICIJ/extract/pull/39)).

## Changes

* chore: upgrade to Extract 6.1.2 which support a block list 
* feat: add a block list file
* test: ensure the tika metadata are filtered correctly

## Default block list

The metadata block list we defined use the following globs:

```
tika_metadata_unknown_tag_*
tika_metadata_custom_*
tika_metadata_mboxparser_*
tika_metadata_xmpmm_*
tika_metadata_message_raw_header_?_*
tika_metadata_message_raw_header_[123456789]*
tika_metadata_message_raw_header__
```


caro3801 (approved):
Lgtm! 
small curiosity questions : 
* would it be interesting for users to use their own block list? 
* is it interesting to add an api endpoint about this?

pirhoo:
Yes, good point. I think if you choose to do it that should be a proper user story.

Thanks for the review!

PR #1218, opened 2023-10-13
feat: rename saved search 

caro3801:
This PR is link to #1203 
It aims to rename a saved search by transforming the API endpoint to add an event into an upsert if the eventId to update is passed in the request.

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/web/UserResource.java
+++ b/datashare-app/src/main/java/org/icij/datashare/web/UserResource.java
@@ -19,6 +19,7 @@ import org.icij.datashare.text.Project;
 import org.jetbrains.annotations.NotNull;
 
 import java.net.URI;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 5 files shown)


datashare-app/src/main/java/org/icij/datashare/web/UserResource.java
```diff
@@ -104,19 +106,24 @@ private static boolean parseBooleanQueryArg(String desc) {
         return parseBoolean(desc) || getStringValue(desc).isEmpty() || !desc.equalsIgnoreCase("false");
     }
 
-    @Operation(description = "Add event to history. The event's type, the project ids and the uri are passed in the request body.<br>" +
+    @Operation(description = "Add or update an event to user's history. The event's type, the project ids and the uri are passed in the request body.<br>" +
+            "To update the event's name, the eventId is required to retrieve the corresponding event." +
             "The project list related to the event is stored in database but is never queried (no filters on project)")
     @ApiResponse(responseCode = "200", description = "returns 200 when event is added or updated.")
     @Put("/me/history")
```
pirhoo, line 113: I think there is some tech dept here: the creation should be done through a "POST" request. Therefore, the partial update should be done in a separated method with a "PATCH". This is non-blocking as it's not increasing the technical dept.

pirhoo (approved):

datashare-app/src/main/java/org/icij/datashare/web/UserResource.java
caro3801: Yes definitively something that can be improved with your suggestion, in recommendations PUT should be idempotent.

PR #1224, opened 2023-10-18
fix: datashare parent pom for CLI module

bamthomas:
A plugin that needs a CLI extension must include a dependency on [datashare-cli](https://central.sonatype.com/artifact/org.icij.datashare/datashare-cli). The problem is that it has a parent pom `datashare.pom` that is not published on maven central.

This PR is decoupling the CLI module from it's parent. The datashare main pom has always `datashare-cli` as an aggregation module, including it in the build:

```
<modules>
    <module>commons-test</module>
    <module>datashare-db</module>
    <module>datashare-cli</module>
    <module>datashare-index</module>
    <module>datashare-nlp-corenlp</module>
    <module>datashare-dist</module>
    <module>datashare-app</module>
</modules>
```

The only problem doing this is that it cannot inherit from datashare main pom:

- dependencies versions
- dependencies management
- other properties (module version and so on)

We tried to refactor using the property plugin, that allows to put in a property file all the project properties but it seems that maven needs [dependencies versions to be fixed before its initialization phase](https://stackoverflow.com/questions/75800209/how-can-i-dynamically-add-version-of-any-dependency-in-pom-xml-without-profiling/76317839#76317839).

That means that the version is not in sync with the other datashare modules.

We should also decide when we want to publish datashare-cli module on the maven central repository (it has been done manually).

That shows also that we could bring back `datashare-api` in this repo.


ClemDoum (approved):

ClemDoum:
Thank you !

PR #1231, opened 2023-10-20
feature:`datashare-cli` `CliExtension` with dependencies

ClemDoum:
# Descriptions

Completes the work done in #1207 to allow to extend Datashare CLI with `CliExtension` which have dependencies.

Previously the extension mechanism required the `CliExtension` to have a [zero-arg constructor](https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html) which was very limiting as most CLI extension will need to use dependency injection

This PR leverages a factory-like pattern to allow extending the CLI with extensions relying on dependency injection.

Since dependency injection is only needed if the extension **if the extension is actually run**, the PR decouples extended the CLI (by adding the `Cli.CliExtender` interface) and running the actual CLI given properties (`Cli.CliRunner` interface).

A `CliExtensionFactory` interface was introduce to enable developers to create extensions which needs dependency injection, this factory class mainly needs to return implemented extension class through the `type` method and inherit from this class in order to implement `identifier` and `addOptions`. 
```java
public class CliExtensionFooFactory extends CliExtensionFoo implements CliExtensionFactory {
    public Class<? extends CliExtension> type() {
        return CliExtensionFoo.class;
    }
}
```

This design is a bit heavy but the fundamental problem is that, `identifier` and `addOptions` methods should actually be `static`  since nothing in them depends on the instance. If they were, we could access them through the class return by `type()` (and we couldn't have to `extend` that exact same type).

However Java doesn't allow for [`abstract static` methods](https://stackoverflow.com/a/1595000) so no simple solution could be found.

### DIFF
--- a/datashare-cli/src/test/java/org/icij/datashare/cli/CliExtensionFoo.java
+++ b/datashare-cli/src/test/java/org/icij/datashare/cli/CliExtensionFoo.java
@@ -1,19 +1,32 @@
 package org.icij.datashare.cli;
 
-import joptsimple.OptionParser;
-import org.icij.datashare.cli.spi.CliExtension;
-
-import java.util.Properties;
 
 import static java.util.Collections.singletonList;
 
+import com.google.inject.Inject;
+import java.util.Map;
+import java.util.Properties;
+import java.util.stream.Stream;
+import joptsimple.OptionParser;
+import org.icij.datashare.cli.spi.CliExtension;
+
 public class CliExtensionFoo implements CliExtension {
+    protected final Properties fooProperties;
+
+    public CliExtensionFoo() {
+        this.fooProperties = null;
+    }
+
+    @Inject
+    public CliExtensionFoo(Properties fooProperties) {
+        this.fooProperties = fooProperties;
+    }
+
     @Override
     public void addOptions(OptionParser parser) {
-        parser.acceptsAll(
-                        singletonList("foo"), "Test foo")
-                .withRequiredArg()
-                .ofType(String.class);
+        parser.acceptsAll(singletonList("foo"), "Test foo")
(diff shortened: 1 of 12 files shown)


datashare-cli/src/test/java/org/icij/datashare/cli/CliExtensionFoo.java
```diff
@@ -23,5 +36,13 @@ public String identifier() {
...
 
     @Override
-    public void run(Properties properties) {}
+    public void run(Properties properties) throws Exception {
+        Stream<Map.Entry<Object, Object>> props = properties.entrySet().stream();
+        if (this.fooProperties != null) {
+            props = props.filter(e -> !this.fooProperties.containsKey(e.getKey()));
+        }
+        props.forEach(
+            e -> System.out.println("extra key: " + e.getKey() + " extra value: " + e.getValue()));
```
pirhoo, line 45: This is a bit unusual to me. Can you explain a little bit?

pirhoo (approved):

datashare-cli/src/test/java/org/icij/datashare/cli/CliExtensionFoo.java
ClemDoum: I wanted to mock some CLI extension which needed to have some attributes injected. So I created an extension which get some properties injected by when create and outputs they diff with the properties provided at runtime on `stdout`

pirhoo: Make sense, thanks!

ClemDoum:
Closed by #1237

PR #1236, opened 2023-10-24
chore: allow extending env `CLASSPATH` in docker image

ClemDoum:
# Context

When using CLI extensions (#1207), it will be necessary to start `datashare` with an extended classpath pointing to the CLI extension jars. This PR allows to extend the docker image `CLASSPATH` rather than (re)-setting it in the `datashare`'s docker entrypoint.

# Changes
### Changed
- extend the image environment `CLASSPATH` with datashare's class path when calling docker entry point


ClemDoum:
@bthomas should we update:
- the mac installer
- the debian dist
- windows dist

bamthomas (approved):

PR #1242, opened 2023-10-25
refactor: cache batch search list

caro3801:
Related to  #1240
Use GET HTTP verb instead of POST to have 304 not modified HTTP status when query/response are identical (eTag) 
This is particularly important during polling to prevent response to be downloaded for each (unchanged) query (previous behavior returned 200 OK with same response each time)


bamthomas (approved):

bamthomas:
aww gros effort de doc des paramètres

pirhoo:
This is great, thanks!

PR #1246, opened 2023-10-31
feat: some minor oauth improvements

erikkri:
Here are some small improvements that we needed to get Datashare working with Okta and OIDC.

1. Support for `X-Forwarded-Proto` 
You already have support for `X-Forwarded-For` so this was needed for us to get the protocol correct for the oauth callback.

2. Support for scope in the callback url
To use OIDC we need scope in the callback url, added a cli option for this.

3. Support for sending oauth users to a specific project
We would like to send all okta users to the same project, and also the OIDC mappings in Okta didn't support adding arrays of string/list pairs, so we couldn't set `groups_by_applications` on the Okta side.


pirhoo:
Thanks @erikkri! We will have a look shortly :)

PR #1250, opened 2023-10-31
Refactor: reduce active connections

bamthomas:
The goal of this PR is to use only one DB connection pool for all repositories.

* RepositoryFactory is caching the pool and instantiate each repository with the same pool.
* the LocalMode that was manually creating a pool is using guice injection


bamthomas:
duplicate of #1245

PR #1252, opened 2023-10-31
refactor: warnings hunting

caro3801:
In relation to #1250 this PR goals is fix warnings raised by intelliJ

# Point of interest:
- a duplication has been removed [e42f886](https://github.com/ICIJ/datashare/pull/1252/commits/e42f88696d27abe2f15524b3fa820a4260993668). was it a bug?
- Use of assert to prevent null pointer exceptions when necessary ([f1d97d6](https://github.com/ICIJ/datashare/pull/1252/commits/f1d97d64118a2ffa8da641a1912dd6cb316c5b16) and [4e66350](https://github.com/ICIJ/datashare/pull/1252/commits/4e6635016d00e668c0a2994182ec2f26a2333c7b)
Are you ok with the use of assert in the code ?

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/tasks/BatchSearchLoop.java
+++ b/datashare-app/src/main/java/org/icij/datashare/tasks/BatchSearchLoop.java
@@ -72,6 +72,7 @@ public class BatchSearchLoop {
                 logger.warn("batch was not executed : {}", notFound.toString());
             } catch (BatchSearchRunner.CancelException cancelEx) {
                 logger.info("cancelling batch search {}", currentBatchId);
+                assert currentBatchId != null;
                 batchSearchQueue.offer(currentBatchId);
                 repository.reset(currentBatchId);
             } catch (SearchException sex) {
--- a/datashare-app/src/main/java/org/icij/datashare/tasks/BatchSearchLoop.java
+++ b/datashare-app/src/main/java/org/icij/datashare/tasks/BatchSearchLoop.java
@@ -13,6 +13,7 @@ import sun.misc.Signal;
 import java.io.Closeable;
 import java.io.IOException;
 import java.util.List;
+import java.util.Objects;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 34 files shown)


pirhoo (approved):

ClemDoum (approved):
Thanks @caro3801, I love it. On my side nothing to add about the refactoring, I'm just wondering how we could harmonize the warnings in the feature since they depends on which IDE you use and your IDE settings.

For the IDE that might not be a real concern (we can expect a lot of IntelliJ users) however for the settings that might vary from one dev to another. 

If that's relevant we could try:
- to share IDE settings for the project, so that devs can adopt them for the project
- enrich the `Makefile` with commands to perform the checks, so that devs which don't want to adopt project settings can still perform checks

bamthomas:
great @caro3801 thanks.

I think the duplication was a bug.
For the assert I like it specifically if we can disable them. It relates for me as "prechecks" in [design by contract](https://en.wikipedia.org/wiki/Design_by_contract). 

I agree with Clement: we should provide settings for IDE to harmonize the warnings inside the ICIJ team (and also for ppl that do PR's).

Did you refer to this?

bamthomas (approved):

caro3801:
Thanks for the review comments and suggestions. 
Researching on the assert issue, I found this [thread](https://stackoverflow.com/questions/45632920/why-should-one-use-objects-requirenonnull)) which explains the benefits of using Object.requireNonNull(obj) to fail faster on null pointer exception in production instead of assert. So I replaced the asserts.
Find attached my IDE settings (default ones for IntelliJ IDEA 2022.3.1 (Ultimate Edition)) : 
[Project_Default.xml.json](https://github.com/ICIJ/datashare/files/13265902/Project_Default.xml.json)
<= it's an XML file but Github does not allow `.xml` uploads so I changed the name by adding `.json`.

datashare-app/src/main/java/org/icij/datashare/tasks/BatchSearchLoop.java
```diff
@@ -71,8 +72,8 @@ public void run() {
             } catch (JooqBatchSearchRepository.BatchNotFoundException notFound) {
                 logger.warn("batch was not executed : {}", notFound.toString());
             } catch (BatchSearchRunner.CancelException cancelEx) {
+                Objects.requireNonNull(currentBatchId,"BatchSearch Id cannot be null");
```
caro3801, line 75: add optionnal on string batch search

PR #1254, opened 2023-11-07
fix: the dummy auth filter was broken

bamthomas:
Two issues:
* there was a misspelled and copy paste mistake in the test "protectedUrPrefix"
* the ApiKey filter was called before the auth filter, thus sending back a 401 that displayed the auth page


pirhoo (approved):

caro3801 (approved):
Nice catch

Issue #1081, opened 2023-06-05
refactor: migrate batch download to AMQP


ClemDoum:
Depends on #1079

Let's start with batch download that only uses the databus (no DB).

* [x]  remove `save(Task)` that was made for progress to use event bus
* [x]  replace `Queue<BatchDownload>` by `TaskSupplier`  (in TaskManagerRedis, BatchDownloadLoop)
* [x]  make task manager memory display progress
* [x]  make task manager redis display progress
* [x]  make batchDownloadLoop run in background as batchSeachLoop in local mode
* [x]  refactor `startTask(Callable<V> task, Map<String, Object> properties)` to  `startTask(Class<T> taskName, Map<String, Object> properties)` and  make changes to task managers (memory, redis) to make this refactor work
* [x]  implement `startTask(Class<T> taskName, Map<String, Object> properties)` for AMQPTaskManager 
* [x]  implement `TaskModifier`, `TaskSupplier` for AMQP
* [x] allow the CLI/modes to instantiate AMQP `TaskManager`/`TaskModifier`/`TaskSupplier`
* [x] in local mode instantiate Qpid broker
* [x] test manually that batch download is working with AMQP and no-regression for Memory/Redis 

The goal is to uncouple the task scheduling (that is made from the "client" and could be remote) from task executing that will be made by "generic" task runners somewhere in the network.

The idea is to push "taskview" that is mainly a task "name" and task properties map. In  a first move, task name could be used to instantiate by introspection real executable tasks, providing task properties to the constructor.

These tasks are sent over the wire when scheduling. For the update of the tasks (progress, result, errors) it will be sent by events in another AMQP channel. So we have to think carefully how it could work for the local/memory case.

The smallest move found for now is to split TaskManager that was executing tasks into : 

```java
@FunctionalInterface
public interface TaskModifier {
    void progress(String taskId, double rate);
}
```

whose responsibility is to allow task runners to update tasks (maybe it could be split into 3 functional interfaces?)
And 

```java
public interface TaskSupplier extends TaskModifier {
    <V> TaskView<V> get(int timeOut, TimeUnit timeUnit) throws InterruptedException;
    <V> void result(String taskId, V result);
    void error(String taskId, Throwable error); // necessary ?
}
```

And a remodeled proper TaskManager as a client

```java
public interface TaskManager {
    <V> TaskView<V> startTask(String taskName, Map<String, Object> properties);
    boolean stopTask(String taskId);
[...]

PR #1264, opened 2023-11-29
Feat/amqp batch download

bamthomas:
Merge refactor batch download before implementing AMQP.

cf https://github.com/ICIJ/datashare/issues/1081


datashare-app/src/main/java/org/icij/datashare/tasks/TaskModifier.java
```diff
@@ -0,0 +1,6 @@
+package org.icij.datashare.tasks;
+
+@FunctionalInterface
+public interface TaskModifier {
```
ClemDoum, line 4: In my opinion the `TaskModifier` is not actually modifying the task, but it's rather publishing/broadcasting updates or event related to task changes.

For now it's only progress, but that could be anything that might be helpful to track tasks. The `TaskManager.getTask` will of course be the source of truth, but there could be other consumers of these events (say for instance we want to implement an event source pattern, then there will be a a component listening for these updates and record them somewhere).

I'm the neo4j implementation I have chosen the naming [TaskEvent](https://github.com/ICIJ/datashare-extension-neo4j/blob/main/neo4j-app/neo4j_app/icij_worker/task.py#L227) and [EventPublisher](https://github.com/ICIJ/datashare-extension-neo4j/blob/main/neo4j-app/neo4j_app/icij_worker/event_publisher/event_publisher.py#L6), but anything like conveying the idea of broadcasting/publishing an task related update/event would be appropriate in my opinion.

datashare-app/src/main/java/org/icij/datashare/tasks/TaskView.java
```diff
@@ -39,19 +51,27 @@ public TaskView(MonitorableFutureTask<V> task) {
...
         }
     }
+
+    public TaskView(String name, User user, Map<String, Object> properties) {
+        this(randomUUID().toString(), name, State.INIT, 0, user, null, properties);
+    }
+
     @JsonCreator
-    TaskView(@JsonProperty("name") String name,
+    TaskView(@JsonProperty("id") String id,
```
ClemDoum, line 60: I think we talked about it early (and this might be commented somewhere else) but that would be nice to add as a TODO idea to seperate the `result` from the view.
In my mind the view is supposed to reflect the task state and metadata, it should be quickly and frequently accessed. Result on the other hand might be heavy and long to deserialize, which could prevent quick and frequent access to the view.
That can be address in a later PR, but that's to keep in mind in my opinion.

Issue #1248, opened 2023-10-31
feat: batch search, save elastic body in DB instead of query


pirhoo:
Batch searches are currently storing the query as a string. It's complicated to debug since 
Storing the elastic body string instead of the query would help to have a better reproducibility of the search: we would be certain of the query triggered instead of relying on the interpretation of the string stored in DB.

However this changes can lead to important changes and also require retro compatibility adaptors to be able to get the ES body from old batch searches

update:
Dev team discussed and this issue is blocked until update of elastic search client update

PR #1278, opened 2023-12-11
feat: adds tags in batch search

mvanzalu:
PR related to [this issue](https://github.com/ICIJ/datashare/issues/1071) and [this one](https://github.com/ICIJ/datashare/issues/1248).

### DIFF
--- a/datashare-index/src/main/java/org/icij/datashare/text/indexing/elasticsearch/ElasticsearchSearcher.java
+++ b/datashare-index/src/main/java/org/icij/datashare/text/indexing/elasticsearch/ElasticsearchSearcher.java
@@ -4,7 +4,10 @@ import co.elastic.clients.elasticsearch.ElasticsearchClient;
 import co.elastic.clients.elasticsearch._types.FieldValue;
 import co.elastic.clients.elasticsearch._types.Time;
 import co.elastic.clients.elasticsearch._types.query_dsl.BoolQuery;
-import co.elastic.clients.elasticsearch.core.*;
+import co.elastic.clients.elasticsearch.core.ClearScrollRequest;
+import co.elastic.clients.elasticsearch.core.ScrollRequest;
+import co.elastic.clients.elasticsearch.core.SearchRequest;
+import co.elastic.clients.elasticsearch.core.SearchResponse;
 import co.elastic.clients.elasticsearch.core.search.Hit;
 import co.elastic.clients.elasticsearch.core.search.ResponseBody;
 import com.fasterxml.jackson.databind.JsonNode;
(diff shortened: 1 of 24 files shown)


datashare-index/src/main/java/org/icij/datashare/text/indexing/elasticsearch/ElasticsearchSearcher.java
```diff
@@ -154,19 +159,41 @@ public Indexer.Searcher set(JsonNode jsonQuery) {
...
+    }
+
     @Override
     public Indexer.Searcher with(String query, int fuzziness, boolean phraseMatches) {
+        String queryString = buildQueryString(query, fuzziness, phraseMatches, "\"");
+        this.boolQueryBuilder.must(m -> m.matchAll(ma -> ma));
+        this.boolQueryBuilder.must(m -> m.queryString(qs -> qs.query(queryString)));
+        return this;
+    }
+    private static String buildQueryString(String query, int fuzziness, boolean phraseMatches, String phraseMatchDoubleQuotes) {
```
mvanzalu, line 187: I added the last parameter `String phraseMatchDoubleQuotes` because it seems `qs.query(queryString)` (l. 184) and `m.withJson(new StringReader(queryBody)` (l.176) are parsing the jsonString in different ways. There must be a better way to do this

datashare-index/src/main/java/org/icij/datashare/text/indexing/elasticsearch/ElasticsearchSearcher.java
```diff
@@ -154,19 +159,41 @@ public Indexer.Searcher set(JsonNode jsonQuery) {
...
+        }
+
+        String queryString = buildQueryString(query, fuzziness, phraseMatch, "\\\\\"");
+        jsonQueryTemplate = queryTemplate.matcher(jsonQueryTemplate).replaceAll(queryString);
+        final String queryBody = jsonQueryTemplate;
+
+        this.boolQueryBuilder.must(m -> m.withJson(new StringReader(queryBody)));
+        return this;
+    }
+
```
bamthomas, line 179: I'm not sure to understand why we need those regexps?

As it is string manipulation, could we do that in the client?
Or could we use the BatchSearch object fuzziness/phrase match?

datashare-index/src/main/java/org/icij/datashare/text/indexing/elasticsearch/ElasticsearchSearcher.java
bamthomas: yes cf my other comment I see duplication between BatchSearch attributes and body template.

datashare-index/src/main/java/org/icij/datashare/text/indexing/elasticsearch/ElasticsearchSearcher.java
mvanzalu: I thought that the queryBody could have also contained information about the fuzziness and phrase match but the information is already in the `BatchSearch` object indeed, the implementation would be easier and more straightforward with it.

The client can't do the whole string manipulation because it doesn't know the query, for example in the case where the `<query>` contains multiple words and the phrase_match is off it would have to build something like `term01~1 term02~1 ...`

mvanzalu: Anyway I'll do the change, the front will just send a `<query>` to replace and the back will use the BatchSearch object to get the fuzziness and the phrase_match and the regexp won't be needed anymore

bamthomas (approved):

PR #1284, opened 2023-12-20
Feat: amqp cli wirering

bamthomas:
Instantiate AMQP task manager when `batchQueueType` option is set to AMQP.

Fix TaskManager Redis that has an unknown queueName.

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/com/bus/amqp/AmqpInterlocutor.java
+++ b/datashare-app/src/main/java/org/icij/datashare/com/bus/amqp/AmqpInterlocutor.java
@@ -35,7 +35,7 @@ public class AmqpInterlocutor {
 
     @Inject
     public AmqpInterlocutor(PropertiesProvider propertiesProvider) throws IOException, URISyntaxException {
-        this(new Configuration(new URI(propertiesProvider.get("messageBusAddress").orElse("amqp://localhost:5672"))));
+        this(new Configuration(new URI(propertiesProvider.get("messageBusAddress").orElse("amqp://rabbitmq:5672"))));
     }
 
     AmqpInterlocutor(Configuration configuration) throws IOException {
--- a/datashare-app/src/main/java/org/icij/datashare/com/bus/amqp/AmqpInterlocutor.java
+++ b/datashare-app/src/main/java/org/icij/datashare/com/bus/amqp/AmqpInterlocutor.java
@@ -11,7 +11,6 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
-import java.net.URL;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -36,7 +35,7 @@ public class AmqpInterlocutor {
 
     @Inject
     public AmqpInterlocutor(PropertiesProvider propertiesProvider) throws IOException, URISyntaxException {
-        this(new Configuration(new URI(propertiesProvider.get("amqpAddress").orElse("amqp://localhost:5672"))));
+        this(new Configuration(new URI(propertiesProvider.get("messageBusAddress").orElse("amqp://localhost:5672"))));
     }
 
     AmqpInterlocutor(Configuration configuration) throws IOException {
@@ -73,6 +72,17 @@ public class AmqpInterlocutor {
             throw new UnknownChannelException(queue);
         }
         return channel;
+    }
+
+    public AmqpInterlocutor createAllPublishChannels() {
+        for (AmqpQueue queue: AmqpQueue.values()) {
+            try {
+                createAmqpChannelForPublish(queue);
+            } catch (IOException e) {
+                logger.error("cannot create channel for publish for queue {}", queue);
+            }
+        }
+        return this;
     }
 
     public synchronized AmqpInterlocutor createAmqpChannelForPublish(AmqpQueue queue) throws IOException {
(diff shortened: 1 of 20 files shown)


datashare-app/src/main/java/org/icij/datashare/com/bus/amqp/AmqpInterlocutor.java
```diff
@@ -32,8 +34,8 @@ public class AmqpInterlocutor {
 
 
     @Inject
-    public AmqpInterlocutor(PropertiesProvider propertiesProvider) throws IOException {
-        this(new Configuration(propertiesProvider.getProperties()));
+    public AmqpInterlocutor(PropertiesProvider propertiesProvider) throws IOException, URISyntaxException {
+        this(new Configuration(new URI(propertiesProvider.get("messageBusAddress").orElse("amqp://localhost:5672"))));
```
pirhoo, line 38: I think most default values use then docker-compose hostnames (such as `redis://redis`, `http://elasticsearch:9200`, etc). Should we use the same here?

```suggestion
        this(new Configuration(new URI(propertiesProvider.get("messageBusAddress").orElse("amqp://amqp:5672"))));
```

datashare-app/src/main/java/org/icij/datashare/mode/CommonMode.java
```diff
@@ -259,8 +291,8 @@ Routes addExtensionConfiguration(Routes routes) {
         return routes;
     }
 
-    protected boolean hasRedisProperty() {
-        return propertiesProvider.getProperties().contains(QueueType.REDIS.name());
+    protected boolean hasProperty(QueueType queueType) {
+        return propertiesProvider.getProperties().contains(queueType.name());
```
pirhoo, line 295: I'm not sure about this. Does it check that any property in the property provider contains the queue type? Shouldn't we check the `queueType` property instead?

pirhoo: Or `batchQueueType`.

datashare-app/src/main/java/org/icij/datashare/tasks/TaskManagerRedis.java
```diff
@@ -130,6 +132,9 @@ public void error(String taskId, Throwable reason) {
         tasks.put(taskId, taskView);
     }
 
+    @Override
+    public void close() throws IOException {}
```
pirhoo, line 136: Nothing to do with Redis to close ongoing tasks?

datashare-app/src/test/java/org/icij/datashare/tasks/TaskManagerAmqpTest.java
```diff
@@ -29,8 +28,9 @@ public class TaskManagerAmqpTest {
     @ClassRule static public AmqpServerRule qpid = new AmqpServerRule(5672);
     TaskManagerAmqp taskManager;
     TaskSupplierAmqp taskSupplier;
+    CountDownLatch nextMessage;
 
-    @Test
+    @Test(timeout = 2000)
```
pirhoo, line 33: Why do we need to increase the timeout here?

pirhoo (approved):

datashare-app/src/main/java/org/icij/datashare/mode/CommonMode.java
bamthomas: yes it was any property because there are `queueType` and `batchQueueType`. This is inherited  from all changes but as we talked about this yesterday with MV it will have to be harmonized. So the idea was : if there is AMQP anywhere we instantiate AMQP interlocutor, and if there is redis anywhere we do a RedisClient also;

datashare-app/src/test/java/org/icij/datashare/tasks/TaskManagerAmqpTest.java
bamthomas: because we've seen that the test may block indefinitely.

datashare-app/src/main/java/org/icij/datashare/mode/CommonMode.java
pirhoo: Got it, thanks!

datashare-app/src/test/java/org/icij/datashare/tasks/TaskManagerAmqpTest.java
pirhoo: Ah ok, make sense!

datashare-app/src/main/java/org/icij/datashare/com/bus/amqp/AmqpInterlocutor.java
bamthomas: Right. 
Actually we've put the names of the services that we had in the development environment. For automated integration tests and for manual local testing. Here `localhost` is used because we use qpid for unit tests, not the docker container. 

If we'd want to follow the same rule, we should use the _broker_ name i.e. `amqp://rabbitmq:5672`.
What do you think?

pirhoo: Yes, `rabbitmq` makes much more sense to me :)

bamthomas: https://github.com/ICIJ/datashare/pull/1284/commits/3b582f0454a9059e0997ce89033bb706ef4e38da

datashare-app/src/main/java/org/icij/datashare/tasks/TaskManagerRedis.java
bamthomas: see https://github.com/ICIJ/datashare/pull/1284/commits/ffaafe117be48a2e525ff8a9123cdf144f2afeb0

PR #1334, opened 2024-02-22
feat: ensure consistancy when creating/deleting queues in LOCAL mode

pirhoo:
## PR description

While debugging #1333 we discovered the queues are not correctly deleted when deleting a project. To avoid this issue from happening again, this PR adds more constraints to the queues created in LOCAL (from the client).

## Changes

* feat: scan and index endpoint constraint the queueName based on the project name 
* feat: project deletion endpoint deletes queues from every stage
* refactor: find queues to delete using the PipelineHelper class

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/web/TaskResource.java
+++ b/datashare-app/src/main/java/org/icij/datashare/web/TaskResource.java
@@ -120,11 +120,12 @@ public class TaskResource {
     @Post("/batchDownload")
     public TaskView<File> batchDownload(final OptionsWrapper<Object> optionsWrapper, Context context) throws Exception {
         Map<String, Object> options = optionsWrapper.getOptions();
-        Path downloadDir = get(propertiesProvider.getProperties().getProperty(BATCH_DOWNLOAD_DIR_OPT));
+        Properties properties = applyProjectProperties(optionsWrapper);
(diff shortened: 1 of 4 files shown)


datashare-app/src/main/java/org/icij/datashare/web/TaskResource.java
```diff
@@ -249,17 +248,39 @@ public Payload stopAllTasksPreflight() {
...
+    public static Properties applyProjectTo(Properties properties) {
+        Properties clone = (Properties) properties.clone();
+        clone.setProperty(QUEUE_NAME_OPTION, getQueueNameFor(properties));
+        clone.setProperty(MAP_NAME_OPTION, getReportMapNameFor(properties));
+        return clone;
+    }
+
+    public static String getQueueNameFor(Properties properties) {
+        String projectName = properties.getOrDefault("defaultProject","local-datashare").toString();
+        return "extract:queue:" + projectName;
```
mvanzalu, line 276: The return value is the whole queue name or is it just the prefix ?

pirhoo: Just a prefix, hence why we still need to delete all candidates values: https://github.com/ICIJ/datashare/pull/1334/files#diff-eccbec11c2a85c47442becddc5cb5d77c93a4c44eaf29565d6fe51a00a19d6d3R237-R240

mvanzalu (approved):

Issue #1270, opened 2023-12-05
feat: add option to specify scroll duration for Batch Search, Batch Download and SCANIDX


pirhoo:
In batch download mode, we could use:

* `--scroll <String>` with the duration of the scroll as [time value](https://www.elastic.co/guide/en/elasticsearch/reference/current/api-conventions.html#time-units)
* `--scrollSize <Integer>` with the size
* `--scrollSlices <Integer>` with the number of parallel slices

PR #1349, opened 2024-02-28
feat: adds scrollSize for batch and enqueue tasks

mvanzalu:
Related to [this issue](https://github.com/ICIJ/datashare/issues/1270)

This PR aims to add the scroll size option for Batch Search/Download and EnqueueTask.
It also define more default constants values options.

Scroll slice option for the other tasks can be issued in another story as the implementation is very specific for each tasks (we can also dig into the using of elasticsearch's [PIT](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/point-in-time-api.html#point-in-time-keep-alive)

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/tasks/BatchDownloadRunner.java
+++ b/datashare-app/src/main/java/org/icij/datashare/tasks/BatchDownloadRunner.java
@@ -76,10 +76,10 @@ public class BatchDownloadRunner implements Callable<FileResult>, Monitorable, U
 
     @Override
     public FileResult call() throws Exception {
-        int throttleMs = parseInt(propertiesProvider.get(BATCH_THROTTLE_OPT).orElse("0"));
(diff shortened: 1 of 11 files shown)


pirhoo (changes requested):
This looks good but it's not retro-compatible (CLI using the existing `scrollSize` will be ignored). Instead we could use the scrollSize value as default value in BatchSearchRunner and BatchDownloadRunner.

mvanzalu:
> This looks good but it's not retro-compatible (CLI using the existing `scrollSize` will be ignored). Instead we could use the scrollSize value as default value in BatchSearchRunner and BatchDownloadRunner.

Alright and having `batchSearchScrollSize`in the params would override the `scrollSize` parameter for example ?

pirhoo:
Exactly!

pirhoo (approved):

Issue #1080, opened 2023-06-05
refactor: migrate batch search to AMQP


ClemDoum:
Depends on #1079, #1081 and #1283  

At the end of this story, BatchSearchLoop and BatchDownloadLoop should be the same. 

* [x] do #1283 to make it easier for this story
* [x] refactor Event -> Message
* [x] remove `BatchDownloadCleaner` from `BatchDownloadLoop` and any specific batch download coupling then rename class
* [x] use `TaskView<T>` for sending batch search to worker(s)
* [x] use `TaskManager.startTask(String id, String taskName, User user)` like  `BatchDownload` 
* [x] use TaskRunnerLoop for batch search
* [x] adapt error handling with batch search (repost, dead letter queue) in TaskRunnerLoop
* [x] remove method `startTask(Callable<V> task, ...)` to use same methods as BatchDownload and BatchSearch
* [x] implement AMQP for `BatchSearch`
* [ ] use AMQP for working queues between SCAN, INDEX, NLP (full redis, full AMQP, full memory) if possible
* [x] manually test non regression with memory, Redis, AMQP

caro3801:
Closed in favor of testing issues in Sprint 25:
- Enhancement: AMQP for working queues between SCAN, INDEX, NLP (full redis, full AMQP, full memory) if possible
- batch search migration #1372
- test of the installers

PR #1365, opened 2024-04-02
fix: avoid to launch task manager in taskrunner #1080

bamthomas:
for BatchSearch daemon we were requeing queued batch searches with the taskmanager. The consequence was the instantiation of a task manager in the task runner daemon.

We also merged BatchSearchApp and BatchDownloadApp as TaskRunnerApp. 

It will change datashare CLI.


datashare-app/src/main/java/org/icij/datashare/WebApp.java
```diff
@@ -48,6 +52,7 @@ static void start(Properties properties) throws Exception {
             waitForServerToBeUp(parseInt(mode.properties().getProperty(PropertiesProvider.TCP_LISTEN_PORT)));
             Desktop.getDesktop().browse(URI.create(new URI("http://localhost:")+mode.properties().getProperty(PropertiesProvider.TCP_LISTEN_PORT)));
         }
+        requeueDatabaseBatches(mode.get(BatchSearchRepository.class), mode.get(TaskManager.class));
```
caro3801, line 55: rename to requeueDatabaseBatchSearches ?

caro3801 (approved):
The doc must be updated before merging

datashare-app/src/main/java/org/icij/datashare/WebApp.java
bamthomas: ok

PR #1390, opened 2024-04-24
chore: extract async task lib

ClemDoum:
# TODO
- [x] handle redisson version properly (try to remove the explicit version in `datashare-app`)
- [ ] CI for `datashare-tasks`
- [ ] JAR publication for `datashare-tasks`
- [x] validate naming (`datashare-tasks` vs. `datashare-async-tasks`)
- [ ] review TODOs left in the code

# PR description

This PRs extracts the asynchronous tasks logic of DS in its own package. The goal of the operation is that the package can be used by other repos than DS if they want to implement async tasks (neo4j proeject for instance).

# Changes

### Removed
- removed async tasks logic and added a dependency on `datashare-tasks`

### Added
- created the `datashare-tasks` repo

### Changed
- replaced `CancellableCallable` by `CancellableTask`
- renamed `CanceledEvent` into `CancelledEvent`
- task function now take a progress which is only called with the `Double progress` argument. The `String taskId` part of the progress is forwarded inside the `TaskRunnerLoop`. This avoid the task implementer to possibly call `progress(taskId, progress)` with the wrong `taskId`.

# Left for later
- wrap task cancellation logic in its own function and keep track of the original exception is something goes wrong while handling task error
- align Java and Python
- decouple task name and object name as the current task mapping relies as the current task name mapping heavily relies on class naming and will prevent refactor/renaming/moving classes


datashare-app/src/test/java/org/icij/datashare/extract/RedisBlockingQueueTest.java
```diff
@@ -13,10 +14,11 @@
 import static org.fest.assertions.Assertions.assertThat;
 
 public class RedisBlockingQueueTest {
-    RedisBlockingQueue<String> queue = new RedisBlockingQueue<>(new PropertiesProvider(new HashMap<>() {{
+    RedissonClient client = new RedissonClientFactory().withOptions(Options.from(new HashMap<>() {{
         put("redisAddress", "redis://redis:6379");
         put("redisPoolSize", "5");
```
caro3801, line 19: suggestion (if-minor): extract redis configuration values to constant variable

datashare-db/src/main/resources/liquibase/changelog/db.changelog.yml
```diff
@@ -80,6 +80,7 @@ databaseChangeLog:
   - include:
       file: changes/026-drop-user-history-column-sqlite.yml
       relativeToChangelogFile: true
+      precondition:
```
caro3801, line 83: typo: remove line ?

datashare-tasks/src/test/java/org/icij/datashare/asynctasks/bus/amqp/AmqpServerRuleTest.java
```diff
@@ -1,8 +1,8 @@
-package org.icij.datashare.com.bus.amqp;
+package org.icij.datashare.asynctasks.bus.amqp;
 
 import com.rabbitmq.client.Connection;
 import com.rabbitmq.client.ConnectionFactory;
-import org.icij.datashare.com.bus.amqp.AmqpServerRule;
+import org.fest.assertions.Fail;
```
caro3801, line 5: remark : it looks like there is 2 assertion frameworks. It could become a challenge to maintain at some point.

datashare-tasks/src/main/java/org/icij/datashare/asynctasks/CheckedBiFunction.java
```diff
@@ -0,0 +1,6 @@
+package org.icij.datashare.asynctasks;
+
+@FunctionalInterface
+public interface CheckedBiFunction<T, U, R, E extends Exception> {
```
caro3801, line 4: question: I did not find where it is in use. I noticed that Bifunction has been removed in tasks, is this one still relevant?

caro3801 (commented):
Looks clear and well tested even if I'm not familiar with concurrent applications in java.
The comments about handling tasks closing in different languages is quite relevant.

PR #1395, opened 2024-04-30
Refactor/oauth2 filter

bamthomas:
This PR merges OAuth2 classes and harmonizes the oauth authentication process.

* it enabled a full login test
* it adds `oauthClaimIdAttribute` to CLI
* it makes the OAuth2CookieFilter final to avoid multiple implementation of oauth2


bamthomas:
could you tell us @erikkri and @neoReuters if that still works for you?

erikkri:
Thanks for the heads up! I’m not able to check this until thursday at the earliest.

neoReuters (approved):
Will test this out fully on our AWS Cognito instance when I'm back from annual leave in about two weeks - just a quick look through code however the changes definitely make sense - from our side the main change needed was being able to push through a custom claim Id attribute - the ALBCognito cookie filter was pretty much the same as the OAuth2 cookie filter - just did not want to introduce any breaking changes hence why we created a separate filter. This should suffice with that anyways.

Cheers @bamthomas!

pirhoo (approved):

bamthomas:
ok we merge the PR and make a release, it will make it easier for you to test. If there are some issues, don't hesitate to reopen this PR.

PR #1405, opened 2024-06-04
fix: various errors

mvanzalu:
This PR fixes 3 things : 

- The java opens error on the backend when we trying to retrieve tasks : 
```
Unable to make protected final void java.util.concurrent.locks.AbstractQueuedSynchronizer.setState(int) accessible: module java.base does not "opens java.util.concurrent.locks" to unnamed module @21213b92
 at [Source: (io.netty.buffer.ByteBufInputStream); line: 1, column: 11]
```
- The task runner couldn't start due to missing binding in `CommonMode`
```
Exception in thread "main" com.google.inject.ConfigurationException: Guice configuration errors:
1) No implementation for org.icij.datashare.asynctasks.TaskFactory was bound.
  while locating org.icij.datashare.asynctasks.TaskFactory
1 error
	at com.google.inject.internal.InjectorImpl.getProvider(InjectorImpl.java:1120)
	at com.google.inject.internal.InjectorImpl.getProvider(InjectorImpl.java:1078)
	at com.google.inject.internal.InjectorImpl.getInstance(InjectorImpl.java:1131)
	at org.icij.datashare.mode.CommonMode.get(CommonMode.java:126)
	at org.icij.datashare.TaskRunnerApp.start(TaskRunnerApp.java:16)
	at org.icij.datashare.Main.main(Main.java:27)
```
- The flaky tests in `UserTaskResource` class


bamthomas (approved):

datashare-app/src/main/java/org/icij/datashare/mode/CommonMode.java
```diff
@@ -132,6 +132,7 @@ public Injector createChildInjector(Module... modules) {
     protected void configure() {
         bind(PropertiesProvider.class).toInstance(propertiesProvider);
         install(new FactoryModuleBuilder().build(TaskFactory.class));
+        install(new FactoryModuleBuilder().build(org.icij.datashare.asynctasks.TaskFactory.class));
```
bamthomas, line 135: as `TaskFactory` is having its parent `org.icij.datashare.asynctasks.TaskFactory` (that is empty) we shouldn't install the other : either the parent is known in the class path and the children is defined, either not and there should be exceptions while compiling.

bamthomas (changes requested):

datashare-app/src/main/java/org/icij/datashare/mode/CommonMode.java
bamthomas: it seems that guice is looking for a concrete instance for the parent class `org.icij.datashare.asynctasks.TaskFactory`. So we have to define the parent as a factory.

bamthomas (approved):

PR #1413, opened 2024-06-18
Fix/plugin neo4j exceptions

bamthomas:
the main goal of the PR is to fix the erroneous `module-info.class` loading by the `ExtensionLoader`, and more generally the `META-INF` folder loading.

This _may_ have some issues with multi-release feature because for multi-release (i.e. releases that contains different class path for the same class, each of which will be compatible with a given JDK). This multi-release compatibility of the `ExtensionLoader` could be added later if needed.

So a test have been added. It is using the ERROR/WARNING logs to check if the loading is skipping the `META-INF` files.

This has had some consequences on the build: we had to make the `commons-test.LogbackCapture` available for `datashare-api`. Pb : `commons-test` is depending on `datashare-api` so it would introduce a circular dependency. So a utility class has been created. 

Then there was a new transitive dependency issue: `RemoteFile` is pulling a lot of dependencies with `spring-boot` and as we are using the old v1 AWS API, it pulls an old version of `apache.commons-logging` and we are using a more recent version of logging layers that could not live with the old commons-logging. 

To fix this we used the s3mock in a docker container rather than the java dependency as [advised](https://github.com/adobe/S3Mock?tab=readme-ov-file#s3mock-java) by the documentation.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -72,7 +72,7 @@ jobs:
       - run:
           name: acceptance test for embedded mode
           command: |
-            VERSION=$(cat pom.xml | grep '<version>[0-9.]\+' | sed 's/<version>\([0-9.]\+\)<\/version>/\1/g' | tr -d '[:space:]')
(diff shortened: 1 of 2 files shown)


pirhoo (approved):

ClemDoum (approved):
Well done @bamthomas

Issue #1165, opened 2023-08-28
bug: ensure tika extracts embedded doc with a stream


pirhoo:
When an embedded file content is requested by a user, instead of extracting content in memory, we will generate a file on disk. It will create a cache that will be used by datashare to stream the content of embedded files.

We will add : 
- an option `artifact-dir` where the cache will be stored
- we will generate a binary contents file called `raw` and document metadata with the following structure:

```shell
    /artifact-dir/<id_first_two_chars>/<id_following_two_chars>/<doc_id>/raw <--- the file binary content
    /artifact-dir/<id_first_two_chars>/<id_following_two_chars>/<doc_id>/raw.json <--- the file metadata
    # and maybe later it could be extended with
    /artifact-dir/<id_first_two_chars>/<id_following_two_chars>/<doc_id>/text/1.png <-- first page text
    /artifact-dir/<id_first_two_chars>/<id_following_two_chars>/<doc_id>/text/2.png <-- second page text
    /artifact-dir/<id_first_two_chars>/<id_following_two_chars>/<doc_id>/thumbnail/1.png <-- page 1 thumbnail
    /artifact-dir/<id_first_two_chars>/<id_following_two_chars>/<doc_id>/thumbnail/2.png <-- page 2 thumbnail
```

For example:

```
/configured/path/to/artifacts/12/34/1234f5fc76b8e243c8b0ae42cbee55afd3b0c0ffe67d31a5a8f2a9b13f2998e8/raw
/configured/path/to/artifacts/12/34/1234f5fc76b8e243c8b0ae42cbee55afd3b0c0ffe67d31a5a8f2a9b13f2998e8/raw.json
# ...
```

If the option `artifactDir` is not provided, datashare will use memory as before.

part of #1397

bamthomas:
the result of extraction is stored into TikaDocumentSource class :

```java
public class TikaDocumentSource {
    public final Metadata metadata;
    public final byte[] content;

    public TikaDocumentSource(final Metadata metadata, final byte[] content) {
        this.metadata = metadata;
        this.content = content;
    }
}
```

And is is used in SourceExtractor class.

The content should not be stored as an array of byte but as an InputStream (or subclass).

mvanzalu:
No solution for disk issue for the moment

PR #1536, opened 2024-09-03
feat: embedded artifacts

bamthomas:
See #1165 for the list of commits. Changes are made in extract-lib. https://github.com/ICIJ/extract/commit/f09254eb0d05c54c8df657acf73641d3193e2f5b

If null is provided in the [EmbeddedDocumentExtractor](https://github.com/ICIJ/extract/blob/7ed996daa13e211bb2df488f712b2b46d30473a8/extract-lib/src/main/java/org/icij/extract/extractor/EmbeddedDocumentExtractor.java) as the `artifactPath` then memory is used, else the provided path will be used as file cache for embedded files.

### DIFF
--- a/datashare-cli/src/main/java/org/icij/datashare/cli/DatashareCliOptions.java
+++ b/datashare-cli/src/main/java/org/icij/datashare/cli/DatashareCliOptions.java
@@ -346,10 +346,8 @@ public final class DatashareCliOptions {
     static void artifactDir(OptionParser parser) {
         parser.acceptsAll(
                 asList(ARTIFACT_DIR_OPT),
-                "Artifact directory for embedded caching" )
-                .withRequiredArg()
-                .ofType(File.class)
-                .defaultsTo(new File(DEFAULT_ARTIFACT_DIR));
+                "Artifact directory for embedded caching. If not provided datashare will use memory." )
+                .withRequiredArg();
     }
 
     static void rootHost(OptionParser parser) {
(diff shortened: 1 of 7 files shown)


datashare-cli/src/main/java/org/icij/datashare/cli/DatashareCliOptions.java
```diff
@@ -119,13 +119,15 @@ public final class DatashareCliOptions {
...
     public static final String VERSION_OPT = "version";
+    public static final String ARTIFACT_DIR_OPT = "artifactDir";
 
     private static final Path DEFAULT_DATASHARE_HOME = Paths.get(System.getProperty("user.home"), ".local/share/datashare");
     private static final Integer DEFAULT_NLP_PARALLELISM = 1;
     private static final Integer DEFAULT_PARALLELISM = Runtime.getRuntime().availableProcessors() == 1 ? 2 : Runtime.getRuntime().availableProcessors();
     private static final Integer DEFAULT_PARSER_PARALLELISM = 1;
     public static final DigestAlgorithm DEFAULT_DIGEST_METHOD = DigestAlgorithm.SHA_384;
     public static final String DEFAULT_DATA_DIR = Paths.get(System.getProperty("user.home")).resolve("Datashare").toString();
+    public static final String DEFAULT_ARTIFACT_DIR = DEFAULT_DATA_DIR;
```
pirhoo, line 130: I think we should avoid mixing the artifact dir and the data dir. I would create a dedicated directory in the Datashare's home:

```suggestion
    public static final String DEFAULT_ARTIFACT_DIR = DEFAULT_DATASHARE_HOME.resolve("artifact").toString();
```

pirhoo (changes requested):

datashare-cli/src/main/java/org/icij/datashare/cli/DatashareCliOptions.java
bamthomas: yes I removed the default value from the option, but I forgot this.
Thank you.

pirhoo (approved):

pirhoo:
Well done!

Issue #1452, opened 2024-07-09
feature: implement spacy NER


ClemDoum:

ClemDoum:
# Batch NLP processing and worker lifecycle

## CoreNLP 
### CLI mode using `MEMORY` batch queue
- create NLP batches and process them using CoreNLP `./datashare --mode CLI --nlp CORENLP --batchQueueType MEMORY --nlpParallelism <p> --step CREATENLPBATCHESFROMINDEX,BATCHNLP`

### Local mode
- TM is `MEMORY` by default, CoreNlp workers are created using with a single Java threads create via `startTask` in `POST /findNames/:pipeline`
- no parallelism supported  (or I might be missing it ?), if it's the case should we expose `nlpParallelism` in setting when using `MEMORY`

## Spacy
Batch NLP task must be available to Python, this argument is in favor of using `--batchQueueType AMQP` to put created task in the broker and make them available to Python

### CLI mode (TODO)
- create NLP batches and make them available to Python `./datashare --mode CLI --nlp CORENLP --batchQueueType AMQP --step CREATENLPBATCHESFROMINDEX`
- problem with this approach is that we have several concurrent TMs running, one in the CLI, one in the DS backend, both can write task status which will end up in concurrency issues while handling task states
- Python workers can be launched independently from the DS CLI, but that would involve documenting the process (how to deploy the docker image...)

### Local mode (TODO ?)
- how do we make sure Python workers are running when hitting `POST /findNames/:pipeline` endpoint...
- how is the Python Spacy worker packaged (should we package it using `pyinstaller` which is probably painful, should alternatively use docker ?, should we distribute a python package via `pipx`)
- how is the Python worker lifecycle handled by Java (via an extension ?)

### Overall
Since Spacy and Python are meant to be the new standard for NER and other ML tasks to come, how do we
- offer a consistent and smooth user experience between CLI/LOCAL
- smoothen the process of shipping a Python worker from a dev perspective

PR #1541, opened 2024-09-05
feature: batch ner

ClemDoum:
# TODO
- [ ] merge #1538

# PR description

Implemenent batch text processing for NER, this change is made in the context of #1452, as batch processing is necessary for Spacy.

# Changes

### Added
- added the batch text processing API `List<List<NlpTag>> processText(Stream<String> batch, Language language) throws InterruptedException` to `Pipeline`

### Changed
- made `NlpTag` a record and json serializable class

### Added
- implemented batch processing for stanford core nlp)

### Added
- added `bool Pipeline.Type.extractFromDoc()` which indicates if the pipeline should preferrably used on full documents or can be used on text chunks
- implemented batch text processing inside the `ExtractNlpTask` for pipelines which do not require prediction on documents

### DIFF
--- a/datashare-api/src/main/java/org/icij/datashare/function/Pair.java
+++ b/datashare-api/src/main/java/org/icij/datashare/function/Pair.java
@@ -17,6 +17,10 @@ public class Pair<T1, T2> {
         second = snd;
     }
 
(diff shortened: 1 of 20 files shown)


ClemDoum:
Closed in favor of #1597

Issue #1534, opened 2024-08-28
feat: new stage ARTIFACT


bamthomas:
**Is your feature request related to a problem? Please describe.**

Avoid embedded access errors.

**Describe the solution you'd like**

When a project has already been indexed, we would like to add embedded artifacts on disk for a given project.

part of #1397

PR #1577, opened 2024-09-18
Feat/artifact task

bamthomas:
see #1534


datashare-cli/src/main/java/org/icij/datashare/cli/DatashareCliOptions.java
```diff
@@ -789,6 +790,12 @@ public static void oauthClaimIdAttribute(OptionParser parser) {
                 .ofType(String.class);
     }
 
+    public static void searchQuery(OptionParser parser) {
+        parser.acceptsAll(singletonList(SEARCH_QUERY_OPT), "Json query for filtering index matches for EnqueueFromIndex task.")
```
pirhoo, line 794: For the `searchQuery` option I would expect to have a lucene query string instead of a JSON body. Would be to have both (`searchQuery` and `searchBody`) but if I had to choose one I'll keep the first.

pirhoo (changes requested):
This is amazing, the ability to filter the enqueued document with a `searchQuery` is also a great improvement! I would just use it differently as explained in my comment :)

datashare-cli/src/main/java/org/icij/datashare/cli/DatashareCliOptions.java
bamthomas: It works with both I just added a test to show this. see 454b8333f9d2fc44542c2196c23e71210d6e44a0

pirhoo: So cool! Thanks

PR #1580, opened 2024-09-27
feat: parallelize cli with multithreaded task manager

bamthomas:
when we refactored the task management system we changed `CliApp` class to execute sequentially the tasks.

To distribute tasks like INDEX,ARTIFACT it could be better to execute these tasks in parallel.


mvanzalu (approved):

bamthomas:
it doesn't work with artifact

PR #1393, opened 2024-04-25
feat: custom classes for authUsersProvider & authFilter + new auth filter class for AWS ALB + Cognito

neoReuters:
## Overview
This pull request introduces several enhancements to authentication handling, specifically for projects integrating with AWS ALB and Cognito. It provides greater flexibility through custom classes and supports a new authentication scenario. 

## Changes
- **Custom Classes for authUsersProvider & authFilter**: These classes allow for more nuanced customisation of the authentication process, accommodating unique business requirements not previously covered. We ensure that we are using the system's class loader so that custom classes (even those via extensions for example) can be loaded by the Datashare application.
- **New Auth Filter Class for AWS ALB + Cognito**: This new filter supports scenarios where AWS ALB is used in conjunction with Cognito to manage authentication, providing a streamlined integration.

## Motivation
The current implementation of authentication classes is too rigid for our needs. By introducing these custom classes and the new filter, we can handle a broader range of authentication flows, specifically leveraging AWS services which are critical to our infrastructure.

## Impact
These changes are backward compatible with existing configurations. They provide optional enhancements and do not alter the current authentication process unless explicitly configured.

## Request for Feedback
I would appreciate feedback on the implementation of the new auth filter class, especially concerning its interaction with AWS Cognito. Any suggestions for improvements or potential issues you foresee are welcome.

Issue #504, opened 2020-05-29
[DaaP API] generate and get one API key from the command line


Soliine:
Part of #501 epic

bamthomas:
see https://www.freecodecamp.org/news/best-practices-for-building-api-keys-97c26eabfea9/

PR #1583, opened 2024-10-01
fix: overwriting what projects users have been assigned every time you login via oauth2

neoReuters:
### Overview
This pull request addresses an issue where logging in via the OAuth2 filter would overwrite the projects a user is assigned to, replacing them with the default project. This behavior was particularly undesirable for users with multiple project assignments, as it caused them to lose their existing projects. The fix ensures that existing projects are preserved, and the default project is only appended if it is not already present in the datashare list.

### Changes
1. Check for Existing groups_by_applications and datashare: The logic now checks if groups_by_applications already exists in the user's data and reuses it rather than overwriting it. If the datashare list exists, it appends the oauthDefaultProject only if it doesn't already exist.
2. Preserving Existing Projects: Instead of overwriting the list of projects in datashare, this fix ensures that additional projects are appended, preserving any projects that the user was already assigned to.

### Motivation
The issue was identified when logging in via the OAuth2 filter, where the user’s assigned projects were being overwritten with the default project. This was especially problematic for users with multiple project assignments, as they would lose their previously assigned projects each time they logged in. This fix provides a more flexible and reliable approach, ensuring that projects are appended rather than overwritten, preserving the existing assignments.

### Impact
These changes are backward compatible with existing configurations and will not alter the current behavior unless there is an issue with project overwrites. It improves the system’s data integrity, ensuring that multiple projects can be tracked without losing existing assignments.

### Request for Feedback
I would appreciate feedback on the overall approach for preserving and appending to datashare. Specifically, I’d welcome any suggestions for further optimizations or concerns about edge cases where existing project data could still be affected.

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/session/OAuth2CookieFilter.java
+++ b/datashare-app/src/main/java/org/icij/datashare/session/OAuth2CookieFilter.java
@@ -27,9 +27,7 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
-import java.util.Date;
-import java.util.Map;
-import java.util.Objects;
+import java.util.*;
 import java.util.concurrent.ExecutionException;
 
 import static java.lang.String.format;
@@ -161,7 +159,37 @@ public final class OAuth2CookieFilter extends CookieAuthFilter {
         }
         if (!oauthDefaultProject.isEmpty()) {
             userMap.put("provider", "icij");
-            userMap.put("groups_by_applications", Map.of("datashare", singletonList(oauthDefaultProject)));
+
+            // Check if groups_by_applications already exists in userMap
+            Object groupsByApplicationsObj = userMap.get("groups_by_applications");
+
+            Map<String, Object> groupsByApplications;
+
+            if (groupsByApplicationsObj instanceof Map) {
+                groupsByApplications = (Map<String, Object>) groupsByApplicationsObj;
+            } else {
+                // Initialize a new map if groups_by_applications is missing or is of a wrong type
+                groupsByApplications = new HashMap<>();
+                userMap.put("groups_by_applications", groupsByApplications);
+            }
+
+            // Check if datashare exists in groups_by_applications
+            Object datashareObj = groupsByApplications.get("datashare");
+
+            List<String> datashare;
+
+            if (datashareObj instanceof List) {
+                datashare = (List<String>) datashareObj;
+            } else {
+                // If datashare does not exist or is not a list, create a new one
+                datashare = new ArrayList<>();
+                groupsByApplications.put("datashare", datashare);
+            }
+
+            // Add the default project only if it is not already in the list
+            if (!datashare.contains(oauthDefaultProject)) {
+                datashare.add(oauthDefaultProject);
+            }
         }
         DatashareUser datashareUser = new DatashareUser(userMap);
         writableUsers().saveOrUpdate(datashareUser);


bamthomas:
Hello @neoReuters 

Fist, thanks for your PR. It took us some time to understand why this was done like it is.

So to give you some context:

* the source of the update was probably the api key epic: #504 we wanted to do ACL on projects with the api key
* we stored the users in a redis cache and use this cache to find the project keys
* so the source of "trust" is still the Identity Provider. That's why the user's cache is overridden at each login (the ACL could have been updated in Identity Provider)

Now if I understand your need:

* you put project ACL's for datashare in Redis or a Database 
* you use AWS Cognito as identity provider (you mentioned it in #1393 ) for authenticating your users
* each time they login they erase the project list with the default oauth project

Is that correct?

bamthomas:
so would this story fix your issue? 
#1585 1585

neoReuters:
Hi @bamthomas - your assumptions are correct. Just for further brevity (not sure that it matters); we are using `UsersInDB` `authUsersProvider` over `UsersInRedis`. I tried both initially to see if it was an issue with the `authUsersProvider` but found it was something baked into the `oauth2CookieFilter` class instead.

To my understanding now after you've broken it down - the way the system works; it expects the list of projects to essentially be part of the identity provider's responsibility. So, in a sense, we must send through something like `groups_by_applications.datashare` as part of our token - I'm guessing this is the ideal behaviour?

As a solution then, I guess it would be helpful if we could have a custom key, kind of like what we did with `oauthClaimAttribute`, so that it can be adjusted as such, read when going through the `processOAuthApiResponse` function, and perhaps also it should be documented to help people understand how it works alongside maybe with how you should go about creating new projects (currently this involved us creating projects in the `project` table, which we then mapped to users in the `user_inventory` table) in server mode in general.

The current solution proposed assumed that it was Datashare's responsibility to map projects to users hence why we're adjusting this inside the `user_inventory` table. How do you think we should proceed with this? Open to ideas!

neoReuters:
> so would this story fix your issue? #1585

Ah just replied before seeing your reply, I guess something like this would work fine. That means essentially the identity provider, in our case Cognito, is in charge of enriching the token with projects that user is responsible for, correct?

Funnily enough, before quite recently it was quite tough to do this, but AWS have finally provided a way to enrich tokens with a Lambda trigger (https://aws.amazon.com/about-aws/whats-new/2023/12/amazon-cognito-user-pools-customize-access-tokens/). So I can imagine it working something like this - we have something like a DynamoDB or even a relational database which holds the projects the users are mapped to. Then we fire the lambda to go and check and enrich the token with this user's projects. 

This works well but would also be interested in seeing whether you can **also** possibly provide a solution similar to the code this PR has, for people who might not have the ability to manipulate their oauth tokens? Thoughts @bamthomas?

bamthomas:
sorry for the late answer. 

> would also be interested in seeing whether you can also possibly provide a solution similar to the code this PR has, for people who might not have the ability to manipulate their oauth tokens?

Personally I'd avoid hacking the goups by adding them each time the connection is made on the Identity Provider. It introduces side effects, that is hard to follow and understand afterward (why this user has this group BTW?).

To me, either you can't configure projects in the Identity Provider and you can store them in the redis/postgres database as a project ACL (datashare should not touch the information stored that is going to be changed with #1585), either you can do it in your Identity Provider and then datashare should be configurable to pick the list in a given field.

neoReuters:
Understood, intrigued with how you'll manage that with something like basic auth though? Unless you then also provide an interface for the system to point to to fetch from one source as mentioned.

bamthomas:
so @neoReuters I'm closing this PR if you don't mind.
if you have some issues, don't hesitate to reopen another issue and/or PR.

Thanks for your feedback and fix.

Issue #1585, opened 2024-10-02
OAuth: remove hard coded `groups_by_applications` and users cache from `OAuthCookieFilter`


bamthomas:
**Is your feature request related to a problem? Please describe.**

This is related to the PR #1583 
For projects ACL's and api key we use a user cache in Redis. This cache is overridden at each login. 
The project ACL is used for example by  `DocumentRessource` to check if users have access to a project by `getProjects/getProjectNames` in  [User.java](https://github.com/ICIJ/datashare/blob/main/datashare-api/src/main/java/org/icij/datashare/user/User.java). It  is finally using the field `groups_by_applications` returned by the Identity Provider. 

**Describe the solution you'd like**

1. the cache should be made by another implementation class so the line `writableUsers().saveOrUpdate(datashareUser);` could be in a subclass of the `OAuth2CookieFilter.processOAuthApiResponse` 
2. the user field containing the project could have another location in the json. It is now in `groups_by_applications.datashare` but it could be located elsewhere in the json returned by the Identity Provider. That could allow other deployments to use the IP configuration for project ACL without using another persistence location.

**Additional context**
see 
When cache has been introduced #504.
latest refactor #1395

PR #1593, opened 2024-10-14
Feat: oauth2 parametrize ICIJ identity provider specific values

bamthomas:
This PR isolate ICIJ settings in a new class.
And adds a `--oauthUserProjectsAttribute` option to be able to specify a custom JSON attribute path (dot separated).
see #1585


bamthomas:
would that seem ok for you @neoReuters?

neoReuters:
> Would that seem ok for you @neoReuters?

Looks solid - will give it a spin sometime this or next week when you get this live. Thanks @bamthomas for the work!

Issue #1598, opened 2024-11-05
Gracefull stop of workers


bamthomas:
**Is your feature request related to a problem? Please describe.**

Until now, we used a [poison pill pattern](https://java-design-patterns.com/patterns/poison-pill/) to stop workers. It is used with a blocking queue. So it is 1:1, if we have 2 workers, one worker will get the poison pill and exit, the other won't have the pill and will stay alive.

**Describe the solution you'd like**

We implemented a `WORKER_QUEUE` with AMQP for sending 1:n events (fanout/pubsub mode) from TaskManager to workers. We used it for cancelling tasks.

We could implement a similar message `STOP` or `SHUTDOWN` that would stop all workers.  

It would replace the `POISON` objects that are used to exit the task loop.

There are 4 cases where we want task workers to be gracefully shutdown:

- [x] CLI memory task manager
- [x] CLI AMQP task manager
- [x]  CLI Redis task manager

PR #1622, opened 2024-12-11
Fix: workers shutdown cli

bamthomas:
This PR fixes several issues:

- task CLI should handle the task `GROUP`
- missing singleton annotations that created several instances of services
- the waitForTasks that used an object instance lock with network access. It doesn't work because instance lock is changing at each network retrieval
- split of `shutdownAndAwaitTermination(timeout)` in two methods `shutdown()` and `awaitTermination(timeout)` 

see #1598

### DIFF
--- a/datashare-tasks/src/main/java/org/icij/datashare/asynctasks/TaskManager.java
+++ b/datashare-tasks/src/main/java/org/icij/datashare/asynctasks/TaskManager.java
@@ -12,7 +12,6 @@ import org.slf4j.LoggerFactory;
 import java.io.Closeable;
 import java.io.IOException;
 import java.io.Serializable;
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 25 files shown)


datashare-tasks/src/main/java/org/icij/datashare/asynctasks/TaskManager.java
```diff
@@ -168,11 +187,43 @@ default <V extends Serializable> Task<V> handleAck(TaskEvent e)  {
...
 
+    /**
+     * wait for all the tasks to have a result.
+     *
+     * This method will poll the task list. So if there are a lot of tasks or if tasks are
+     * containing a lot of information, this method call could be very intensive on network and CPU.
+     *
+     * @param timeout amount for the timeout
+     * @param timeUnit unit of the timeout
+     * @return the list of tasks
```
ClemDoum, line 198:
```suggestion
     * @return the list of alive tasks
```

ClemDoum, line 198:
```suggestion
     * @return the list of unfinished/alive tasks
```

ClemDoum (approved):
Very nice I only have a minor suggestion

Issue #1535, opened 2024-08-30
bug: delete all doesn't delete tasks


mvanzalu:
**Describe the bug**
Deleting everything (the red bin) does not delete task related items (Batch Downloads, Scan/index... tasks)

**To Reproduce**
1. Create a batch download or add documents
2. Click on the red bin and delete everything
3. Go to 'Batch Downloads' or 'Analyze your documents
4. See the items are still present

**Expected behavior**
Everything should be deleted

pirhoo:
@mvanzalu @bamthomas I'm removing the stale tag

caro3801:
The work on the cancellable tasks is related #1600 #1492

PR #1636, opened 2025-01-15
fix: delete all incl tasks

mvanzalu:
Related to  #1535 , this PR includes the task manager in the "delete all" functionality. It stops the running tasks and delete these along the QUEUED and CREATED ones.

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/web/ProjectResource.java
+++ b/datashare-app/src/main/java/org/icij/datashare/web/ProjectResource.java
@@ -206,7 +206,7 @@
         @Operation(description = "Deletes all user's projects from database and elasticsearch index.")
         @ApiResponse(responseCode = "204", description = "if projects are deleted")
         @Delete("/")
-        public Payload deleteProjects(Context context) {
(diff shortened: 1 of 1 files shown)


datashare-app/src/main/java/org/icij/datashare/web/ProjectResource.java
```diff
@@ -201,13 +208,21 @@ public Payload deleteProject(String id, Context context) throws IOException {
...
                     deleteProject(project.name, context);
                 } catch (IOException e) {
                     throw new RuntimeException(e);
                 }
             });
+            try {
+                logger.info("Stopping tasks : {}", taskManager.stopAllTasks(user));
+                taskManager.waitTasksToBeDone(TaskManager.POLLING_INTERVAL*2, MILLISECONDS);
+                logger.info("Deleted tasks : {}", !taskManager.clearDoneTasks().isEmpty());
+            } catch (IOException e) {
```
bamthomas, line 223: I'd let the IOException be thrown directly instead of wrapping it in a RuntimeException but maybe I'm missing something?

bamthomas (approved):

PR #1641, opened 2025-01-23
feat: health status task manager

mvanzalu:
Adds TaskManager to health check : 

- TaskManagerMemory : checks if Executor is not shutdown
- TaskManagerRedis : executes a Redis command (fails if impossible)
- TaskManagerAmqp : sends a event in Monitoring queue (queue is deleted within an hour)

### DIFF
--- a/datashare-tasks/src/main/java/org/icij/datashare/asynctasks/bus/amqp/AmqpQueue.java
+++ b/datashare-tasks/src/main/java/org/icij/datashare/asynctasks/bus/amqp/AmqpQueue.java
@@ -22,7 +22,7 @@ public enum AmqpQueue {
 	WORKER_EVENT("exchangeWorkerEvents", BuiltinExchangeType.FANOUT, "routingKeyWorkerEvents"),
 	MONITORING("exchangeMonitoring", BuiltinExchangeType.DIRECT, "routingKeyMonitoring", Map.of(
 			"x-message-ttl", 5000,
-			"x-expires", 10000), null);
+			"x-expires", 3600 * 1000), null);
 
 	public final String exchange;
 	public final String routingKey;


datashare-tasks/src/main/java/org/icij/datashare/asynctasks/bus/amqp/AmqpQueue.java
```diff
@@ -19,7 +19,10 @@ public enum AmqpQueue {
 			"x-consumer-timeout", 3600 * 1000), TASK_DLQ),
 	MANAGER_EVENT_DLQ  ("exchangeDLQManagerEvents",  BuiltinExchangeType.DIRECT,"routingKeyDLQManagerEvents"),
 	MANAGER_EVENT  ("exchangeManagerEvents",  BuiltinExchangeType.DIRECT,"routingKeyManagerEvents", new HashMap<>(), MANAGER_EVENT_DLQ),
-	WORKER_EVENT("exchangeWorkerEvents", BuiltinExchangeType.FANOUT, "routingKeyWorkerEvents");
+	WORKER_EVENT("exchangeWorkerEvents", BuiltinExchangeType.FANOUT, "routingKeyWorkerEvents"),
+	MONITORING("exchangeMonitoring", BuiltinExchangeType.DIRECT, "routingKeyMonitoring", Map.of(
+			"x-message-ttl", 5000,
+			"x-expires", 10000), null);
```
ClemDoum, line 25: I'm wondering if we shouldn't set a longer `x-expires`.
Creating a queue is a "long" operation that we should avoid unless the queue isn't needed any longer.
I'm afraid that recreating the queue for every healtcheck will result in more compute used as well a longer health check latency.
Maybe setting it to 1h will keeping a very low TTL would be a good trade off to balance between the queue not growing too much / deleting it if not used / not recreating it too often

ClemDoum (approved):

PR #1643, opened 2025-01-30
fix: amqp health check

mvanzalu:
Fix RabbitMQ crashing due to excessive AMQP channel openings during health checks

This PR addresses an issue where RabbitMQ was crashing because each health check triggered the opening of a new AMQP channel. Over time, this resulted in an out-of-bounds error due to the accumulation of channels.

Changes made:

- Optimized health check logic to prevent excessive AMQP channel openings for health check.  
- The monitoring channel and queue are created when datashare starts, the queue is never deleted, only the messages are (only last 5 seconds)
- The Amqp Configuration class has a "monitoring" boolean parameter to use in the query of the Amqp adress
- Ensured that channels are only opened by the AmqpInterlocutor class (method's visibility are changed to package)

### DIFF
--- a/datashare-tasks/src/main/java/org/icij/datashare/asynctasks/bus/amqp/AmqpInterlocutor.java
+++ b/datashare-tasks/src/main/java/org/icij/datashare/asynctasks/bus/amqp/AmqpInterlocutor.java
@@ -85,7 +85,7 @@ public class AmqpInterlocutor implements Closeable {
             try {
                 if (AmqpQueue.MONITORING.equals(queue) && configuration.monitoring) {
                     createAmqpChannelForMonitoring(queue);
-                } else {
+                } else if (!AmqpQueue.MONITORING.equals(queue)) {
                     createAmqpChannelForPublish(queue);
                 }
             } catch (IOException e) {


ClemDoum:
I have the feeling that the channel use could be optimized/simplified even further.
For now I have the feeling that each time we call `createAmqpChannelForConsume` or `createAmqpChannelForPublish` we create a new channel via `connection.createChannel()`.

In practice we don't need to create one channel per queue. RabbitMQ [doc](https://www.rabbitmq.com/docs/3.13/channels#lifecycle) says that channels are mean to be long lived objects just like connection and that "limiting the number of channels used per connection is highly recommended."

In Python workers use 1 channel for publishing one for consuming from all queue (I think I could use a single one actually). The TM has a single queue.

Basically we don't need the `AmqpInterlocutor.publishChannels` which maps each queue to a channel, we can just use the same channel and let the routing key hit the right queue.

So I was wondering if would be possible to refactor the `AMQPInterlocutor` APIs to do something like this:
```java
public AmqpInterlocutor(Configuration configuration) throws IOException {
        this.configuration = configuration;
        ConnectionFactory connectionFactory = createConnectionFactory(configuration);
        this.channel = createConnection(connectionFactory).createChannel()
    }
```
and then use the same channel everywhere else :
```java
public void publish(AmqpQueue queue, Event event) throws IOException {
        this.channel.publish(event);
}

public synchronized AmqpInterlocutor initAmqpChannelForPublish(AmqpQueue queue) throws IOException {
        AmqpChannel channel = new AmqpChannel(this.channel, queue);
        channel.initForPublish();
        logger.info("publish channel {} has been created for exchange {}", channel, queue.exchange);
        return this;
}
```

ClemDoum (commented):
I have the feeling that the channel use could be optimized/simplified even further.
For now I have the feeling that each time we call `createAmqpChannelForConsume` or `createAmqpChannelForPublish` we create a new channel via `connection.createChannel()`.

In practice we don't need to create one channel per queue. RabbitMQ [doc](https://www.rabbitmq.com/docs/3.13/channels#lifecycle) says that channels are mean to be long lived objects just like connection and that "limiting the number of channels used per connection is highly recommended."

In Python workers use 1 channel for publishing one for consuming from all queue (I think I could use a single one actually). The TM has a single queue.

Basically we don't need the `AmqpInterlocutor.publishChannels` which maps each queue to a channel, we can just use the same channel and let the routing key hit the right queue.

So I was wondering if would be possible to refactor the `AMQPInterlocutor` APIs to do something like this:
```java
public AmqpInterlocutor(Configuration configuration) throws IOException {
        this.configuration = configuration;
        ConnectionFactory connectionFactory = createConnectionFactory(configuration);
        this.channel = createConnection(connectionFactory).createChannel()
    }
```
and then use the same channel everywhere else :
```java
public void publish(AmqpQueue queue, Event event) throws IOException {
        this.channel.publish(event);
}

public synchronized AmqpInterlocutor initAmqpChannelForPublish(AmqpQueue queue) throws IOException {
        AmqpChannel channel = new AmqpChannel(this.channel, queue);
        channel.initForPublish();
        logger.info("publish channel {} has been created for exchange {}", channel, queue.exchange);
        return this;
}
```

ClemDoum:
That might be beyond the scope of the healthcheck itself, but I think it would be a good the `AMQPInterlocutor` in general.

ClemDoum (approved):

Issue #1733, opened 2025-03-31
spike: delegate task argument and result ser/de to `datashare-app`


ClemDoum:
# Description

Currently task ser/de happens in `datashare-tasks`, this has the main limitation of biding `datashare-tasks` to all tasks arguments and results types.

Any type of task argument or result must be defined inside `datashare-tasks` to be defined. However `datashare-tasks` is meant to be agnostic to applications and should be able to handle any kind of result/args.

A potential solution to support any kind of results/args would be to delegate the ser/de of task args/results to apps. `datashare-tasks` would has `Serializable` args and results and apps would be in charge to define the supported type of results and args. However `datashare-tasks` could still provide utilities for ser/de such as factories/helpers using the types/classes provided by app to create serde tools.

While appealing this idea might have strong implication in the codebase, performing a quick spike could help us get an idea of the implications.

PR #1748, opened 2025-04-09
spike(datashare-task): delegate serde to datashare-app

ClemDoum:
addresses #1733 / #1452


ClemDoum:
closed in favor of #1797

PR #1777, opened 2025-04-18
fix; task manager memory termination

mvanzalu:
This PR fixes Datashare's shutdown which wasn't done properly

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/tasks/TaskManagerAmqp.java
+++ b/datashare-app/src/main/java/org/icij/datashare/tasks/TaskManagerAmqp.java
@@ -7,6 +7,8 @@ import org.icij.datashare.PropertiesProvider;
 import org.icij.datashare.asynctasks.TaskRepository;
 import org.icij.datashare.asynctasks.bus.amqp.AmqpInterlocutor;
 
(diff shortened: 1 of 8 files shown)


datashare-app/src/main/java/org/icij/datashare/tasks/TaskManagerAmqp.java
```diff
@@ -17,6 +17,7 @@ public TaskManagerAmqp(AmqpInterlocutor amqp, TaskRepository taskRepository, Pro
 
     TaskManagerAmqp(AmqpInterlocutor amqp, TaskRepository taskRepository, PropertiesProvider propertiesProvider,
                     Runnable eventCallback) throws IOException {
-        super(amqp, taskRepository, Utils.getRoutingStrategy(propertiesProvider), eventCallback);
+        super(amqp, taskRepository, Utils.getRoutingStrategy(propertiesProvider), eventCallback,
+                Integer.parseInt(propertiesProvider.get("taskManagerPollingInterval").orElse(String.valueOf(DEFAULT_TASK_POLLING_INTERVAL))));
```
ClemDoum, line 21: I would include the unit in the variable names `taskManagerPollingIntervalMilliseconds` and `DEFAULT_TASK_POLLING_INTERVAL_MS` to be consistent with the DatashareCLI naming

ClemDoum (changes requested):

Issue #1687, opened 2025-03-12
Add a nbQueriesNoResults in batchsearches


bamthomas:
**Is your feature request related to a problem? Please describe.**

As for `nbQueries` and `nbResults` we would like to cache the number of queries that have no results.

pirhoo:
Please use `NULL` as default value so we can detect legacy batch searches.

pirhoo:
I changed the name of the new column to `nbQueriesWithoutResults` for consistency.

pirhoo:
I modified the issue title. We should ensure that:

* [ ] when a batch search fails, the task status is set to "fail" as well
* [ ] when a batch search fails, store the "errorMessage" and "errorQuery" in the "result" field of the task
* [ ] when a batch search has results, store "nbResults" in the "result" field (using the same attribute same)

PR #1784, opened 2025-04-29
feat: sync bs failure status and task status

mvanzalu:
Related #1687 
This PR aims to synchronize the batch search status and the task status when the batch search fails.

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/tasks/BatchSearchRunner.java
+++ b/datashare-app/src/main/java/org/icij/datashare/tasks/BatchSearchRunner.java
@@ -4,14 +4,9 @@ import co.elastic.clients.elasticsearch._types.ElasticsearchException;
 import com.google.inject.Inject;
 import com.google.inject.assistedinject.Assisted;
 import jakarta.json.JsonException;
-import java.util.concurrent.Callable;
-import java.util.function.Function;
 import org.icij.datashare.Entity;
 import org.icij.datashare.PropertiesProvider;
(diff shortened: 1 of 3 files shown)


datashare-app/src/main/java/org/icij/datashare/tasks/BatchSearchRunner.java
```diff
@@ -156,11 +144,15 @@ public Integer call() throws Exception {
             logger.info("done batch search {} with success", batchSearch.uuid);
         } catch (ElasticsearchException esEx) {
             logger.error("ES exception while running batch " + taskView.id, esEx);
-            repository.setState(taskView.id, new SearchException(query,
-                    ElasticSearchAdapterException.createFrom(esEx)));
+            SearchException searchException = new SearchException(query,
+                    ElasticSearchAdapterException.createFrom(esEx));
+            repository.setState(taskView.id, searchException);
```
ClemDoum, line 149: I'm wondering if the `setState` is even necessary.
For other tasks, exceptions which are thrown at caught at the `TaskWorkerLoop` level which takes care of handling the error properly. Task status handling shouldn't happen in task themselves, but I might be missing something

ClemDoum: Ok i get it, the `repository` is a `BatchSearchRepository` and not a `TaskRepository`

ClemDoum (approved):

PR #1795, opened 2025-05-02
feat: batch searches proxy

pirhoo:
This PR add retro-compatibility for listing batch searches without tasks (#1794). To address this problem, the backend create on the fly "BatchSearchRunnerProxy" tasks that are merged with the actual list of tasks.

This also has the benefit to add access control to the batch searches tasks belonging to other user in server mode. Because the list of tasks is built based on the batch searches a user has access to, this create a way for them to see tasks they did not authored.

The frontend was already modified to support this new type of task.

### DIFF
--- a/datashare-tasks/src/main/java/org/icij/datashare/asynctasks/TaskManager.java
+++ b/datashare-tasks/src/main/java/org/icij/datashare/asynctasks/TaskManager.java
@@ -1,10 +1,12 @@
 package org.icij.datashare.asynctasks;
 
+import org.apache.commons.collections4.ListUtils;
 import org.icij.datashare.asynctasks.bus.amqp.CancelledEvent;
 import org.icij.datashare.asynctasks.bus.amqp.ErrorEvent;
 import org.icij.datashare.asynctasks.bus.amqp.ProgressEvent;
 import org.icij.datashare.asynctasks.bus.amqp.ResultEvent;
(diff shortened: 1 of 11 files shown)


datashare-tasks/src/main/java/org/icij/datashare/asynctasks/TaskManager.java
```diff
@@ -58,6 +65,35 @@ default List<Task<?>> getTasks(User user, Map<String, Pattern> filters, WebQuery
         taskStream = getFilteredTaskStream(filters, taskStream);
         return taskStream.filter(t -> user.equals(t.getUser())).skip(pagination.from).limit(pagination.size).collect(toList());
     }
+
+    default List<Task<?>> getTasks(User user, Map<String, Pattern> filters, WebQueryPagination pagination, List<BatchSearchRecord> batchSearchRecords) throws IOException {
```
ClemDoum, line 69: I'm wondering if the addition of this API would be better located in `datashare-app` since it's really an applicative API/patch and not a generic `datashare-tasks` API.

You could probably add a `DatashareTaskManager` interface in `datashare-app` extending `TaskManager` and then use this class in `datashare-app` instead of the `TaskManager`

pirhoo: Thanks for the suggestion! As agreed, this should be done in another refactoring PR.

PR #1802, opened 2025-05-12
chore: refactor task manager to return stream of tasks

ClemDoum:
# PR description

Currently each time we filter tasks we always materialize them in a list which contains all tasks. As we run more tasks to Datashare this list can grow very large, using a stream API will help reduce the memory footprint of task search.

# Changes

### Changed
- change `TaskManager` APIs to return task streams

### DIFF
--- a/datashare-tasks/src/main/java/org/icij/datashare/asynctasks/TaskManager.java
+++ b/datashare-tasks/src/main/java/org/icij/datashare/asynctasks/TaskManager.java
@@ -1,6 +1,6 @@
 package org.icij.datashare.asynctasks;
 
-import org.apache.commons.collections4.ListUtils;
+import org.icij.datashare.Entity;
 import org.icij.datashare.asynctasks.bus.amqp.CancelledEvent;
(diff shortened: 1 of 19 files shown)


datashare-tasks/src/main/java/org/icij/datashare/asynctasks/TaskManager.java
```diff
@@ -41,7 +38,8 @@ public interface TaskManager extends Closeable {
     <V extends Serializable> Task<V> clearTask(String taskId) throws IOException, UnknownTask;
     boolean stopTask(String taskId) throws IOException, UnknownTask;
 
-    List<Task<?>> getTasks() throws IOException;
+    Stream<Task<?>> getTasks() throws IOException;
+    // clearDoneTasks keeps a List return type otherwise tasks are cleared unless the stream is consumed
```
caro3801, line 42: thanks for this comment, it's helpful to understand while reviewing

caro3801 (approved):
Great refactor by changing tasks from list to stream. Thanks

PR #1807, opened 2025-05-13
chore: unify and optimize task search APIs (`TaskManager.getTasks`)

### LINKED ISSUE #1803: feat: add regex on arbitrary filter with multiple values on tasks
related to #1801

ClemDoum:
# TODO
- [x] merge #1802 (the branch is based on #1802)

# Description

Fixes #1803 (as a side effect of the refactoring)

Currently the task search APIs support arbitrary filters which in practice are not used arbitrarily.
The frontend only searches by:
- user ID
- arg value
- task name

On the other hand, supporting arbitrary filters negatively impacts task search, indeed to perform this arbitrary search we:
1. fecth **all task records** from the DB
2. convert them to JSON strings
3. apply the filter on the JSON string

This approach won't scale to many task records.

In contrast, this PR refactors the current task search API, unifies it using a common `TaskManager.getTask(TaskFilters filters)` API supporting all current task searchs (+ searching by task state).

In addition, the `TaskManager.getTask(TaskFilters filters)` API is implemented in each underlying `TaskRepository` API using the dedicated search tools offered by the different implements.

In particular when using SQL, we perform plain SQL queries rather than fecthing all records from the DB and then filter them in Java.

# Changes

### Added
- created a `TaskFilters`. This class is used for user filter validation and task stream filtering 

### Changed
-  unified all task search APIs using `TaskManager.getTasks(TaskFilters filters)` and `TaskRepository(TaskFilters filters)`. By default `TaskManager.getTasks(TaskFilters filters)` filters the result of `TaskManager.getTasks()` using the `TaskFilters::filter`. However the implementer can decide to override this default logic if the repo DB implementation supports more efficient search.
- remove the inheritance between `TaskRepository` and `Map<String, TaskMetadata>` forcing to implement unused `Map` methods (kept only the ones used for task search)
 

### Added
- created a `taskFiltersFromContext` helper to build task filters from the HTTP query context

### DIFF
--- a/datashare-api/pom.xml
+++ b/datashare-api/pom.xml
@@ -11,7 +11,7 @@
     </parent>
 
     <artifactId>datashare-api</artifactId>
-    <version>15.0.0</version>
+    <version>14.6.0</version>
     <packaging>jar</packaging>
 
     <dependencies>
--- a/datashare-api/pom.xml
+++ b/datashare-api/pom.xml
@@ -11,7 +11,7 @@
     </parent>
 
     <artifactId>datashare-api</artifactId>
-    <version>14.5.0</version>
+    <version>15.0.0</version>
     <packaging>jar</packaging>
 
     <dependencies>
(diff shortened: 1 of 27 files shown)


caro3801 (approved):
Looks much cleaner and efficient, thanks. The frontend api call to /tasks should be checked before releasing

pirhoo:
Very nice refactoring, thanks!

PR #1873, opened 2025-08-12
spike: SCAN_INDEX_NLP workflow

ClemDoum:


ClemDoum:
Conductor serializes inputs and outputs as `Map<String, Object>` where each `Object` has to be serializable using plain JSON data types (float, string, int etc...). 

It delegate the responsability of serialization and deserialization to the task implementer a bit like here: https://github.com/ICIJ/datashare/pull/1810

ClemDoum:
I haven't looked at the `Main` of the conductor server to see if it could be embedded in local mode. 

I saw that they made a [`lite` version of the app](https://github.com/conductor-oss/conductor/blob/main/server-lite/src/main/java/org/conductoross/conductor/Conductor.java#L33) (removing UI routes for instance), i don't know if we could embed it

ClemDoum:
@bamthomas to be more precise on the serialization, from [what I can see in this utils](https://github.com/conductor-oss/conductor/blob/6147d61d1babf47f5a0a328d114f1eb5d3d5ecb1/common/src/main/java/com/netflix/conductor/common/utils/SummaryUtil.java#L51) you can either:
- serialize objects as plain JSON strings
- serialize objects as protobuf binaries

Issue #1890, opened 2025-09-09
feat: enable/disable OCR with a query param when using pagination API


pirhoo:
As a user i want to be able to retrieve pagination of a document wether it has been OCRed or not during indexing

- [x] add query param to enable/disable OCR on pagination end point
- [x] store in ES if a doc has been OCRed during indexing

caro3801:
This is used when the document has been indexed without OCR. Therefore the pagination won't retrieve OCR indices

PR #1927, opened 2025-10-20
feat: retrieve pagination according to document OCR parsing 

caro3801:
Related to #1890 
* Update to extract 9.4.12 which adds the new property to doc if they are using OCR parsing ("ocr_parser") or not and fix breaking changes
* When retrieving doc pages and doc pages indices, check if doc has been indexed with OCR. If not disable OCR during pagination to prevent inconsistency between the doc and its pagination


mvanzalu (approved):

pirhoo (approved):
LGTM! Well done!

PR #1943, opened 2025-11-18
fix: checks duplicate document size

mvanzalu:


datashare-app/src/main/java/org/icij/datashare/utils/DocumentVerifier.java
```diff
@@ -39,7 +41,13 @@ public boolean isRootDocumentSizeAllowed(Document document) {
...
         }
         long maxSizeBytes = getEmbeddedDocumentDownloadMaxSizeBytes();
-        Document rootDocument = indexer.get(document.getProjectId(), document.getRootDocument());
+        Document rootDocument;
+        Entity entity = indexer.get(document.getProjectId(), document.getRootDocument());
+        if (entity.getClass().equals(Duplicate.class)) {
+            rootDocument = indexer.get(document.getProjectId(), ((Duplicate) entity).documentId);
+        } else {
+            rootDocument = (Document) entity;
+        }
```
pirhoo: This works but it would be cleaner to move it to a `getRootDocument` method that returns a Document instance.

datashare-app/src/main/java/org/icij/datashare/utils/DocumentVerifier.java
```diff
@@ -60,4 +54,18 @@ private long getEmbeddedDocumentDownloadMaxSizeBytes() {
...
+     *
+     * @return The max size in bytes.
+     */
+    private Document getRootDocument(Document document) {
+        Entity entity = indexer.get(document.getProjectId(), document.getRootDocument());
+        if (entity.getClass().equals(Duplicate.class)) {
+            return indexer.get(document.getProjectId(), ((Duplicate) entity).documentId);
+        } else {
+            return (Document) entity;
+        }
```
pirhoo: Thanks for the change! We also generally avoid to use else after return (for instance in pylint it's the rule called [`no-else-return`](https://pheanex.github.io/pylint/#R1705)). So this would be more elegant:


```suggestion
        }
        return (Document) entity;
```

PR #2053, opened 2026-03-02
refactor(datashare-tasks): create namespace and custom search attributes at app startup

ClemDoum:
### Changed
- removed namespace related logic from `TaskManagerTemporal` and create a  `TemporalInterlocutor` mirroring the `AmqpInterlocutor` logic (but for temporal). The interlocutor, creates and sets up the temporal DS namespace

### Changed
- set up temporal namespace at app startup


ClemDoum:
@caro3801 I understand ! So the `TemporalInterlocutor` is eventually meant to handle the transport/connection setup to temporal.

At the moment I'm not sure we need to cover it with unit test since if you don't get the transport right, all other tests will fail quite obviously. I'm not saying there should not be test in this class eventually, but at the moment a bit like with the `TaskManagerAmqp` and `AmqpInterlocutor` most of the testing logic happens the TaskManager side

Issue #2100, opened 2026-04-02
feat: inform user when the size limit of a BatchDownload was reached


lschim:
Currently, when the batchDownload of a user reaches the size limit (currently set at 100MB), the user is not notified, and thus, some documents that were in his/her search in the UI are not part of the download, without him/her knowing it.

It would be great for the user to at least know that some of the files he/she was expecting were not part of the batchDownload.

This information will be known by the server when processing the batchDownload and adding each file one by one, so the information can probably not be displayed anywhere else than the batchDownload list screen.

PR #2103, opened 2026-04-02
feat: add truncation reason

lschim:
Back-end part of #2100 

### Modifications in the API
⚠️ This PR introduces a small change in the API. When returning Tasks, the field `@type` of `result` field for a BatchDownload is now `BatchDownloadRunnerResult`, when it was previously `UriResult`. A new optionnal field was also added `truncationReason`. When present, its values can be : `SIZE_LIMIT`,`FILE_COUNT_LIMIT` or `UNKNOWN`. If absent it means no truncation was made (there was no difference between the number of files in the search and in the generated archive)

### Extract of value returned by GET /api/task

```json
"result": {
  "value": {
    "@type": "BatchDownloadRunnerResult",
    "uri": "file:///Users/jack/.local/share/datashare/tmp/archive_louis_2026-04-02T15_04_46.183Z%5BGMT%5D.zip",
    "size": 129545557,
    "truncationReason": "SIZE_LIMIT"
  }
}
```
when it was previously
```json
"result": {
  "value": {
    "@type": "UriResult",
    "uri": "file:///Users/jack/.local/share/datashare/tmp/archive_louis_2026-04-02T15_04_46.183Z%5BGMT%5D.zip",
    "size": 129545557
  }
}
```

### AI use
I used Claude to help me check the SQL script of migration

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/tasks/BatchDownloadRunner.java
+++ b/datashare-app/src/main/java/org/icij/datashare/tasks/BatchDownloadRunner.java
@@ -15,7 +15,6 @@ import org.icij.datashare.asynctasks.CancelException;
 import org.icij.datashare.asynctasks.CancellableTask;
 import org.icij.datashare.asynctasks.Task;
 import org.icij.datashare.asynctasks.TaskGroup;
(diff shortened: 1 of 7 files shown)


pirhoo (approved):
LGTM, thanks!

Issue #1409, opened 2024-06-12
refactor: CLI with subcommands


bamthomas:
**Is your feature request related to a problem? Please describe.**

By now datashare CLI is implicitly handling subcommands:
- index
- scan
 - NER extraction
 - NER extraction server
 - HTTP server
 - CLI services   

But they are mixed in the numerous CLI options with inconsistencies.

**Describe the solution you'd like**

I would be convenient to have an exhaustive list of these subcommands and a specific help for each of them with options, 

**Additional context**

It seems that it is [not possible with joptsimple](https://github.com/jopt-simple/jopt-simple/issues/102). There is a list of command line parsers [here](http://jewelcli.lexicalscope.com/related.html)

We had considered [JCommander](http://jcommander.org/) in previous discussions.

PR #2104, opened 2026-04-02
refactor: new CLI with picocli

pirhoo:
As part of #1409, this PR introduces a picocli-based subcommand CLI architecture to replace the flat, flag-based jopt-simple interface. The legacy jopt-simple path is fully preserved and routed via `Main.isLegacyInvocation()`, ensuring backward compatibility during the transition. Several code quality improvements are included alongside the new CLI infrastructure.

## Glossary

- **Root command** — top-level entry point (`datashare`)
- **Subcommand** — named child of a parent command (`app`, `stage`, `plugin`)
- **Leaf command** — deepest subcommand, the one that executes (`start`, `run`, `list`)
- **Mixin** — reusable option group shared across commands (`GlobalOptions`, `ServerOptions`)
- **Option** — named flag with `--` prefix (`--dataDir`, `-d`)
- **Parameter** — positional argument without `--` (`<user>`, `<id>`)
- **Scope** — controls option inheritance; `ScopeType.INHERIT` propagates to all subcommands
- **Execution strategy** — decides which command runs after parsing; walks to the deepest leaf
- **Legacy invocation** — backward-compatible jopt-simple path, used when first arg starts with `-`

## To-do after merging

* [ ] https://github.com/ICIJ/datashare-docs/pull/12
* [ ] https://github.com/ICIJ/datashare-installer/pull/24

## Changes

- **feat:** picocli subcommand hierarchy (`app`, `worker`, `stage`, `plugin`, `extension`, `api-key`)
- **feat:** dual-path invocation with routes between jopt-simple and picocli
- **feat:** global options accepted at any argument position via `ScopeType.INHERIT` on `GlobalOptions`
- **feat:** custom help display with `DatashareHelpFactory`
- **refactor:** split `Main.main()` into focused private methods
- **refactor:** extract shared `putIfNotNull` to `DatashareOptions`
- **refactor:** add missing `OAUTH_CLAIM_ID_ATTRIBUTE_OPT` constant
- **fix:** resource leak in `DatashareVersionProvider` (try-with-resources)
- **fix:** empty `defaultProject` now falls back to `local-datashare` instead of passing through
- **test:** global option position
- **test:** option name conflict detection
- **test:** empty `defaultProject` fallback
- **test:** legacy invocation routing

## Breaking changes

The new subcommand interface is not yet the default path (the legacy invocation is still selected automatically for existing callers). However, users who integrate with the CLI programmatically or parse its output should be aware:

- The new picocli path produces a structurally different help output (two-column layout, separate Global Options section) compared to the jopt-simple `--help` output
- Subcommand names (`app start`, `worker run`, `stage run`, `plugin list/install/delete`, etc.) are new entry points that do not exist in the legacy interface; tooling that wraps the CLI will need updating when the legacy path is eventually removed

### Invocation routing

`Main.main()` is now a two-liner that delegates to `resolveProperties()`, which picks the right parser:

```mermaid
flowchart TD
    A["./datashare &lt;args&gt;"] --> B{isLegacyInvocation?}

    B -->|"no args\nfirst arg starts with -\nfirst arg not a known subcommand"| C[Legacy path\nDatashareCli / jopt-simple]
    B -->|"first arg is app / worker /\nstage / plugin / extension /\napi-key / help"| D[Picocli path\nrunPicocli]

    C --> E[cli.properties]
    D --> F[cmd.collectProperties]
    E --> G[startApplication]
    F --> G
```

### Subcommand tree

```
datashare [GLOBAL OPTIONS]
├── app
│   └── start      ← AppServeCommand    (mode, HTTP server, OAuth, batch…)
├── worker
│   └── run        ← WorkerRunCommand   (task worker)
├── stage
│   └── run        ← StageRunCommand    (pipeline stages)
├── plugin
│   ├── list       ← PluginListCommand
│   ├── install    ← PluginInstallCommand
│   └── delete     ← PluginDeleteCommand
├── extension
│   ├── list       ← ExtensionListCommand
│   ├── install    ← ExtensionInstallCommand
│   └── delete     ← ExtensionDeleteCommand
├── api-key
│   ├── create     ← ApiKeyCreateCommand
│   ├── get        ← ApiKeyGetCommand
│   └── delete     ← ApiKeyDeleteCommand
└── help
```

Parent commands (`app`, `worker`, etc.) print usage when invoked without a subcommand. All leaf commands implement both `Runnable` and `DatashareSubcommand`.

### Global options and position independence

Every option in `GlobalOptions` carries `scope = ScopeType.INHERIT`. picocli propagates them down to every leaf command spec, so they are accepted at any position:

```
datashare --elasticsearchAddress http://es:9200 app start
datashare app start --elasticsearchAddress http://es:9200   # identical result
```

`DatashareHelpFactory` filters `o.inherited()` from each subcommand's own option list to prevent duplicate rendering, then renders inherited options in a separate **Global Options** section.

### Help rendering

`DatashareHelpFactory` applies a yarn-inspired style to every command in the tree:

| Section | Content |
|---|---|
| Synopsis | Abbreviated (`[OPTIONS]`) |
| Description | Plain text; lines ending `:` are auto-bolded as headings |
| Arguments | Positional parameters (leaf commands only) |
| Options | Command-specific options (leaf commands only, inherited excluded) |
| Global Options | Options from `GlobalOptions` (leaf subcommands only) |
| Commands | Subcommand list (parent commands only) |

Column widths are sized to the widest option label across both the Options and Global Options sections so both blocks align. Options are sorted alphabetically by long name.

### Shell script routing

The `datashare` launch script detects the invocation style from the first argument:

```
┌───────────────────────────────────────────────────────────────────┐
│  first arg = app / worker / stage / plugin / extension / api-key  │
│                          │                                        │
│           ┌──────────────┴──────────────┐                         │
│     Subcommand path               Legacy path                     │
│     inject ES paths               inject all defaults             │
│     pass "$@" to Java             pass "$@" to Java               │
└───────────────────────────────────────────────────────────────────┘
```

Both paths inject `--elasticsearchPath`, `--elasticsearchSettings`, and `--elasticsearchDataPath` from shell-resolved defaults (`$DATASHARE_HOME`). User-provided options are appended last and override the injected defaults (picocli's `setOverwrittenOptionsAllowed(true)`). ES setup runs for `--mode EMBEDDED` in both paths.

## AI Disclosure

This branch was developed with Claude. AI assisted with:

- Scaffolding the picocli subcommand classes (commands, options, help factory)
- Refactoring Main.java into focused methods and DatashareHelpFactory.apply() for readability
- Writing tests in DatashareCommandTest, RetroCompatibilityTest, MainTest, LegacyInvocationTest, exit-code and parity tests
- Shell script fixes in subcommand detection, ES setup skip logic, help routing
- Code quality by converting String options to enums, extracting DatashareOptions helpers (put, putIfNotNull, putIfTrue, putAll), replacing hardcoded strings with constants
- Squashing the commit history from 25 iterative commits to 4 logical ones

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/Main.java
+++ b/datashare-app/src/main/java/org/icij/datashare/Main.java
@@ -124,19 +124,17 @@ public class Main {
     }
 
     /**
-     * Returns true (use legacy jopt-simple path) when: no args are given, the first arg is
-     * blank, the first arg starts with - (flag-style), or the first arg is not a recognized
-     * subcommand name. Returns false (use picocli path) only when the first arg exactly
-     * matches a known subcommand.
+     * Returns true (use legacy jopt-simple path) when no argument matches a recognized
+     * subcommand name. Returns false (use picocli path) when any argument exactly matches
+     * a known subcommand. This allows parent-command options (e.g. --elasticsearchPath)
+     * to appear before the subcommand name in the args array.
      */
     static boolean isLegacyInvocation(String[] args) {
-        if (args.length == 0) {
-            return true;
+        for (String arg : args) {
+            if (SUBCOMMAND_NAMES.contains(arg.trim())) {
+                return false;
+            }
         }
-        String first = args[0].trim();
-        if (first.isEmpty() || first.startsWith("-")) {
-            return true;
-        }
-        return !SUBCOMMAND_NAMES.contains(first);
+        return true;
     }
 }
--- a/datashare-app/src/main/java/org/icij/datashare/Main.java
+++ b/datashare-app/src/main/java/org/icij/datashare/Main.java
@@ -71,9 +71,9 @@ public class Main {
     }
 
     private static void applyColorScheme(CommandLine commandLine, String[] args) {
-        boolean noColor = System.getenv("NO_COLOR") != null
+        boolean isColorDisabled = System.getenv("NO_COLOR") != null
                 || java.util.Arrays.asList(args).contains("--no-color");
-        if (noColor) {
+        if (isColorDisabled) {
             commandLine.setColorScheme(CommandLine.Help.defaultColorScheme(CommandLine.Help.Ansi.OFF));
         }
     }
--- a/datashare-app/src/main/java/org/icij/datashare/Main.java
+++ b/datashare-app/src/main/java/org/icij/datashare/Main.java
@@ -35,53 +35,68 @@ public class Main {
      * parser based on isLegacyInvocation(), then starts the appropriate mode.
      */
     public static void main(String[] args) throws Exception {
-        Properties properties;
+        Properties properties = resolveProperties(args);
+        startApplication(properties);
+    }
 
+    private static Properties resolveProperties(String[] args) throws Exception {
         if (isLegacyInvocation(args)) {
             DatashareCli cli = new DatashareCli().parseArguments(args);
-            properties = cli.properties;
-        } else {
-            DatashareCommand cmd = new DatashareCommand();
-            CommandLine commandLine = DatashareHelpFactory.configure(new CommandLine(cmd));
-            commandLine.setOverwrittenOptionsAllowed(true);
-            commandLine.setAbbreviatedOptionsAllowed(false);
-            commandLine.setAbbreviatedSubcommandsAllowed(false);
-            commandLine.setCaseInsensitiveEnumValuesAllowed(true);
-            boolean noColor = System.getenv("NO_COLOR") != null
-                    || java.util.Arrays.asList(args).contains("--no-color");
-            if (noColor) {
-                commandLine.setColorScheme(CommandLine.Help.defaultColorScheme(CommandLine.Help.Ansi.OFF));
-            }
-            // The execution strategy works in two phases:
-            // 1. It scans the parse-result hierarchy for any --help or --version request and
-            //    delegates to RunLast if found, so the correct subcommand-level help page is shown.
-            // 2. It walks down to the deepest executed subcommand in the hierarchy (mirroring
-            //    picocli's RunLast behavior) and registers it on DatashareCommand so its
-            //    properties can be collected after execution.
-            commandLine.setExecutionStrategy(parseResult -> {
-                // Walk the parse-result hierarchy to the deepest subcommand, short-circuiting
-                // immediately if any level has requested help/version so RunLast shows the
-                // right help page.
-                CommandLine.ParseResult current = parseResult;
-                while (current != null) {
-                    if (current.isUsageHelpRequested() || current.isVersionHelpRequested()) {
-                        return new CommandLine.RunLast().execute(parseResult);
-                    }
-                    if (!current.hasSubcommand()) break;
-                    current = current.subcommand();
-                }
-                Object userObject = current.commandSpec().userObject();
-                if (userObject instanceof DatashareSubcommand) {
-                    cmd.setExecutedSubcommand((DatashareSubcommand) userObject);
-                }
-                return new CommandLine.RunLast().execute(parseResult);
-            });
-            int exitCode = commandLine.execute(args);
-            if (exitCode != 0) System.exit(exitCode);
-            if (cmd.getExecutedSubcommand() == null) System.exit(0);
-            properties = cmd.collectProperties();
+            return cli.properties;
         }
+        return runPicocli(args);
+    }
 
+    private static Properties runPicocli(String[] args) {
+        DatashareCommand cmd = new DatashareCommand();
+        CommandLine commandLine = DatashareHelpFactory.configure(new CommandLine(cmd));
+        commandLine.setOverwrittenOptionsAllowed(true);
+        commandLine.setAbbreviatedOptionsAllowed(false);
+        commandLine.setAbbreviatedSubcommandsAllowed(false);
+        commandLine.setCaseInsensitiveEnumValuesAllowed(true);
+        applyColorScheme(commandLine, args);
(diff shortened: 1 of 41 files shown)


datashare-cli/src/main/java/org/icij/datashare/cli/command/ServerOptions.java
```diff
@@ -0,0 +1,218 @@
...
+
+    @Option(names = {"--sessionTtlSeconds"}, description = "Session TTL in seconds", defaultValue = "43200")
+    int sessionTtlSeconds;
+
+    @Option(names = {"--sessionStoreType"}, description = "Session store type", defaultValue = "MEMORY")
+    QueueType sessionStoreType;
+
+    // Batch / Download
+    @Option(names = {"--batchQueueType"}, description = "Batch queue type", defaultValue = "MEMORY")
+    QueueType batchQueueType;
```
lschim, line 93: Question : That's not the same as "busType" in GlobalOptions ? I also feel it should actually be an option of PipelineOptions, which is mixed in in this options set, so it could be removed from here

lschim (commented):
praise: That's an amazing PR so far ! I reviewed only the first commit, but thanks for all the little details, all documentation and comments.

thought: The commits are big to review, so maybe they could be smaller (for instance scaffold + one command or one set of commands per commit) but for the history maybe that's interesting to see the change from one version to the other quickly.

suggestion : Maybe the commands and subcommands could be in separate packages (for instance command>apikey> ApiKey*Command) ?

datashare-cli/src/main/java/org/icij/datashare/cli/command/ServerOptions.java
pirhoo: `batchQueueType` and `busType` aren't the same actually, even though they look similar (both are QueueType):

- `busType`: backend for the message/data bus (event publishing)
- `batchQueueType`: backend for the batch task queue / `TaskManager`. It's read in `CommonMode.runWorkers()` and `shouldRunWorker()` to pick between TaskManagerRedis / AMQP / Memory, and to decide whether to spin up Temporal workers.

So they address different concerns and aren't interchangeable.

On moving it to `PipelineOptions`, I'd actually push back a bit on that one. PipelineOptions is scoped to doc/NLP/OCR pipeline stuff, whereas batchQueueType is purely worker/task-manager wiring.

I think the better home is WorkerOptions as it's already mixed into ServerOptions, and batchQueueType is basically the missing sibling of taskRepositoryType / taskRoutingStrategy. Bonus: the worker run subcommand actually needs it too (today it only gets it because CommonMode reads the property directly), so putting it in WorkerOptions makes that explicit.

Either way, agreed it shouldn't stay in ServerOptions because it's not server-specific.

pirhoo:
@lschim thanks for the review!

Yeah I think moving command to packages is a reasonable idea. The package has many files now and the grouping is already pretty obvious, so subpackages would make navigation nicer.

However single-subcommand groups wouldn't really benefit. For `app` (just AppServeCommand), `stage` (StageRunCommand), `worker` (WorkerRunCommand), moving each into their own package is mostly noise. Maybe only split the ones that actually have multiple leaves (apikey, extension, plugin), or just split everything for consistency. Both are defensible.

I'd lean toward a separate PR. The current branch is already huge, across 41 files introducing the whole picocli infrastructure. Adding a package shuffle on top would bloat the diff with pure file moves that reviewers have to mentally filter out from the actual behavior changes.

datashare-app/src/main/java/org/icij/datashare/Main.java
```diff
@@ -2,42 +2,141 @@
...
+/**
+ * Datashare supports two CLI invocation styles: legacy (jopt-simple) and new picocli subcommands.
+ * The legacy path is used when: no args are provided, the first arg starts with -, or the
+ * first arg is an unknown word (i.e. not a recognized subcommand).
+ * The new picocli path is used when the first arg is a known subcommand: app, worker,
+ * stage, plugin, extension, api-key, or help.
+ */
 public class Main {
     private static final Logger LOGGER = LoggerFactory.getLogger(Main.class);
+    private static final Set<String> SUBCOMMAND_NAMES = Set.of("app", "worker", "stage", "plugin", "extension", "api-key", "help");
```
bamthomas: Could that be an Enum ?

bamthomas (commented):
Great !
It will be much more convenient now.

datashare-app/src/main/java/org/icij/datashare/Main.java
pirhoo: Yes, but an enum would just move the duplication: those seven names are already declared as `@Command(name = ...)` on the subcommand classes registered in DatashareCommand (datashare-cli/src/main/java/org/icij/datashare/cli/command/DatashareCommand.java:12-20). An `enum Subcommand { APP("app"), WORKER("worker"), ... }` would be a third place where the same strings live.

A better fix, if the goal is to remove the hardcoded list: ask picocli directly, since it already knows. Build a CommandLine once and read its registered subcommands:

```java
  private static final Set<String> SUBCOMMAND_NAMES =
      new CommandLine(new DatashareCommand()).getSubcommands().keySet();
```

That makes DatashareCommand's `subcommands = {...}` the single source of truth. For instance if we later add FooCommand.class there and isLegacyInvocation picks it up automatically, no second list to keep in sync.

How does that sound?

pirhoo: Done in https://github.com/ICIJ/datashare/pull/2104/changes/6616c644ecc265c99504b704f5572d105a7a68de

lschim (approved):

PR #2113, opened 2026-04-14
Refactor/task manager

lschim:
### Context

  DatashareTaskManager was an interface extending TaskManager with a single default method that reconciled tasks with BatchSearchRecords, only implemented by 4 wrapper classes (TaskManagerMemory, TaskManagerRedis, TaskManagerAmqp, TaskManagerTemporal) which duplicated their datashare-tasks
  counterparts solely to add @Singleton, @Inject.

  At the same time, TaskResource was directly handling business logic: fetching tasks, falling back to BatchSearchRecords when a task wasn't found, and merging both streams — none of which belongs in the HTTP layer.

###  Changes

  - Introduces TaskService as a service layer between TaskResource and TaskManager:
    - getVisibleTasksFor(User, TaskFilters) merges tasks from TaskManager with BatchSearchRecords converted to tasks
    - getVisibleTaskFor(User, String id) encapsulates the fallback to BatchSearchRecords on id lookup
  - Removes the DatashareTaskManager interface and its 4 wrapper implementations in datashare-app
  - Core classes from datashare-tasks are now bound directly in CommonMode via @Provides @Singleton methods
  - All injection sites previously using DatashareTaskManager now use TaskManager
  - TaskResource delegates to TaskService and only handles HTTP translation (UnknownTask → 404)
  
 ### AI usage 
 Used Claude to generate this message, and used ralph-tui to be exhaustive on the removal of injections of services (replace DatashareTaskManager by TaskManager in classes and in tests)


datashare-app/src/main/java/org/icij/datashare/mode/CommonMode.java
```diff
@@ -39,14 +39,6 @@
 import org.icij.datashare.session.UsersInDb;
 import org.icij.datashare.session.UsersWritable;
 import org.icij.datashare.tasks.*;
```
ClemDoum, line 41: I think wildcard are not supposed to be used. We had trouble with them in the past.

datashare-app/src/main/java/org/icij/datashare/mode/CommonMode.java
```diff
@@ -226,6 +219,58 @@ TemporalInterlocutor provideTemporal(final PropertiesProvider propertiesProvider
         return new TemporalInterlocutor(propertiesProvider);
     }
 
+    @Provides @Singleton
```
ClemDoum, line 222: This is a nice refactoring and avoids duplicating classes !

datashare-app/src/main/java/org/icij/datashare/tasks/TaskFinder.java
```diff
@@ -0,0 +1,119 @@
+package org.icij.datashare.tasks;
```
ClemDoum, line 1: Super nice

ClemDoum (approved):

datashare-app/src/main/java/org/icij/datashare/mode/CommonMode.java
lschim: I strongly agree that wildcards shouldn't be used ! I didn't introduce that change though, it was already like that. I think it raises the question to use a linter :)

Issue #2122, opened 2026-04-21
feat(cli): manage users and projects


pirhoo:
Here is the list of commands:

```
datashare user create [<uid>]
datashare user delete [<uid>]
datashare project create [<project>]
datashare project delete [<project>]
datashare project grant  [<project> <uid>]
datashare project revoke [<project> <uid>]
```

Conventions following [clig.dev](https://clig.dev/):

- **Interactive by default.** Missing required fields are prompted when stdin is a TTY.
- **`--no-input`** disables prompts and exits 2 if any required field is missing (for scripts/CI).
- **`--yes` / `-y`** skips confirmation on destructive actions. `--no-input` without `--yes` on a destructive command exits 2.
- **Exit codes**: 0 success, 1 runtime error, 2 misuse, 3 not found, 4 conflict, 5 validation.
- **Idempotency**: `create` accepts `--if-not-exists`; `delete` and `revoke` accept `--if-exists`. `grant` on an existing membership updates the role and warns.

### `user create`

| Flag              | Required | Notes                                          |
|-------------------|----------|------------------------------------------------|
| `--login`         | yes      | Also accepted as positional                    |
| `--email`         | yes      | RFC 5322                                       |
| `--name`          | no       | Defaults to login                              |
| `--password`      | yes*     | Masked prompt, confirmed. *Not required if `--provider external` |
| `--provider`      | no       | `local` (default), `oauth`, `external`         |
| `--groups`        | no       | Comma-separated                                |
| `--if-not-exists` | no       | Exit 0 on duplicate                            |

### `user delete`

| Flag                  | Required | Notes                                            |
|-----------------------|----------|--------------------------------------------------|
| `--login`             | yes      | Also accepted as positional                      |
| `--yes`               | no       | Skip confirmation                                |
| `--keep-memberships`  | no       | Leave project memberships untouched              |
| `--if-exists`         | no       | Exit 0 if user absent                            |

Confirmation shows user, memberships count, active sessions.

### `project create`

| Flag                 | Required | Notes                                    |
|----------------------|----------|------------------------------------------|
| `--name`             | yes      | Also accepted as positional              |
| `--label`            | no       | Defaults to name                         |
| `--description`      | no       |                                          |
| `--source-path`      | no       | Must exist and be readable               |
| `--allow-from-mask`  | no       | CIDR, default `0.0.0.0/0`                |
| `--creator`          | no       | Defaults to authenticated user           |
[...]

PR #2143, opened 2026-05-07
feat: CLI user creation and deletion

pirhoo:
As defined in #2122, Adds a `datashare user` subcommand group for creating and deleting Datashare users from the CLI, plus the service and repository plumbing they need. Both commands support flag-driven, interactive, and `--no-input` (scripted) usage, with `--if-not-exists` / `--if-exists` for idempotent runs and `--json` for machine-readable output.

## List of changes

- feat(repository): add `Repository.deleteUser` with transactional cascade
- perf(db): collapse `user_history_project` cleanup into a subquery `DELETE`
- feat(api): add `User.OAUTH` and `User.EXTERNAL` provider constants
- feat(user-admin): add `UserAdminService` with create/delete and idempotent variants
- feat(cli): add `Validators` and `Prompter` for input validation and prompts
- feat(cli): add `user create` and `user delete` picocli commands
- feat(cli-dispatch): wire commands into `CliApp` and bind `UserAdminService` in `CommonMode`
- test: cover cascade, service, validators, prompter and dispatch paths
- docs(db): document `deleteUser` cascade scope (intentionally untouched tables)
- refactor(cli): drop `char[]` wiping in `Prompter` and assert `MAX_RETRIES >= 1`
- refactor(cli): make `Validators.login/email/provider` return `void`
- refactor(user-admin): `UserAdminService` self-validates null/empty fields
- feat(cli): `datashare user` exits 2 (not 0) when no subcommand is given
- refactor(cli-dispatch): replace JSON blob with typed sibling-key Properties

## Preview

https://github.com/user-attachments/assets/bcef8332-71dc-4524-a841-772aae92ddde

## Implementation choices

**Three-layer architecture.** Each layer has a single responsibility:

1. Picocli commands (`UserCreateCommand`, `UserDeleteCommand`) own all IO: flag parsing, syntactic validation via `Validators`, interactive prompts via `Prompter`, and (for delete) the destructive-action confirmation. They marshal results into typed sibling-key Properties: `userCreate` carries the login, `userCreate.email` / `.password` / ... carry the rest. Booleans use `putIfTrue`; absent keys mean "not set".

2. `CliApp` dispatch is IO-free. It reads sibling-key Properties, constructs a `UserCreateRequest`, calls `UserAdminService`, and maps exceptions to exit codes (0/3/4/5). The password is removed from the in-memory `Properties` in a `finally` block. No JSON parsing, no prompting.

3. `UserAdminService` owns semantic validation (`ValidationException` for null/empty login, email, provider, and missing-password-when-local) and persistence. A programmatic caller that bypasses the CLI still gets the same validation contract.

Validation is duplicated by design. CLI-side `Validators` give immediate, field-by-field feedback during interactive prompts; service-side `validate()` is the source of truth for non-CLI callers. The `ValidationException` javadoc records why these can't be merged (the CLI module cannot depend on `datashare-app`).

Local-provider passwords are read via `System.console().readPassword()` and warned about when passed via `--password`. The CLI does not zero the `char[]` after reading: the password is hashed before `Repository.save`, but it transits as a `String` through the `Properties` map, so the `char[]` wipe would only mask a small fraction of the in-memory lifetime. The dispatch handler removes `userCreate.password` from the Properties as soon as the request is built (including on the error path).

## Files to review

- `datashare-cli/src/main/java/org/icij/datashare/cli/command/UserCreateCommand.java`:
    Picocli command, syntactic validation, interactive prompts, sibling-key marshalling.
- `datashare-cli/src/main/java/org/icij/datashare/cli/command/UserDeleteCommand.java`:
    Same shape as create, plus the destructive-action confirmation (now CLI-side, not in dispatch).
- `datashare-cli/src/main/java/org/icij/datashare/cli/Prompter.java`:
    Line / password / confirm prompts; the password path no longer wipes `char[]`.
- `datashare-cli/src/main/java/org/icij/datashare/cli/Validators.java`:
    Login / email / provider / groups syntactic checks.
- `datashare-cli/src/main/java/org/icij/datashare/cli/DatashareCliOptions.java`:
    Sibling-key constants and the header comment explaining the dispatch ABI.
- `datashare-app/src/main/java/org/icij/datashare/CliApp.java`:
    IO-free dispatch handlers reading sibling-key Properties; password is scrubbed in `finally`.
- `datashare-app/src/main/java/org/icij/datashare/user/admin/UserAdminServiceImpl.java`:
    Service-side validation as the source of truth, plus persistence.
- `datashare-db/src/main/java/org/icij/datashare/db/JooqRepository.java#deleteUser`: 
    Transactional cascade with a header comment listing tables intentionally not touched.

## AI disclosure

Mostly implemented with [superpowers skills](https://github.com/obra/superpowers) with many validation steps. The entire code was reviewed by human before opening the PR.

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/CliApp.java
+++ b/datashare-app/src/main/java/org/icij/datashare/CliApp.java
@@ -110,12 +110,12 @@ class CliApp {
             System.exit(0);
         }
 
-        if (properties.getProperty("userCreate") != null) {
+        if (properties.getProperty(USER_CREATE_OPT) != null) {
             UserAdminService userAdminService = mode.get(UserAdminService.class);
             System.exit(handleUserCreate(userAdminService, properties));
         }
 
-        if (properties.getProperty("userDelete") != null) {
+        if (properties.getProperty(USER_DELETE_OPT) != null) {
             UserAdminService userAdminService = mode.get(UserAdminService.class);
             System.exit(handleUserDelete(userAdminService, properties));
         }
@@ -164,20 +164,21 @@ class CliApp {
     private static final ObjectMapper MAPPER = new ObjectMapper();
 
     static int handleUserCreate(UserAdminService service, Properties properties) {
-        String payload = (String) properties.remove("userCreate"); // remove after read
+        // Consume the property so the worker pipeline below does not re-process it.
+        String payload = (String) properties.remove(USER_CREATE_OPT);
         try {
             @SuppressWarnings("unchecked")
-            Map<String, Object> p = MAPPER.readValue(payload, Map.class);
-            boolean ifNotExists = Boolean.TRUE.equals(p.get("ifNotExists"));
-            boolean json = Boolean.TRUE.equals(p.get("json"));
+            Map<String, Object> fields = MAPPER.readValue(payload, Map.class);
+            boolean ifNotExists = Boolean.TRUE.equals(fields.get("ifNotExists"));
+            boolean json = Boolean.TRUE.equals(fields.get("json"));
 
             UserCreateRequest req = new UserCreateRequest(
-                    (String) p.get("login"),
-                    (String) p.get("email"),
-                    (String) p.get("name"),
-                    (String) p.get("password"),
-                    (String) p.get("provider"),
-                    ((List<?>) p.getOrDefault("groups", List.of())).stream()
+                    (String) fields.get("login"),
+                    (String) fields.get("email"),
+                    (String) fields.get("name"),
+                    (String) fields.get("password"),
+                    (String) fields.get("provider"),
+                    ((List<?>) fields.getOrDefault("groups", List.of())).stream()
                             .map(Object::toString).collect(java.util.stream.Collectors.toList()));
 
             UserCreated created = ifNotExists ? service.createIfNotExists(req) : service.create(req);
@@ -208,14 +209,15 @@ class CliApp {
     }
 
     static int handleUserDelete(UserAdminService service, Properties properties) {
-        String payload = (String) properties.remove("userDelete");
+        // Consume the property so the worker pipeline below does not re-process it.
+        String payload = (String) properties.remove(USER_DELETE_OPT);
         try {
             @SuppressWarnings("unchecked")
-            Map<String, Object> p = MAPPER.readValue(payload, Map.class);
-            boolean ifExists = Boolean.TRUE.equals(p.get("ifExists"));
-            boolean json = Boolean.TRUE.equals(p.get("json"));
-            boolean yes = Boolean.TRUE.equals(p.get("yes"));
-            String login = (String) p.get("login");
+            Map<String, Object> fields = MAPPER.readValue(payload, Map.class);
+            boolean ifExists = Boolean.TRUE.equals(fields.get("ifExists"));
+            boolean json = Boolean.TRUE.equals(fields.get("json"));
+            boolean yes = Boolean.TRUE.equals(fields.get("yes"));
+            String login = (String) fields.get("login");
 
             if (!yes) {
                 // Confirmation prompt — only matters in interactive mode. The CLI
@@ -264,8 +266,8 @@ class CliApp {
         if (payload == null) return false;
         try {
             @SuppressWarnings("unchecked")
-            Map<String, Object> p = MAPPER.readValue(payload, Map.class);
-            return Boolean.TRUE.equals(p.get("json"));
+            Map<String, Object> fields = MAPPER.readValue(payload, Map.class);
+            return Boolean.TRUE.equals(fields.get("json"));
         } catch (Exception e) { return false; }
     }
 
--- a/datashare-app/src/main/java/org/icij/datashare/CliApp.java
+++ b/datashare-app/src/main/java/org/icij/datashare/CliApp.java
@@ -2,6 +2,7 @@ package org.icij.datashare;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.icij.datashare.cli.CliExtensionService;
+import org.icij.datashare.cli.Prompter;
 import org.icij.datashare.cli.spi.CliExtension;
 import org.icij.datashare.mode.CommonMode;
 import org.icij.datashare.tasks.ArtifactTask;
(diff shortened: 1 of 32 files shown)


datashare-app/src/main/java/org/icij/datashare/CliApp.java
```diff
@@ -142,4 +161,95 @@ private static void runTaskWorker(CommonMode mode, Properties properties) throws
         taskManager.awaitTermination(Integer.MAX_VALUE, SECONDS);
         taskManager.shutdown();
     }
+
+    // Single ObjectMapper shared across handlers for serializing JSON output.
+    // Input is read from typed sibling properties, not parsed JSON, so this
+    // mapper never sees untrusted strings.
+    private static final ObjectMapper MAPPER = new ObjectMapper();
+
+    static int handleUserCreate(UserAdminService service, Properties properties) {
```
caro3801, line 170: **CliApp.handleUserCreate / handleUserDelete — consider extracting a CliUserPresenter**                                                                                                                       
                                                                                                                                                                                                            
These two handlers currently mix dispatch (calling the service, catching exceptions, returning exit codes) with presentation (formatting JSON vs. plain-text output, writing to stdout/stderr). In Clean Architecture terms the presenter is a separate concern from the controller/dispatcher.                                                                                                                    
                                                                                                                                                                                                            
Suggestion: extract a small `CliUserPresenter` class that owns the `ObjectMapper`, the `System.out/System.err` streams, and the json flag, and exposes methods like `presentCreated(UserCreated)`,  `presentDeleted(String, boolean)`, and `presentError(String, String, int)`. `CliApp` then only coordinates the use-case call and exception mapping - no formatting logic.                                       
                                                                                                                                                                                                            
Side benefits:                                                                                                                                                                                            
  - `UserDeleteCommand` can drop its own `private static final ObjectMapper MAPPER`  and reuse the same presenter, eliminating the duplication.                                                                  
  - The presenter is easily testable by injecting a `ByteArrayOutputStream`-backed `PrintStream` without redirecting `System.out`.                                                                                
                                                                                                                                                                                                            
This is a non-blocking observation. Just worth considering if the output format is likely to evolve (e.g. adding a --quiet flag or a structured log format later).

datashare-app/src/main/java/org/icij/datashare/user/admin/UserExistsException.java
```diff
@@ -0,0 +1,7 @@
+package org.icij.datashare.user.admin;
+
+public class UserExistsException extends Exception {
```
caro3801, line 3: question/suggestion: if the exceptions (UserExists, UserNotFound, Validation) are Domain exception would it be interesting to moving them to datashare-api so they are available to future callers (if we want to use REST to create users for instance)
This could also prevent CliApp importing from  `org.icij.datashare.user.admin.*`
This is non blocking, but can be done in a follow up refactor if needed

caro3801 (approved):
The PR description was helpful to navigate the changes, thanks. This feature is a great addition and takes a lot of edge cases into account, that will be very helpful in the futures!
The suggestions are non blocking for merging. I used claude code to help formulate the comments.

datashare-app/src/main/java/org/icij/datashare/CliApp.java
```diff
@@ -142,4 +161,95 @@ private static void runTaskWorker(CommonMode mode, Properties properties) throws
...
+                System.out.println(MAPPER.writeValueAsString(Map.of(
+                        "created", !created.noop(),
+                        "noop", created.noop(),
+                        "login", created.login(),
+                        "email", created.email(),
+                        "name", created.name(),
+                        "provider", created.provider(),
+                        "groups", created.groups())));
+            } else if (created.noop()) {
+                System.out.println("user '" + created.login() + "' already exists (no-op)");
```
mvanzalu, line 195: question: I see that other options such as CRE_API_KEY_OPT or GET_API_KEY_OPT that displays information with `logger`, is it intended that the new handleUserCreate / handleUserDelete handlers write to System.out ?

pirhoo: Yes, this is intentional. The two are different output models:

Logger (CRE_API_KEY_OPT, GET_API_KEY_OPT): output is decorated by the logging framework (timestamps, level prefix, configurable sinks) which is fine for one-off operational messages.

System.out / System.err (user create / user delete): these are proper CLI subcommands with three features that don't mix with a logger:

1. `--json` machine-parseable output. A logger would prepend INFO 2026-05-13 ... [main] to each line and break datashare user create --json ... | jq. Stdout has to be the raw result.
2. stdout/stderr split. Success goes to stdout, errors to stderr (error(...) at line 242). Lets callers redirect the result while still seeing failures: datashare user create `--json ... > user.json`.
3. Exit codes. 0 / 3 / 4 / 5 / 1 are part of the command's contract. The handler owns both the exit code and the message that explains it.

So the existing API-key handlers are the legacy pattern (they were bolted onto the long-running datashare process and inherited its logger), and the new user subcommands are the deliberate CLI shape. Worth keeping them on System.out / System.err.

datashare-app/src/main/java/org/icij/datashare/CliApp.java
pirhoo: Good call, however I'd like to defer for now: the two outpus live in different modules (picocli `PrintWriter` in `UserDeleteCommand`, raw `System.out` in `CliApp` after dispatch), so a shared presenter would have to bridge both stream mechanics. 

The only real duplication today is a one-line `ObjectMapper` constant, and the output shapes differ per command (aborted payload vs created payload vs deleted payload), so the presenter API would either be command-specific (no real reuse) or a thin facade over `println(MAPPER.writeValueAsString(...))`.

Happy to revisit if we add `--quiet` (good idea!), another user subcommand, or structured logs: at that point the shape is forced and the extraction pays for itself.

datashare-app/src/main/java/org/icij/datashare/user/admin/UserExistsException.java
pirhoo: Agree the gap is real (the `user.admin.*` import from `CliApp` is a soft layer crossing). 

I'd like to defer to a follow-up refactor though: the cleanest move is exceptions plus `UserAdminService` (the interface) together so that a future REST resource depends only on datashare-api. 

Moving the exceptions alone leaves `CliApp` still importing `user.admin.UserAdminService`, so the coupling is only half-resolved. `ValidationException` also has a doc comment about its CLI twin that needs rewording on the way out.

datashare-app/src/main/java/org/icij/datashare/CliApp.java
mvanzalu: Makes sense ! Thanks !

mvanzalu (approved):

PR #2147, opened 2026-05-15
feat: CLI project creation and deletion

pirhoo:
As defined in #2122, adds a `datashare project` subcommand group for creating and deleting projects from the CLI. Both commands support flag-driven, interactive, and `--no-input` usage, with `--if-not-exists` / `--if-exists` for idempotent runs and `--json` for machine-readable output. `delete` runs a typed-name confirmation before cascading through DB, ES index, queues, report map, and artifact dir.

## List of changes

- feat(indexer): add `Indexer.count(indexName)` using the ES `_count` API
- feat(api): hoist precompiled `Project.NAME_PATTERN` / `ALLOW_FROM_MASK_PATTERN`
- feat(project-admin): add `ProjectAdminService` with create / delete (and idempotent variants), stats, and admin-grant
- feat(project-admin): delete cascade across DB, ES, queues, report map, artifacts
- feat(cli): add `Validators` for `projectName`, `allowFromMask`, `uri`, `iso8601`
- feat(cli): add `project create` and `project delete` picocli commands
- feat(cli-dispatch): wire commands into `CliApp` via typed sibling-key Properties
- feat(cli-dispatch): auto-grant `PROJECT_ADMIN` to `--creator` (defaults to `defaultUserName`)
- fix(dist): route the `project` subcommand through the new picocli path
- test: cover service, validators, dispatch, and end-to-end picocli graph

## Implementation choices

**Three-layer architecture.** Each layer has a single responsibility:

1. Picocli commands own all IO: flag parsing, syntactic validation, prompts, and the typed-name confirmation. They marshal results into typed sibling-key Properties: `projectCreate` carries the name, `projectCreate.label` / `.sourcePath` / ... carry the rest. Absent keys mean "not set".

2. `CliApp` dispatch is IO-free. It reads sibling-key Properties, builds a `ProjectCreateRequest` (or `ProjectDeleteOptions`), calls the service, and maps exceptions to exit codes (0 / 1 / 3 / 5).

3. `ProjectAdminService` owns semantic validation (`ValidationException`) and persistence. Programmatic callers get the same validation contract.

Validation is duplicated by design: CLI-side validators give immediate field-by-field feedback; service-side `validate()` is the source of truth. Both share the precompiled `Pattern` constants on `Project`.

**Delete cascade containment.** Each cleanup step is wrapped so a failure logs and sets its boolean to `false` instead of aborting the cascade. Operators can re-run with `--if-exists` after fixing the failing dependency.

**Auto-grant of `PROJECT_ADMIN`.** On successful create, the dispatcher resolves the grantee from `--creator` then `defaultUserName`. The service updates `user_inventory.details` then adds the casbin policy; if casbin fails, the inventory write rolls back. `USER_NOT_FOUND` is a soft failure: warn and exit 0 (suppressed on the `defaultUserName` fallback).

**Launcher wiring.** The launcher now injects `--defaultUserName` and the shared paths (`--dataSourceUrl`, `--settings`, `--dataDir`, ...) so the CLI talks to the same DB and config as the web server.

## Files to review

- `datashare-cli/.../command/ProjectCreateCommand.java`: flag-vs-prompt resolution, sibling-key marshalling.
- `datashare-cli/.../command/ProjectDeleteCommand.java`: same shape as create, plus missing-name error paths.
- `datashare-cli/.../Validators.java`: `projectName`, `allowFromMask`, `uri`, `iso8601`.
- `datashare-cli/.../DatashareCliOptions.java`: sibling-key constants for the dispatch ABI.
- `datashare-app/.../CliApp.java`: dispatch handlers, auto-grant, typed-name confirmation, exit codes.
- `datashare-app/.../project/admin/ProjectAdminServiceImpl.java`: validation, persistence, cascade, auto-grant rollback.
- `datashare-api/.../indexing/Indexer.java` and `ElasticsearchIndexer.java`: `count(indexName)`.
- `datashare-dist/src/main/deb/bin/datashare`: `project` routing and shared-path injection.

## AI disclosure

Mostly implemented with superpowers skills with many validation steps. The entire code was reviewed by human before opening the PR.


datashare-cli/src/test/java/org/icij/datashare/cli/command/DatashareCommandTest.java
```diff
@@ -1213,4 +1213,201 @@ public void test_user_delete_confirmed_at_confirm_dispatches() {
         Properties props = cmd.getSubcommandProperties();
         assertThat(props.getProperty("userDelete")).isEqualTo("alice");
     }
+
+    @Test
```
caro3801: this file starts to be very long such as DatashareCommand. 
suggestion: refactor to split by CLI section

datashare-app/src/main/java/org/icij/datashare/project/admin/ProjectAdminServiceImpl.java
```diff
@@ -0,0 +1,377 @@
...
+        if (!currentProjects.contains(projectName)) {
+            currentProjects.add(projectName);
+        }
+        apps.put("datashare", currentProjects);
+        newDetails.put("groups_by_applications", apps);
+
+        User updated = new User(user.id, user.name, user.email, user.provider, newDetails);
+        repository.save(updated);
+
+        // 2. Casbin grouping policy: g <user.id> PROJECT_ADMIN datashare::<projectName>.
```
caro3801: Currently the the default domain is "default", changing this to "datashare" might create issues in the current instances running.

caro3801: Check the original task to see if it has identical behavior: https://github.com/ICIJ/datashare/blob/a95bab4268eb4043d1955790c7202c3552a50dba/datashare-app/src/main/java/org/icij/datashare/tasks/GrantAdminPolicyTask.java

datashare-app/src/main/java/org/icij/datashare/project/admin/ProjectAdminServiceImpl.java
```diff
@@ -0,0 +1,377 @@
...
+    }
+
+    private ProjectDeleted cascade(Project project, ProjectDeleteOptions options) {
+        // Per-step containment: each cleanup step swallows its own failures and
+        // sets the corresponding bool to false. A partial-failure cascade is
+        // worse than a continuing one -- operators can rerun with --if-exists
+        // to converge once they have fixed the failing dependency, but a
+        // half-deleted project where (say) the index is gone but queues are
+        // intact is a sticky state if the cascade aborts mid-stream.
+        String name = project.getName();
```
caro3801, line 206: suggestion: I'm wondering if this should be handle by a specific repository tlike ProjectDeleteRepository to prevent duplication with the web endpoint (ProjectResource.projectDelete()): same steps, same order, same helpers for queues, report map, and  artifacts. 
Seeing that they've already diverged, the web endpoint uses `allMatch` for queues (short-circuits on failure) while the CLI uses `reduce` (continues through failures), the two paths behave differently on partial queue failures.
Worth extracting a ProjectDeleteRepository that both the web endpoint and the CLI service delegate to.

datashare-app/src/main/java/org/icij/datashare/CliApp.java
```diff
@@ -239,6 +263,394 @@ static int handleUserDelete(UserAdminService service, Properties properties) {
...
+        boolean ifNotExists = Boolean.parseBoolean(properties.getProperty(PROJECT_CREATE_IF_NOT_EXISTS_OPT));
+        try {
+            ProjectCreateRequest request = buildCreateRequest(properties);
+            ProjectCreated created;
+            if (ifNotExists) {
+                created = service.createIfNotExists(request);
+            } else {
+                created = service.create(request);
+            }
+            GrantOutcome grant = attemptAutoGrant(service, created, properties);
```
caro3801, line 318: suggestion: `attemptAutoGrant` in `CliApp` is a business rule : the creator of a project should get `PROJECT_ADMIN`, but it lives in the CLI dispatcher. Any other entry point calling `service.create()` (a future REST endpoint, a migration script) would silently skip the grant.

The creator field is already threaded through `ProjectCreateRequest` via `PROJECT_CREATE_CREATOR_OPT`; the service just never reads it. Moving the grant logic into `ProjectAdminServiceImpl.persist()` would make it apply regardless of caller, and let `CliApp.handleProjectCreate` drop `attemptAutoGrant` entirely.

datashare-app/src/main/java/org/icij/datashare/CliApp.java
```diff
@@ -239,6 +263,394 @@ static int handleUserDelete(UserAdminService service, Properties properties) {
...
+            if (json) {
+                System.out.println(MAPPER.writeValueAsString(createResultMap(created, grant)));
+            } else {
+                emitCreateText(created, grant);
+            }
+            return 0;
+        } catch (ProjectExistsException e) {
+            return error(e.getMessage(), "conflict", 4, json);
```
caro3801: hint: this methods returns `0, 1, 4, 5` but they aren't declared in `ProjectAdminService` or anywhere operators can find them without reading source. A short constant block or enum (mirroring what `handleUserCreate` likely has) would help.

caro3801 (approved):
This PR is great, complete and a great addition for operating projects directly from the CLI. I noticed some code duplication and legacy tasks where operations in CLI and REST diverged it would be great if it stays uniform either by changing it now and in a follow up refactor.

datashare-cli/src/test/java/org/icij/datashare/cli/command/DatashareCommandTest.java
pirhoo: Fair point, I'm on it!

pirhoo: https://github.com/ICIJ/datashare/commit/118fb038be0333a1482cb00778d8677f90f9dfb3

datashare-app/src/main/java/org/icij/datashare/project/admin/ProjectAdminServiceImpl.java
pirhoo: Fixed! https://github.com/ICIJ/datashare/commit/aaaae12ebb2a200c4b5a45f9a2e3fa4535493dad

datashare-app/src/main/java/org/icij/datashare/project/admin/ProjectAdminServiceImpl.java
pirhoo: Good catch, the queue handling difference (allMatch short-circuit vs reduce continue-through) is real, and the web path also lacks the per-step exception containment that the CLI uses, so a failure at one step skips the rest of the cascade.

I'd like to defer the extraction to a follow-up work rather than do it here:

1. It's a semantic decision, not just deduplication. Picking the unified behavior means deciding whether the web endpoint should adopt the CLI's per-step containment (and start surviving partial failures it currently aborts on), and whether the cascade should return a structured per-step result instead of just logging. The CLI also has `--keep-index` which has no web equivalent. Those are choices worth their own review, not a side-effect of this PR.
2. Scope. This PR is the CLI project CRUD feature; refactoring `ProjectResource.projectDelete()` is a separate concern with its own test surface.
3. Repository is already taken by org.icij.datashare.Repository (the DB layer). A shared component for this would want a different name. Something like `ProjectCascadeDeleter` as a domain service, to avoid confusion.

datashare-app/src/main/java/org/icij/datashare/CliApp.java
pirhoo: Two notes:

- creator isn't actually in `ProjectCreateRequest` today (12 fields, none named creator); the dispatcher reads `PROJECT_CREATE_CREATOR_OPT` from properties directly.
- The dispatcher does more than forward a creator string: it falls back to defaultUserName (a CLI launcher concept, meaningless for REST), applies a CLI-specific warning policy (warn on miss only when `--creator` was explicit), and skips the grant on noop. Pushing those into the service bakes CLI semantics into a service that shouldn't know them.

The service exposes create() and addAdminToProject(), the CLI composes them. A future caller would compose them the same way with its own resolution and UX. Happy to handle it later when a second caller actually shows what the caller-agnostic policy should be.

datashare-app/src/main/java/org/icij/datashare/CliApp.java
pirhoo: I was thinking about it too, thanks for the hint! It's done here https://github.com/ICIJ/datashare/commit/d3b3f3dc940b5fb9870500b347a7b475110de53e

pirhoo:
Thanks for the review!

Issue #2153, opened 2026-05-26
feat: use async search to perform heavy searches


pirhoo:
Searches that fan out across many project indices can take long enough that the proxy returns `upstream request timeout` before Elasticsearch responds. We want the frontend to stop waiting on a single blocking request and instead use Elasticsearch's [async search](https://www.elastic.co/guide/en/elasticsearch/reference/current/async-search.html).

## What we want to do

Implement async search for the search on the front:

- Submit the search to `_async_search` with a short `wait_for_completion_timeout` (e.g. `1s`) and a `keep_alive` (e.g. `5m`).
- If ES doesn't finish in time, get back an `id` + partial results and poll  `GET /_async_search/<id>` until `is_running: false`.
- Show progress while polling (e.g. "This might take more time than expected").
- Cancel in-flight work with `DELETE /_async_search/<id>` on dismount / new search.

**Note**: This hides the latency, it doesn't reduce it. Query-cost reduction is tracked separately and should land first/alongside.

PR #2158, opened 2026-05-27
feat: async search proxy

pirhoo:
Searches that fan out across many project indices can take long enough that the gateway in front of Datashare times out before Elasticsearch answers. This adds backend support for Elasticsearch async search so the frontend can stop waiting on a single blocking request: it can submit a search, poll it while it runs, and cancel it when the user moves on. This hides the latency, it does not reduce it. This is part of icij/datashare#2153.

Async searches go through the existing search proxy. Each one is tied to the user who submitted it, so only that user can poll or cancel it, and the result's lifetime in Elasticsearch is kept in step with the backend so it does not linger after the search is finished.

## Changes

- feat: support Elasticsearch async search (submit, poll, cancel) through the index proxy
- feat: bind each async search to the user who submitted it, and authorize poll and cancel per user
- feat: keep async ownership in memory for local mode and in Redis for server modes
- feat: keep the Elasticsearch result lifetime aligned with the backend on submit
- refactor: tidy the index resource
- test: cover the async search flows and the ownership store

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/web/IndexResource.java
+++ b/datashare-app/src/main/java/org/icij/datashare/web/IndexResource.java
@@ -201,7 +201,7 @@ public class IndexResource {
             return PayloadFormatter.error("async search not found", HttpStatus.NOT_FOUND);
         }
         try {
-            String response = indexer.executeRaw(method, withQueryParams(path, context), null);
+            String response = indexer.executeRaw(method, IndexAccessVerifier.getUrlString(context, path), null);
             if ("DELETE".equalsIgnoreCase(method)) {
                 asyncSearchStore.remove(id);
             } else {
@@ -275,7 +275,7 @@ public class IndexResource {
         modeVerifier.checkAllowedMode(Mode.LOCAL, Mode.EMBEDDED);
         try {
             String path = IndexAccessVerifier.checkIndices(index) + "/_close";
-            return PayloadFormatter.json(indexer.executeRaw("POST", withQueryParams(path, context), null));
+            return PayloadFormatter.json(indexer.executeRaw("POST", IndexAccessVerifier.getUrlString(context, path), null));
         } catch (IllegalArgumentException e) {
             return PayloadFormatter.error(e, HttpStatus.BAD_REQUEST);
         }
@@ -293,7 +293,7 @@ public class IndexResource {
         modeVerifier.checkAllowedMode(Mode.LOCAL, Mode.EMBEDDED);
         try {
             String path = IndexAccessVerifier.checkIndices(index) + "/_open";
-            return PayloadFormatter.json(indexer.executeRaw("POST", withQueryParams(path, context), null));
+            return PayloadFormatter.json(indexer.executeRaw("POST", IndexAccessVerifier.getUrlString(context, path), null));
         } catch (IllegalArgumentException e) {
             return PayloadFormatter.error(e, HttpStatus.BAD_REQUEST);
         }
@@ -380,7 +380,7 @@ public class IndexResource {
         modeVerifier.checkAllowedMode(Mode.LOCAL, Mode.EMBEDDED);
         String path = "_snapshot/" + repository + "/" + snapshot;
         String body = request.contentAsBytes().length > 0 ? new String(request.contentAsBytes()) : null;
-        return PayloadFormatter.json(indexer.executeRaw("PUT", withQueryParams(path, context), body));
+        return PayloadFormatter.json(indexer.executeRaw("PUT", IndexAccessVerifier.getUrlString(context, path), body));
     }
 
     @Operation(description = "Restore a snapshot. Only available in LOCAL and EMBEDDED modes.")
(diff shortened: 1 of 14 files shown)


datashare-app/src/main/java/org/icij/datashare/web/IndexResource.java
```diff
@@ -93,18 +119,57 @@ public Payload esHead(final String path) throws IOException {
...
+                && isBlank(context.get("keep_alive"));
+        if (!isSubmitWithoutKeepAlive) {
+            return esUrl;
+        }
+        String parameterSeparator = esUrl.contains("?") ? "&" : "?";
+        return esUrl + parameterSeparator + "keep_alive=" + defaultKeepAliveParam;
+    }
+
+    private static boolean isBlank(String value) {
+        return value == null || value.isEmpty();
```
caro3801: Prevent injection of spaces
```suggestion
        return value == null || value.trim().isEmpty();
```

datashare-app/src/main/java/org/icij/datashare/web/IndexResource.java
```diff
@@ -93,18 +119,57 @@ public Payload esHead(final String path) throws IOException {
...
     @ApiResponse(responseCode = "200", description = "returns 200")
     @ApiResponse(responseCode = "400", description = "returns 400 if there is an error from ElasticSearch")
     @Post("/search/:path:")
     public Payload esPost(@Parameter(name = "index", description = "elasticsearch path", in = ParameterIn.PATH) final String path, Context context, final net.codestory.http.Request request) throws IOException {
         try {
-            return PayloadFormatter.json(indexer.executeRaw("POST", IndexAccessVerifier.checkPath(path, context), new String(request.contentAsBytes())));
+            String esUrl = withInjectedKeepAlive(IndexAccessVerifier.checkPath(path, context), path, context);
+            String response = indexer.executeRaw("POST", esUrl, new String(request.contentAsBytes()));
+            if (IndexAccessVerifier.isAsyncSearchSubmit(path)) {
+                recordAsyncSearchOwnership(path, context, response);
```
caro3801: IOException from `recordAsyncSearchOwnership` escapes the catch block 

suggestion: adding a log could be of interest here to track is there are parsing issues. The try/catch only caught IllegalArgumentException. If ES returned a non-JSON body, readTree() threw IOException which escaped the catch block
```suggestion                                                                              
     try {                                                                                                                                           
         recordAsyncSearchOwnership(path, context, response);                                                                                        
     } catch (IOException e) {                                                                                                                       
         LOGGER.warn("async search submitted but ownership could not be recorded: {}", e.getMessage());                                              
         return PayloadFormatter.error("async search submitted but response could not be parsed", HttpStatus.BAD_GATEWAY);                           
     }  
```

datashare-app/src/main/java/org/icij/datashare/web/IndexResource.java
```diff
@@ -115,17 +180,81 @@ public Payload esPost(@Parameter(name = "index", description = "elasticsearch pa
...
+            return PayloadFormatter.error("async search not found", HttpStatus.NOT_FOUND);
+        }
+        AsyncSearchOwner ownerRecord = owner.get();
+        if (!isOwnedByCurrentUser(ownerRecord, user)) {
+            return PayloadFormatter.error("async search not found", HttpStatus.NOT_FOUND);
+        }
+        try {
+            String response = indexer.executeRaw(method, withQueryParams(path, context), null);
+            if ("DELETE".equalsIgnoreCase(method)) {
+                asyncSearchStore.remove(id);
```
caro3801: if ES DELETE succeeded and Redis is down for example, it would raise an exception and the user gets 500 despite cancel having worked. Fix: ES delete succeded, we want a 200  => Add a log and continue, the TLL will expire on its own.
```suggestion
try {
  asyncSearchStore.remove(id);
} catch (Exception e) {
  LOGGER.warn("async search cancelled on ES but ownership record could not be removed: {}", e.getMessage());
}
```

datashare-app/src/main/java/org/icij/datashare/web/IndexResource.java
```diff
@@ -115,17 +180,81 @@ public Payload esPost(@Parameter(name = "index", description = "elasticsearch pa
...
+        Optional<AsyncSearchOwner> owner = asyncSearchStore.get(id);
+        if (owner.isEmpty()) {
+            return PayloadFormatter.error("async search not found", HttpStatus.NOT_FOUND);
+        }
+        AsyncSearchOwner ownerRecord = owner.get();
+        if (!isOwnedByCurrentUser(ownerRecord, user)) {
+            return PayloadFormatter.error("async search not found", HttpStatus.NOT_FOUND);
+        }
+        try {
+            String response = indexer.executeRaw(method, withQueryParams(path, context), null);
```
caro3801: minor suggestion (found by Claude code): `withQueryParams()` duplicates `IndexAccessVerifier.getUrlString()`   
- File: IndexResource.java ~line 443 - IndexAccessVerifier L78
- Two diverging implementations of query-string serialisation. Neither percent-encodes values. If one is ever fixed, the other will silently lag. 
- Suggested fix: make `IndexAccessVerifier.getUrlString` package-accessible and call it from `withQueryParams`.

caro3801 (commented):
LGTM! Made few suggestions to add logs and prevent unexpected exception throwing. Nothing preventing from merging

caro3801 (approved):

pirhoo:
Thanks! I applied all the suggestions :)

Issue #2157, opened 2026-05-27
bug: i can't view "shared/public" batchsearches that are not my own


caro3801:
When viewing a shared batch search that is not my own, I see:  infinite loading of batch search name and no batch search card on the right + a 403 error in the console.
<img width="1104" height="424" alt="Image" src="https://github.com/user-attachments/assets/9a3fd2ae-ae28-496b-b1a1-3194ceb0084c" />

PR #2160, opened 2026-05-27
fix: resolve 403 for published batch searches viewed by non-owners

caro3801:
Related to https://github.com/ICIJ/datashare/issues/2157

## Summary

- Fix 403 error when a project member views a published batch search they did not create
- The ownership check in `findVisibleTaskFor` was short-circuiting before the batch search repository fallback, which is the only place published batch searches are resolved
- Refactor `findVisibleTaskFor` to use an `Optional`-based pipeline, eliminating a boolean flag

## Root cause

`findVisibleTaskFor` first looks up the task in the task manager, then falls back to `batchSearchRepository.getRecords` for batch searches. The previous fix added an ownership check after the task manager lookup but threw `ForbiddenException` immediately when the user was not the owner — before the repository fallback could check whether the batch search was published and accessible to the user.

## Changes

- **`TaskFinder.findVisibleTaskFor`**: when the task is found in the manager but the user is not the owner, fall through to the repository check instead of throwing. Throw 403 only if the repo check also fails; throw 404 if the task was never in the manager.
- **`TaskFinder.getTaskFromManager`**: new private helper wrapping the `try/catch` as `Optional<Task<?>>`, making the main method read as a clean pipeline.
- **`TaskFinderTest`**: 4 new cases — owner access, non-owner without a batch search record (403), non-owner with a live published batch search in the task manager (the bug scenario), unknown id (404).

Result for a batch search owned by "bar" (not current user): 
- batch search name is loaded, 
- the batch search card is loaded, 
- no 403 error:

image

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/tasks/TaskFinder.java
+++ b/datashare-app/src/main/java/org/icij/datashare/tasks/TaskFinder.java
@@ -2,21 +2,15 @@ package org.icij.datashare.tasks;
 
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
+import net.codestory.http.errors.ForbiddenException;
 import org.icij.datashare.Entity;
-import org.icij.datashare.asynctasks.Task;
-import org.icij.datashare.asynctasks.TaskFilters;
-import org.icij.datashare.asynctasks.TaskManager;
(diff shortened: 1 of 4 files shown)


pirhoo (approved):
LGTM thanks!

mvanzalu (approved):

PR #2170, opened 2026-06-02
feat: sync policies with redisson

caro3801:
# :warning:  This description is stale it was made for choosing between plugin and inhouse solution. Since inhouse dev has be chosen, see new [description](https://github.com/ICIJ/datashare/pull/2170#issuecomment-4680683073) and [test plan](https://github.com/ICIJ/datashare/pull/2170#issuecomment-4680694897) in comments below

Adds a Redis pub/sub watcher (`PolicyWatcher`) that broadcasts Casbin policy changes across server instances in real time, replacing the previous polling-only approach. When a project grant or revoke mutates the policy, the watcher publishes a reload signal on a Redis topic so all instances apply the change immediately rather than waiting for the next poll cycle.

Also fixes several correctness issues in the session and project-admin layers surfaced during review.

**What changed**

- **`PolicyWatcher`** - new `RTopic`-backed `Watcher` that publishes `instanceId:reload` on mutations and subscribes to reload events; suppresses self-triggered reloads; logs WARN when publish fails so the operator knows other instances won't see the change until the next poll cycle
- **`Authorizer`** - new `Watcher` constructor; `startAutoLoadPolicy(intervalMs)` backstop for partition recovery; implements `Closeable`
- **`CommonMode`** - wires `PolicyWatcher` when `busType=REDIS`; registers both watcher and authorizer with `addCloseable`; calls `startAutoLoadPolicy` for failsafe reload
- **`ProjectAdminServiceImpl`** - `findUser` helper checks SQL then Redis (`usersWritable.find`) so OAuth2-only users (never in SQL) are found during grant/revoke; `casbinWithRollback` now rolls back both SQL and Redis on Casbin failure; actionable hint in `UserNotFoundException` when `--authUsersProvider` is not `UsersInRedis`
- **`UsersInRedis.saveOrUpdate`** - WATCH/MULTI/EXEC retry loop eliminates the TOCTOU race between `ttl()` and `EXEC`; extend-only TTL rule (never shorten an active session); bounded to 10 attempts
- **`DatashareCliOptions` / `GlobalOptions` / `StageRunCommand`** - `--policyReloadInterval` option (default 30 s fallback poll)

**Tests added**

- `PolicyWatcherTest` - unit tests for publish, self-suppress, listener registration
- `AuthorizerRedisSyncIntTest` - integration test: two `Authorizer` instances share a real Redis; a grant on instance A is visible on instance B within 500 ms
- `UsersInRedisTest` - WATCH retry loop, extend-only TTL, persistent key handling
- `ProjectAdminServiceImplTest` - CLI option parsing, OAuth2 user fallback, hint message
- `AuthorizerTest` - watcher notification, polling lifecycle (`close()` stops timer)

**Important**
When `busType` is not `REDIS`, it falls into the `else` path on line 320 of `CommonMode.java`:

- **No `PolicyWatcher`** — no Redis pub/sub, no cross-instance notifications
- **`Authorizer(adapter, interval)`** — creates an authorizer with only periodic polling via `startAutoLoadPolicy(interval)`
- If `--policyReloadInterval` is also unset (defaults to `"0"`), then `interval=0` → `startAutoLoadPolicy` is a no-op → **the authorizer is completely static**: policy is loaded once at startup and never refreshed

So the behavior matrix:

| `busType`     | `--policyReloadInterval` | Result                                     |
| ------------- | ------------------------ | ------------------------------------------ |
| `REDIS`       | unset (0)                | Event-driven only; no polling backstop     |
| `REDIS`       | e.g. `30000`             | Event-driven + polling backstop every 30 s |
| anything else | unset (0)                | Static - loaded once, never refreshed      |
| anything else | e.g. `30000`             | Polling-only every 30 s                    |

In a single-server deploy without Redis (`busType=MEMORY`), the static case is fine - there's only one instance so there's nothing to sync. But if someone runs multiple servers without Redis, grants won't propagate to other instances unless `--policyReloadInterval` is set.

--- 
Differences between branches `feat/casbin-redis-watcher` and `feat/casbin-redis-watcher-ex`

|                            | `feat/casbin-redis-watcher`                                                              | `feat/casbin-redis-watcher-ex`                                                                                     |
| -------------------------- | ---------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------ |
| **Watcher implementation** | Hand-rolled `PolicyWatcher` using Redisson `RTopic`                                      | `SafeWatcherEx` wrapping `jcasbin-redis-watcher-ex` library's `RedisWatcherEx`                                     |
| **Watcher interface**      | `Watcher` (basic — one `update()` method)                                                | `WatcherEx` (fine-grained — separate callbacks per operation: `updateForAddPolicy`, `updateForRemovePolicy`, etc.) |
| **Self-suppress**          | Yes — each message is prefixed with `instanceId`; the publisher ignores its own messages | Handled internally by the library                                                                                  |
| **New dependency**         | None — reuses Redisson already in the project                                            | Adds `jcasbin-redis-watcher-ex 1.1.0` + 3 Netty exclusions to avoid classpath conflicts                            |
| **Redis client**           | Redisson (already a project dependency)                                                  | Lettuce (bundled inside `jcasbin-redis-watcher-ex`)                                                                |
| **Failure visibility**     | `PolicyWatcher.update()` catches and logs WARN                                           | `SafeWatcherEx` catches all `updateFor*()` and logs WARN (necessary because jcasbin swallows them silently)        |
| **Close/cleanup**          | `PolicyWatcher implements Closeable` — removes the RTopic listener                       | Library handles cleanup internally; no `Closeable` needed on the watcher                                           |
| **Polling backstop**       | `startAutoLoadPolicy(interval)`                                                          | Same                                                                                                               |
| **Code owned**             | ~60 lines, fully in-repo                                                                 | ~70 lines of shim wrapping a third-party class                                                                     |

**Key trade-off:** the `-watcher` branch adds zero dependencies and keeps full control; the `-watcher-ex` branch delegates to the library but requires managing Netty exclusions and adds a shim layer purely to recover the error visibility that jcasbin strips away. The `WatcherEx` granularity (per-operation callbacks) is the only functional advantage of the library, and Datashare doesn't currently use those callbacks for anything beyond triggering a reload.


caro3801:
## feat(casbin): Redis RTopic watcher for event-driven policy sync

### Summary

Replaces polling-only Casbin policy reload with a Redis pub/sub watcher, so all Datashare nodes pick up permission changes in near-real time when running in Redis mode. Includes a refactor of the user session layer to cleanly separate read-only user lookup from mutable session management.

### Changes

#### Core feature: Redis policy watcher
- Add `PolicyWatcher` that subscribes to a Redis RTopic and triggers Casbin policy reloads on publication; falls back to 30s polling when `busType` is not `REDIS`
- Suppress self-triggered reloads (the node that wrote the policy change skips its own notification)
- Expose `--policyReloadInterval` on the CLI; fall back to `DEFAULT_POLICY_RELOAD_INTERVAL` when unset
- Disable Casbin's verbose built-in logging by default (`SyncedEnforcer` enables it implicitly)
- Roll back Redis state on Casbin failure during project grant/revoke

#### Session layer refactor
- Extract `UsersIdProviderCache` interface + `UsersIdProviderRedisCache` implementation — session-based auth filters now use this focused interface instead of the broader `UsersWritable`
- Make `UsersInRedis` and `UsersInDb` read-only; mutable session writes go through `UsersIdProviderCache` only
- `UserStore` replaces `Repository` as the user management abstraction in session and admin code
- Fix `saveOrUpdate` TTL handling: preserve persistent/long-lived entries, drop the WATCH/retry complexity, always reset TTL on update
- Scope the Redis session cache to OAuth2 mode only

#### Mode wiring
- `CommonMode` now provides `Users` and `UsersIdProviderCache` as separate bindings (activated when `busType=REDIS`)
- Remove `UsersWritable` binding from `CommonMode`; `ProjectAdminServiceImpl` and `UserAdminServiceImpl` updated accordingly
- Close resources in LIFO order

#### CLI
- Remove duplicate options from `ServerOptions` that were already declared in `GlobalOptions`
- Surface an actionable hint when `UserNotFoundException` is thrown during first OAuth2 grant

### Tests
- `PolicyWatcherTest` — unit tests for watcher lifecycle and self-reload suppression
- `AuthorizerRedisSyncIntTest` — integration test verifying cross-node policy sync via Redis
- `UsersIdProviderRedisCacheTest` — covers session cache read/write/TTL behaviour (replaces `UsersInRedisTest`)
- Extended `ProjectCommandTest` and `DatashareCommandTest` for CLI option handling

caro3801:
# Casbin sync manual test scenarios

## Context

The `feat/casbin-redis-watcher` branch adds cross-instance Casbin policy sync via a Redis RTopic.
When one instance grants or revokes a project, all other instances receive a pub/sub notification
and reload their policy from the database. The writing instance suppresses its own notification.

This file is a manual test guide covering three auth/user-provider combinations to confirm the
sync works end-to-end in each configuration.

---

## Common setup

Prerequisites: Redis, PostgreSQL, and (for scenario 1) the <OAuth> OAuth2 provider are running.

Both instances share the same Redis and PostgreSQL. Instance 1 runs on port 8080, instance 2 on
port 8081. Both must have `--busType REDIS` so the RTopic watcher is active.

Common flags used in all scenarios:

```
-d ~/Datashare/Data/
--redisAddress redis://redis:6379
--dataSourceUrl "jdbc:postgresql://postgres/datashare?user=admin&password=admin"
--messageBusAddress redis://redis:6379
--busType REDIS
--mode SERVER
```

To confirm sync happened on instance 2, watch its logs for a line like:
`INFO  Authorizer - policy reloaded`

Or make an API call to instance 2 after the grant and verify the project appears in the user's
project list.

---

## Scenario 1 — OAuth2 auth

Users authenticate via the <OAuth> IdP. Their session is cached in Redis by the OAuth2 filter.
In OAuth2 mode `--authUsersProvider` is not needed: grant/revoke automatically looks up the user
from the session cache. The user must have logged in at least once before being granted.

### Start instance 1 (port 8080)

```bash
./run.sh app start \
  --auth oauth \
  -d ~/Datashare/Data/ \
  --redisAddress redis://redis:6379 \
  --messageBusAddress redis://redis:6379 \
  --dataSourceUrl "jdbc:postgresql://postgres/datashare?user=admin&password=admin" \
  --busType REDIS \
  --mode SERVER \
  --oauthClientId datashareuidforseed \
  --oauthClientSecret datasharesecretforseed \
  --oauthAuthorizeUrl http://<OAuth>:3001/oauth/authorize \
  --oauthTokenUrl http://<OAuth>:3001/oauth/token \
  --oauthApiUrl http://<OAuth>:3001/api/v1/me.json \
  --oauthCallbackPath /auth/callback
```

### Start instance 2 (port 8081)

```bash
./run.sh app start \
--auth oauth \
-d ~/Datashare/Data/ \
--redisAddress redis://redis:6379 \
--messageBusAddress redis://redis:6379 \
--dataSourceUrl "jdbc:postgresql://postgres/datashare?user=admin&password=admin" \
--busType REDIS \
--mode SERVER \
--oauthClientId datashareuidforseed \
--oauthClientSecret datasharesecretforseed \
--oauthAuthorizeUrl http://<OAuth>:3001/oauth/authorize \
--oauthTokenUrl http://<OAuth>:3001/oauth/token \
--oauthApiUrl http://<OAuth>:3001/api/v1/me.json \
--oauthCallbackPath /auth/callback \
--port 8081
```

Same flags, add `--port 8081`.

### Test steps

1. Log in to instance 1 at `http://localhost:8080` via OAuth2. This creates the Redis session key.
2. Note the login (e.g. `jdoe`) from the <OAuth> user list.
3. Grant the project from CLI:
   ```bash
   ./run.sh project grant --project local-datashare --user jdoe \
     --redisAddress redis://redis:6379 \
     --dataSourceUrl "jdbc:postgresql://postgres/datashare?user=admin&password=admin" \
     --busType REDIS
   ```
4. Check instance 2 logs for `policy reloaded`.
5. Check instance 1 logs — it should NOT show `policy reloaded` (self-suppression).
6. Call `GET http://localhost:8081/api/project` with jdoe's session cookie and confirm
   `local-datashare` appears.
7. Revoke and repeat verification.

---

## Scenario 2 — Form auth, database user provider

Users are stored in PostgreSQL. Authentication uses the form login filter.

### Create the test user

```bash
./run.sh user create \
  --login jdoe \
  --email jdoe@example.org \
  --name "Jane Doe" \
  --password secret \
  --provider local \
  --authUsersProvider database \
  --dataSourceUrl "jdbc:postgresql://postgres/datashare?user=admin&password=admin"
```

### Start instance 1 (port 8080)

```bash
./run.sh app start \
  --auth form \
  --authUsersProvider database \
  -d ~/Datashare/Data/ \
  --redisAddress redis://redis:6379 \
  --messageBusAddress redis://redis:6379 \
  --dataSourceUrl "jdbc:postgresql://postgres/datashare?user=admin&password=admin" \
  --busType REDIS \
  --mode SERVER
```

### Start instance 2 (port 8081)
```bash
./run.sh app start \
--auth form \
--authUsersProvider database \
-d ~/Datashare/Data/ \
--redisAddress redis://redis:6379 \
--messageBusAddress redis://redis:6379 \
--dataSourceUrl "jdbc:postgresql://postgres/datashare?user=admin&password=admin" \
--busType REDIS \
--mode SERVER \
--port 8081
```
Same flags, add `--port 8081`.

### Test steps

1. Log in to instance 1 at `http://localhost:8080` with jdoe / secret.
2. Grant from CLI:
   ```bash
   ./run.sh project grant --project local-datashare --user jdoe \
     --authUsersProvider database \
     --dataSourceUrl "jdbc:postgresql://postgres/datashare?user=admin&password=admin" \
     --busType REDIS \
     --redisAddress redis://redis:6379
   ```
3. Check instance 2 logs for `policy reloaded`.
4. Check instance 1 logs — no reload (self-suppression).
5. Log in to instance 2 at `http://localhost:8081` and confirm `local-datashare` is accessible.
6. Revoke and repeat verification.
```bash
./run.sh project revoke --project local-datashare --user jdoe \
  --authUsersProvider database \
  --dataSourceUrl "jdbc:postgresql://postgres/datashare?user=admin&password=admin" \
  --busType REDIS \
  --redisAddress redis://redis:6379
```

---

## Scenario 3 — Form auth, Redis user provider

Users are stored in Redis. Authentication uses the form login filter with password hashed in Redis.

### Create the test user

```bash
./run.sh user create \
  --login jdoe \
  --email jdoe@example.org \
  --name "Jane Doe" \
  --password secret \
  --provider local \
  --authUsersProvider redis \
  --redisAddress redis://redis:6379 \
  --dataSourceUrl "jdbc:postgresql://postgres/datashare?user=admin&password=admin"
```

Confirm: `redis-cli -h redis GET jdoe` should return a JSON blob with `uid`, `email`, and `password`.

### Start instance 1 (port 8080)

```bash
./run.sh app start \
  --auth form \
  --authUsersProvider redis \
  -d ~/Datashare/Data/ \
  --redisAddress redis://redis:6379 \
  --messageBusAddress redis://redis:6379 \
  --dataSourceUrl "jdbc:postgresql://postgres/datashare?user=admin&password=admin" \
  --busType REDIS \
  --mode SERVER
```

### Start instance 2 (port 8081)
```bash
./run.sh app start \
--auth form \
--authUsersProvider redis \
-d ~/Datashare/Data/ \
--redisAddress redis://redis:6379 \
--messageBusAddress redis://redis:6379 \
--dataSourceUrl "jdbc:postgresql://postgres/datashare?user=admin&password=admin" \
--busType REDIS \
--mode SERVER \
--port 8081
```
Same flags, add `--port 8081`.

### Test steps

1. Log in to instance 1 at `http://localhost:8080` with jdoe / secret.
2. Grant from CLI:
   ```bash
   ./run.sh project grant --project local-datashare --user jdoe \
     --authUsersProvider redis \
     --redisAddress redis://redis:6379 \
     --dataSourceUrl "jdbc:postgresql://postgres/datashare?user=admin&password=admin" \
     --busType REDIS
   ```
3. Check instance 2 logs for `policy reloaded`.
4. Check instance 1 logs — no reload (self-suppression).
5. Log in to instance 2 at `http://localhost:8081` and confirm `local-datashare` is accessible.
6. Revoke and repeat verification.
   ```bash
   ./run.sh project revoke --project local-datashare --user jdoe \
     --authUsersProvider redis \
     --redisAddress redis://redis:6379 \
     --dataSourceUrl "jdbc:postgresql://postgres/datashare?user=admin&password=admin" \
     --busType REDIS
   ```
---

## Expected log output on instance 2 after grant

```
INFO  PolicyWatcher - Received policy-update notification from remote instance, reloading
```

## Expected log output on instance 1 after grant (self-suppression)

Nothing — own notifications are suppressed at DEBUG level only. No INFO line appears.

datashare-app/src/main/java/org/icij/datashare/mode/CommonMode.java
```diff
@@ -308,11 +308,16 @@ AsyncSearchStore provideAsyncSearchStore(final PropertiesProvider propertiesProv
     @Provides @Singleton
     Authorizer provideAuthorizer(CasbinRuleAdapter adapter, Provider<RedissonClient> redissonProvider) throws IOException {
         if (QueueType.REDIS.name().equals(propertiesProvider.get(BUS_TYPE_OPT).orElse(null))) {
+            // Event-driven: RTopic notifies all instances immediately on policy change.
             PolicyWatcher watcher = new PolicyWatcher(redissonProvider.get());
             addCloseable(watcher);
```
lschim: Question : Shouldn't you close the Authorizer since it is now closeable, and possibly close the watcher in cascade ?

datashare-app/src/main/java/org/icij/datashare/policies/PolicyWatcher.java
```diff
@@ -0,0 +1,71 @@
...
+    private final String instanceId = UUID.randomUUID().toString();
+    private volatile int listenerId = -1;
+
+    public PolicyWatcher(RedissonClient redisson) {
+        this.topic = redisson.getTopic(TOPIC);
+    }
+
+    @Override
+    public void setUpdateCallback(Runnable callback) {
+        listenerId = topic.addListener(String.class, (ch, msg) -> {
```
pirhoo: suggestion: remove the prior listener before registering a new one, e.g. guard with `if (listenerId != -1) topic.removeListener(listenerId);` at the top of each overload, or track a `List<Integer>` and remove all on `close()`.

datashare-app/src/main/java/org/icij/datashare/policies/Authorizer.java
```diff
@@ -29,7 +30,7 @@
 import static org.icij.datashare.policies.errors.UnknownRoleException.resolveRole;
 
 @Singleton
-public final class Authorizer {
+public final class Authorizer implements Closeable {
 
     private final static String SEPARATOR = "::";
```
lschim: Hint : Absolutely no big deal, but the JRE specs mentions that the customary order for modifiers is `static final` rather than `final static` https://docs.oracle.com/javase/specs/jls/se21/html/jls-8.html#jls-8.3.1

datashare-app/src/main/java/org/icij/datashare/mode/CommonMode.java
```diff
@@ -304,6 +310,27 @@ AsyncSearchStore provideAsyncSearchStore(final PropertiesProvider propertiesProv
         };
     }
 
+    @Provides @Singleton
+    Authorizer provideAuthorizer(CasbinRuleAdapter adapter, Provider<RedissonClient> redissonProvider) throws IOException {
+        if (QueueType.REDIS.name().equals(propertiesProvider.get(BUS_TYPE_OPT).orElse(null))) {
+            // Event-driven: RTopic notifies all instances immediately on policy change.
+            // startAutoLoadPolicy backstop defaults to 0 (disabled); set policyReloadInterval
+            // explicitly to enable periodic reloads as a hedge against missed publications.
+            long interval = parseInt(propertiesProvider.get(POLICY_RELOAD_INTERVAL_OPT).orElse("0"));
```
pirhoo, line 319: suggestion: this creates a discreptencies with the default values in the legacy joptsimple parser, we could change it to:

```
long interval = parseInt(propertiesProvider.get(POLICY_RELOAD_INTERVAL_OPT)
        .orElse(String.valueOf(DEFAULT_POLICY_RELOAD_INTERVAL)));
```

Where `DEFAULT_POLICY_RELOAD_INTERVAL` is 30_000. This is important to change so default value is the same both with joptsimple and picocli.

pirhoo (commented):
Great work!

I've made only two non-blocking suggestions :)

datashare-app/src/main/java/org/icij/datashare/mode/CommonMode.java
caro3801: This is a particular case, in event driven, we don't want backstop polling policy reload by default (0), unless specified.
Howere joptsimple and picocli should behave the same, I'm fixing the initial value in joptsimple

datashare-app/src/test/java/org/icij/datashare/policies/AuthorizerTest.java
```diff
@@ -389,4 +392,29 @@ public void test_require_id_param() {
...
+        Watcher mockWatcher = Mockito.mock(Watcher.class);
+        Authorizer watchedAuthorizer = new Authorizer(adapter, mockWatcher);
+
+        watchedAuthorizer.addRoleForUserInProject(user, Role.PROJECT_MEMBER, domain, project);
+
+        verify(mockWatcher, atLeastOnce()).update();
+    }
+
+    @Test
+    public void test_polling_authorizer_close_stops_autoload() throws Exception {
```
lschim, line 414: Suggestion : This test is currently doing nothing (I tried removing the `stopAutoLoadPolicy()` in the `close()` method of the authorizer and the test still passes.
We need something like :
```
        try (MockedConstruction<SyncedEnforcer> mockedConstruction = Mockito.mockConstruction(SyncedEnforcer.class, (mock, ctx) -> {
            when(mock.getRoleManager()).thenReturn(new DomainManager(10, null, BuiltInFunctions::allMatch));
        })) {
            Authorizer pollingAuthorizer = Mockito.spy(new Authorizer(adapter, 100L));
            SyncedEnforcer mockedEnforcer = mockedConstruction.constructed().getFirst();
            pollingAuthorizer.close();
            verify(mockedEnforcer).stopAutoLoadPolicy();
        }
```

datashare-app/src/test/java/org/icij/datashare/policies/AuthorizerTest.java
```diff
@@ -389,4 +392,29 @@ public void test_require_id_param() {
...
+        CasbinRuleAdapter adapter = Mockito.mock(CasbinRuleAdapter.class);
+        Authorizer pollingAuthorizer = new Authorizer(adapter, 100L);
+        pollingAuthorizer.close();
+    }
+
+    @Test
+    public void test_zero_interval_does_not_start_polling_and_close_is_safe() throws Exception {
+        CasbinRuleAdapter adapter = Mockito.mock(CasbinRuleAdapter.class);
```
lschim, line 428: Suggestion : Same comment as L407, I think this test checks nothing, it probably needs to be reworked

datashare-app/src/main/java/org/icij/datashare/policies/PolicyWatcher.java
```diff
@@ -18,17 +25,21 @@ public PolicyWatcher(RedissonClient redisson) {
 
     @Override
     public void setUpdateCallback(Runnable callback) {
-        listenerId = topic.addListener(String.class, (ch, msg) -> callback.run());
+        listenerId = topic.addListener(String.class, (ch, msg) -> {
+            if (!msg.startsWith(instanceId)) callback.run();
```
lschim: Suggestion : It feels a bit weak and sounds like an "unwritten" convention to rely on the UUID to ignore at the beginning of the string to avoid notifying the sender.
A possibility could be to trigger a message of a class made for this :

```public record PolicyUpdateMessage(String callerID, String message);```

and then use it

```topic.addListener(PolicyUpdateMessage.class, (ch, msg) -> { if (!this.instanceId.equals(msg.calledID)) { ... } }```

File: datashare-app/src/test/java/org/icij/datashare/policies/AuthorizerRedisSyncIntTest.java [hunk 8]
```diff
@@ -0,0 +1,84 @@
...
+        instance2.close();
+        watcher1.close();
+        watcher2.close();
+        client1.shutdown();
+        client2.shutdown();
+    }
+
+    @Test
+    public void policy_change_on_instance1_notifies_instance2() throws Exception {
+        CountDownLatch latch = new CountDownLatch(1);
```
lschim, line 55: Remark : Cool test, a bit technical, but easy to read !

datashare-cli/src/main/java/org/icij/datashare/cli/command/GlobalOptions.java
```diff
@@ -144,6 +147,7 @@ public Properties toProperties() {
         DatashareOptions.put(props, REDIS_POOL_SIZE_OPT, redisPoolSize);
         DatashareOptions.putIfNotNull(props, MESSAGE_BUS_OPT, messageBusAddress);
         DatashareOptions.putIfNotNull(props, BUS_TYPE_OPT, busType);
+        DatashareOptions.put(props, POLICY_RELOAD_INTERVAL_OPT, policyReloadInterval);
```
lschim: Question : In which mode do we want this option to be available ? I get the feeling it is only for Server mode (which means it should go in the ServerOptions class), but I might be wrong

datashare-app/src/test/java/org/icij/datashare/project/admin/ProjectAdminServiceImplTest.java
```diff
@@ -31,10 +32,7 @@
 import static org.junit.Assert.fail;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.eq;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.never;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
lschim, line 37: Remark : Soon there will be a linter that checks the code style, so this will not be allowed anymore, because I think ICIJ code style in Java prevents the use of * imports ;)

datashare-cli/src/main/java/org/icij/datashare/cli/command/GlobalOptions.java
```diff
@@ -115,6 +115,9 @@ private static String userHome() {
     @Option(names = {"-l", "--language"}, description = "Indexing language", scope = ScopeType.INHERIT)
     String language;
 
+    @Option(names = {"--authUsersProvider"}, description = "Auth users provider class", scope = ScopeType.INHERIT)
+    String authUsersProvider;
```
lschim, line 119: Question : Same question as for L153, In which mode do we want this option to be available ? I get the feeling it is only for Server mode (which means it should go in the ServerOptions class), but I might be wrong

datashare-cli/src/main/java/org/icij/datashare/cli/command/ServerOptions.java
```diff
@@ -204,6 +213,8 @@ public Properties toProperties() {
 
         DatashareOptions.putIfNotNull(props, REPORT_NAME_OPT, reportName);
         DatashareOptions.putIfNotNull(props, SMTP_URL_OPT, smtpUrl);
+        DatashareOptions.putIfNotNull(props, TEMPORAL_NAMESPACE_OPT, temporalNamespace);
```
lschim: Important : This looks like a rollback from a commit that was made on main. TEMPORAL_NAMESPACE_OPT and BATCH_QUEUE_TYPE_OPT should not be in ServerOptions but in GlobalOptions because they are required in both SERVER and WORKER mode

datashare-app/src/main/java/org/icij/datashare/session/UsersInRedis.java
```diff
@@ -51,15 +51,22 @@ void removeUser(String login) {
...
+            // Rule: extend-only — never shorten an existing session or add an expiry
+            // to a persistent key. Prevents the CLI (sessionTtlSeconds=1) from
+            // logging out OAuth2 users with active sessions or erasing persistent users.
             long existingTtl = jedis.ttl(key);
             Transaction transaction = jedis.multi();
             transaction.set(key, JsonObjectMapper.serialize(((DatashareUser)user).details));
-            if (existingTtl != -1L) {
+            if (existingTtl == -2L) {
                 transaction.expire(key, this.ttl);
+            } else if (existingTtl > 0L) {
```
lschim: Remark : I don't master Redis very well, but I wonder though, shouldn't this condition be 
`existingTtl >= 0L` ? Just in case sometimes Redis stores an expired key over its limit (with a TTL that is equal to 0) and garbage collect at some point ? 
I feel case disjunction should always be "safe" by default, meaning that in any case there should be an expiry EXCEPT when the existingTtl is -1, instead of the code written here which by default doesn't put an expiry date and adds it for certain cases. Altough this becoming a vulnerability is very unlikely, what's more with the transaction, but still I guess it could lead to weird cases of inifinite user sessions

datashare-app/src/main/java/org/icij/datashare/policies/PolicyWatcher.java
```diff
@@ -39,7 +42,16 @@ public void setUpdateCallback(Consumer<String> callback) {
 
     @Override
     public void update() {
-        topic.publish(instanceId + ":reload");
+        try {
+            topic.publish(instanceId + ":reload");
+        } catch (Exception e) {
```
lschim, line 59: Remark : Nice catch 👍 !

datashare-app/src/main/java/org/icij/datashare/mode/CommonMode.java
```diff
@@ -367,7 +368,20 @@ Class<? extends Users> resolveUsersProviderClass() {
...
+                .filter(m -> !m.isEmpty())
+                .map(m -> { try { return AuthMode.fromString(m) == AuthMode.OAUTH; } catch (IllegalArgumentException e) { return false; } })
+                .orElse(false);
+        if (isOAuth) {
+            return injector.getInstance(UsersIdProviderRedisCache.class);
+        }
+        // Non-OAuth2 auth modes have no Redis session cache: saveOrUpdate is a no-op.
+        // Reads are handled by the Users binding (provideUsers); these find() methods are unused.
+        return new UsersIdProviderCache() {
+            @Override public net.codestory.http.security.User find(String login) { return null; }
```
lschim, line 410: question : I am not exactly sure of the full flow, but shouldn't this anonymous instance forward the result of the UsersProvider for find methods ?

datashare-cli/src/main/java/org/icij/datashare/cli/command/ServerOptions.java
lschim: I saw it was rollbacked in a following commit, but that makes me wonder was that a git mismanipulation in the first place ?

datashare-app/src/main/java/org/icij/datashare/mode/CommonMode.java
caro3801: The closeable for authorizer and watcher are registered in reverse order so the authorizer shuts down before the watcher unsubscribe from Redis.
However, I'm taking your suggestion of closing by cascade, by making the Authorizer responsible for closing its watcher, it'll add cleaner ownership.

datashare-app/src/main/java/org/icij/datashare/policies/Authorizer.java
caro3801: good to know thanks! I changed it across the project

datashare-app/src/main/java/org/icij/datashare/mode/CommonMode.java
caro3801: The `no-op` cache returning `null` from `find()` is intentional, it signals "cache miss" to callers. 
In the non-Redis path, `provideUsers` short-circuits to return `userStore` directly (line 365), so the cache is never consulted for lookups at all. Forwarding to the store here would cause a double lookup in the Redis path. I'm adding a comment in that sense

datashare-app/src/test/java/org/icij/datashare/policies/AuthorizerTest.java
caro3801: yes, you are right, i forgot to update them .. Thanks for the code suggestion

datashare-app/src/test/java/org/icij/datashare/policies/AuthorizerTest.java
caro3801: Same, thanks for spotting it

datashare-cli/src/main/java/org/icij/datashare/cli/command/GlobalOptions.java
caro3801: The options should also be available in CLI mode the option. The project/user management is not using the server mode should be available so it requires to appear in Global (server and cli) or to duplicate it into both mode. First option will not have side effect on local mode that why I chose it but it's open for discussion

datashare-cli/src/main/java/org/icij/datashare/cli/command/ServerOptions.java
caro3801: yes it was

datashare-app/src/main/java/org/icij/datashare/session/UsersInRedis.java
caro3801: Yes this was an issue, it has been later fixed in https://github.com/ICIJ/datashare/blob/feat/casbin-redis-watcher/datashare-app/src/main/java/org/icij/datashare/session/UsersIdProviderRedisCache.java#L61

pirhoo (approved):

datashare-app/src/main/java/org/icij/datashare/policies/PolicyWatcher.java
caro3801: I added it! Thanks for the suggestion

lschim (approved):
Great work !!

PR #2173, opened 2026-06-03
feat: auth users provider enum

pirhoo:
Following up #2172, this PR uses an enum for `--authUsersProvider` too and change the default values to user the "database" by default and the the "form" as auth mechanism. The `--authUsersProvider` param now accept "database" and "redis" as provider, and is retro compatible with the old class notation.

## AI Disclosure

Entirely done using superpowers, based on #2172, supervised and reviewed by a human before submitting the PR.


lschim (approved):
LGTM

mvanzalu (approved):

PR #2175, opened 2026-06-04
feat: add --temporalAddress option

lschim:
Added in picocli as a globalOption for it to be available in SERVER and WORKER commands.
Added in legacy CLI
Homogenize the option across the platform

### DIFF
--- a/datashare-cli/src/main/java/org/icij/datashare/cli/DatashareCliOptions.java
+++ b/datashare-cli/src/main/java/org/icij/datashare/cli/DatashareCliOptions.java
@@ -127,6 +127,7 @@ public final class DatashareCliOptions {
     public static final String SETTING_ABBR_OPT = "s";
     public static final String SMTP_URL_OPT = "smtpUrl";
     public static final String TEMPORAL_NAMESPACE_OPT = "temporalNamespace";
+    public static final String TEMPORAL_ADDRESS_OPT = "temporalAddress";
     public static final String VERSION_ABBR_OPT = "v";
     public static final String VERSION_OPT = "version";
     public static final String ARTIFACT_DIR_OPT = "artifactDir";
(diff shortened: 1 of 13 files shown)


datashare-cli/src/main/java/org/icij/datashare/cli/command/GlobalOptions.java
```diff
@@ -114,6 +115,16 @@ private static String userHome() {
     @Option(names = {"--no-color"}, description = "Disable ANSI color output", defaultValue = "false", scope = ScopeType.INHERIT)
     boolean noColor;
 
+    // Batch / Download
+    @Option(names = {"--batchQueueType"}, description = "Batch queue type", defaultValue = "MEMORY", scope = ScopeType.INHERIT)
+    QueueType batchQueueType;
+
+    @Option(names = {"--temporalAddress"}, description = "Temporal address", scope = ScopeType.INHERIT)
+    String temporalAddress = EnvUtils.resolveUri("temporal", "temporal:7233");
```
pirhoo, line 123: hint: I guess that means temporalAddress can *never* be empty and the putIfNotNull line 160 might shadow the default value.

datashare-cli/src/main/java/org/icij/datashare/cli/DatashareCliOptions.java
```diff
@@ -1036,6 +1037,14 @@ static void temporalNamespaceOpt(OptionParser parser) {
                 .defaultsTo("datashare-default");
     }
 
+    static void temporalAddressOpt(OptionParser parser) {
+        parser.acceptsAll(
+                List.of(TEMPORAL_ADDRESS_OPT), "Temporal address")
+                .withRequiredArg()
+                .ofType( String.class )
+                .defaultsTo("temporal:7233");
```
pirhoo: suggestion: this breaks the file convention to store default value in DEFAULT_* constants. We might do some girl-scouting here and apply it to other default values.

pirhoo (approved):
Only two small suggestions but it's good to go!

datashare-cli/src/main/java/org/icij/datashare/cli/DatashareCliOptions.java
lschim: Noted, the file was girl-scouted to use DEFAULT_* constants !

datashare-cli/src/main/java/org/icij/datashare/cli/command/GlobalOptions.java
lschim: After review with Pierre, it was determined that it would stay that way because the picocli default value in the annotation needs to be a constant, meaning we won't be able to get the value from environment variables or conf files. Also, all the other address options were written exactly the same way

PR #2178, opened 2026-06-08
feat: update the progress field of tasks started using Temporal

### LINKED ISSUE #2185: feat: Enable progress for Java Tasks when using Temporal
Currently the progress is stored in the TaskRepository. If Datashare crashes, upon restart, it would good to retrieve the value from Temporal (as most likely the Task will have kept progressing). 

Also, Python workflows store the progress in the SEARCH_ATTRIBUTES, so Java backend should be able to read them (There was an attempt of an implementation already in place, but it was never tested and was not working as expected)

Search attributes that need to be read :
- Progress
- MaxProgress

In addition, Java workers should be able to update those search attributes to.

For now all the java workflows are singled-activity, so the progress of the workflow is going to be exactly the one of the activity it contains.

If one day, multi activity workflows are created in Java, the notion of weight can be reintroduced so that an activity can update a workflow from 0 to 20% for instead, and the second activity from 0 to 80%.

lschim:
To update the progress field of tasks that are viewed in the frontend, Activities need an access to taskRepository so that it can update the field. 

The PR also fixes the issue where the update of the task in repository upon completion in Temporal overrides all the values, including progress in repository (which results in a 0% progress rate after completion)

The PR fixes https://github.com/ICIJ/datashare/issues/2185

Used Claude to replace all the calls to TemporalActivityImpl that needed TaskRepository, and to detect why adding the TemporalWorkflow interface made the worker crash

### DIFF
--- a/datashare-tasks/src/main/java/org/icij/datashare/asynctasks/temporal/TemporalActivityImpl.java
+++ b/datashare-tasks/src/main/java/org/icij/datashare/asynctasks/temporal/TemporalActivityImpl.java
@@ -7,6 +7,7 @@ import io.temporal.activity.Activity;
 import io.temporal.activity.ActivityInfo;
 import io.temporal.client.WorkflowClient;
 import io.temporal.client.WorkflowStub;
+import java.io.IOException;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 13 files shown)


datashare-tasks/src/main/java/org/icij/datashare/asynctasks/temporal/TemporalActivityImpl.java
```diff
@@ -38,8 +45,18 @@ protected TaskFactory getTaskFactory() {
...
-        return (ignored, progress) -> workflow.signal("progress",
-            new ProgressSignal(info.getRunId(), info.getActivityId(), progress, this.progressWeight));
+        String taskId = info.getWorkflowId();
+        return (ignored, progress) -> {
+            workflow.signal("progress", new ProgressSignal(info.getRunId(), info.getActivityId(), progress, this.progressWeight));
+            try {
+                logger.debug("Setting progress of {} to {}", taskId, progress);
+                Task<?> task = taskRepository.getTask(taskId);
+                task.setProgress(progress);
+                taskRepository.update(task);
```
ClemDoum, line 55: This will work for Java tasks but not the Python tasks.
I think we still need the custom search attribute progress handler for this

ClemDoum: Actually this how Python progress works:
- the activity worker emits a progress signal (+ an heartbeat): https://github.com/ICIJ/datashare-python/blob/c2bfa86da70bc1f339ee01131a17e3054f1dbe33/datashare-python/datashare_python/utils.py#L144
- the Python workflow worker receives the signal and updates the workflow's custom search attribute: https://github.com/ICIJ/datashare-python/blob/c2bfa86da70bc1f339ee01131a17e3054f1dbe33/datashare-python/datashare_python/utils.py#L85

If we want progress to be reflected inside DS we need to read it from the workflow's custom search attributes

lschim: In Java too we use the signal, you can see it l.50.

But the use of the search attribute was not working so far, as we discussed together, neither the update or the read. I fixed it in the second commit of the PR. 

For now though, we don't need to display the progress of python tasks in datashare frontend, but once we need to display it, we will be able to do so using the search attribute

ClemDoum (approved):
Well done ! Can't wait to have this live !

Issue #2168, opened 2026-06-02
bug: wrong number of 'no-match queries' in batch search


Soliine:
**To Reproduce**
1. Go to [this demo batch search](https://datashare-demo.icij.org/#/tasks/batch-search/luxleaks/4c8882c4-edb4-49b8-af69-46508bf1ece3/queries?sort=query_results&order=desc&perPage=100)
2. See that there are 37 no-match queries (queries with 0 results)
3. See error: in the batch search card, it is displayed 'Download 72 no-match queries' while it should display 'Download 37 no-match queries'

**Expected behavior**
Display the right number of no-match queries, not the total number of queries.

(NB: the CSV with no-match queries has no error, this is ok)

**Screenshots**
<img width="1165" height="723" alt="Image" src="https://github.com/user-attachments/assets/11db6341-e35d-4b20-af05-a2b77e00dd46" />

**Desktop (please complete the following information):**
 - OS: iOS
 - Browser: Helium
 - Version: 21.6.0

PR #2180, opened 2026-06-09
fix: report actual nbQueriesWithoutResults in batch search task result

pirhoo:
As described, in #2168, the `nbQueriesWithoutResults` value is currently wrong.

This PR increment this value correctly and add tests to make sure there will be no regressions again.

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/tasks/BatchSearchRunner.java
+++ b/datashare-app/src/main/java/org/icij/datashare/tasks/BatchSearchRunner.java
@@ -82,6 +82,7 @@ public class BatchSearchRunner implements CancellableTask, UserTask, Callable<Ba
     public BatchSearchRunnerResult call() throws Exception {
         int numberOfResults = 0;
         int totalProcessed = 0;
+        int queriesWithoutResults = 0;
 
         int throttleMs = parseInt(propertiesProvider.get(BATCH_THROTTLE_OPT).orElse(DEFAULT_BATCH_THROTTLE));
         int maxTimeSeconds = parseInt(propertiesProvider.get(BATCH_SEARCH_MAX_TIME_OPT).orElse(DEFAULT_BATCH_SEARCH_MAX_TIME));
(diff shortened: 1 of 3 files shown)


bamthomas (approved):

datashare-app/src/main/java/org/icij/datashare/tasks/BatchSearchRunner.java
```diff
@@ -157,7 +164,7 @@ public BatchSearchRunnerResult call() throws Exception {
             repository.setState(taskView.id, searchException);
             throw searchException;
         }
-        return new BatchSearchRunnerResult(numberOfResults, batchSearch.nbQueriesWithoutResults);
```
lschim, line 160: Question : Viewing your comment, NB_QUERIES_WITHOUT_RESULTS is not the source of truth for the number of queries without results. Would it be interesting to manually set NB_QUERIES_WITHOUT_RESULTS before returning to ensure that the values match the one stored in the task ? Would be useful if at some point the way to compute them differs ?

I don't think it is required, but having two values that should represent the same thing, without having them computed with the same function always feel a bit strange...

lschim: Another possibility would be to avoid computing the value within this function, and before returning, reload batchSearch from DB to retrieve the final value of NB_QUERIES_WITHOUT_RESULTS

pirhoo: You're right, and it's actually worse than "two values that might diverge": the value we return today is **stale**, not just computed differently.

The result we return reflects the starting state, not the outcome.

So the DB column is unambiguously the source of truth. It's incrementally maintained, transactional, and already handles the tricky parts (multi-scroll via `isFirstScroll`, the `MAX_BATCH_RESULT_SIZE` cutoff).

Between your two options, I'd go with reload from DB (option 2). It keeps a single computation. Something like:

```java
repository.setState(batchSearch.uuid, BatchSearchRecord.State.SUCCESS);
// ...
int nbQueriesWithoutResults = repository.get(taskView.getUser(), taskView.id).nbQueriesWithoutResults;
return new BatchSearchRunnerResult(numberOfResults, nbQueriesWithoutResults);
```

The cost is one extra read at the end of a (potentially multi-day) batch so it's negligible. In exchange the runner reports exactly what's persisted, by definition, with no second function that "should" agree.

One caveat for the thread: this also means `BatchSearchRunnerTest.java` needs to stop relying on the mock returning the unchanged object. It currently passes because the value is stale. That's a good sign the change fixes real behavior, but the test will need a mock that reflects the post-run count.

Since the PR is already merge, I'm going to open a follow up PR.

pirhoo: It's here https://github.com/ICIJ/datashare/pull/2181

Issue #2166, opened 2026-06-02
feat: edit a batch search


Soliine:
As a user, I edit my batch search (screen 19.6 in Figma):

<img width="865" height="511" alt="Image" src="https://github.com/user-attachments/assets/9051c08e-a7da-4fb8-96c9-11d0453c6b48" />

When I click 'Edit' (button with PencilSimple icon, between Relaunch and Delete), I can change everything in the form (name, visibility, description, filters, etc).

At the end of the form, there are 2 buttons:
- 'Cancel' (.btn-outline-secondary) which closes the form and leads to the batch searches page
- 'Save' (.btn-action) which relaunches the batch search with the new parameters (and thus replace the previous one). It does not change the creation date of the batch search (so users still find it when it was originally launched).

pirhoo:
We refine the feature to only edit title, description and visibility:

<img width="1400" height="780" alt="Image" src="https://github.com/user-attachments/assets/44057327-9e67-46d1-b083-871d165294ea" />

Soliine:
I forgot to change the navigation breadcrumb, it should be : Tasks > Batch searches > [name of the batch search] > Edit your batch search (or simply "Edit"). Thanks!

PR #2183, opened 2026-06-09
feat: batch search edit title and description

pirhoo:
As part of #2166, this PR allow users to edit name and description of a batch search through the API.

What began as "let users edit a batch search's title and description over the API" turned out to touch every layer the field validation has to hold at. Editing names/descriptions meant the rules (non-empty name, max lengths) couldn't live only at the endpoint: they had to be enforced consistently wherever those fields are written, so the validators were added to the `BatchSearch` model and then applied not just on update but on the create and copy paths in `JooqBatchSearchRepository`, each returning a proper 400 instead of a 500.

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/web/TaskResource.java
+++ b/datashare-app/src/main/java/org/icij/datashare/web/TaskResource.java
@@ -244,7 +244,7 @@ public class TaskResource {
                     in = ParameterIn.PATH, examples = @ExampleObject(value = "prj1,prj2"))}
     )
     @ApiResponse(responseCode = "413", description = "if the CSV file is more than 60K lines")
-    @ApiResponse(responseCode = "400", description = "if either name or CSV file is missing")
+    @ApiResponse(responseCode = "400", description = "if name or CSV file is missing, or name/description are empty or exceed their maximum length")
     @Post("/batchSearch/:coma_separated_projects")
     public Payload search(String comaSeparatedProjects, Context context) throws IOException {
         List<Part> parts = context.parts();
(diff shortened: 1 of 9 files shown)


datashare-db/src/main/java/org/icij/datashare/db/JooqBatchSearchRepository.java
```diff
@@ -357,6 +357,14 @@ public boolean setName(User user, String batchId, String name) {
                         and(BATCH_SEARCH.USER_ID.eq(user.id))).execute() > 0;
     }
 
+    @Override
+    public boolean setDescription(User user, String batchId, String description) {
```
lschim, line 361: thought : I have often seen in the past patterns where the repository provides an "update" method (which modifies all the allowed fields for modification) and it is the service layer that builds the object to update. It reduces the number of calls to the DB if the user modifies 3 fields at the same time, and the code is a bit easier to read in the service layer than at the repository level. But since we don't really have a service layer, I guess it makes sense to keep it that way

pirhoo: I hesitated too between using update and setter methods. But with the current implementation, we can potentially update the record partially. I'm still not sure if it's a good or a bad thing!

datashare-app/src/main/java/org/icij/datashare/web/TaskResource.java
```diff
@@ -271,7 +271,12 @@ public Payload search(String comaSeparatedProjects, Context context) throws IOEx
 
         BatchSearch batchSearch = new BatchSearch(stream(comaSeparatedProjects.split(",")).map(Project::project).collect(Collectors.toList()), name, description, queries,uri,
                 (User) context.currentUser(), published, fileTypes, queryTemplate, paths, fuzziness,phraseMatches);
-        boolean isSaved = batchSearchRepository.save(batchSearch);
+        boolean isSaved;
+        try {
+            isSaved = batchSearchRepository.save(batchSearch);
+        } catch (IllegalArgumentException e) {
```
lschim, line 277: question : Wouldn't it be interesting to implement that at the filter level, so that any IllegalArgumentException will always return an http code 400 ?

pirhoo: I totally agree but for the sake of simplicity, I rather keep it out of this PR. Since `IllegalArgumentException` is thrown in a lot of places in the app, I want to avoid the "blast radius" effect of that change.

It would be even better to catch it at a filter level but throw custom specific exception type to customize messages.

lschim (approved):

datashare-db/src/main/java/org/icij/datashare/db/JooqBatchSearchRepository.java
lschim: Let's see in the future, this is a very easy to change pattern

PR #2210, opened 2026-06-18
Reduce the number of connection reset datashare gets

lschim:
With the recent changes in the network configuration (starting around the 13th of may), Datashare started getting `IOException : Connection reset` upon sending requests to ES.
It was due to the fact that there is in DS a connection pooling mechanism in place which kept connections to ES alive for probably 10 minutes (we were never able to find exactly the value, we think there is a chance it is linked to the scroll duration, sent back by ES in an header) BUT in production and staging the load balancer would close connections after 350 seconds.
So Datashare would try to reuse a closed connection, and get a connection reset on its first attempt.
Since they are not handled gracefully (at least in BatchDownload), the Task would fail. 

This PR adds a configuration entry `elasticsearchMaxIdleConnectionTime` which sets the TTL of an idling ES connection, adding the capability to adjust to the network infra. In particular it is recommended to set the TTL of those idling connections lower than the value of the network elements that would shut idling connections (proxy, load balancer etc...)

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/tasks/BatchDownloadRunner.java
+++ b/datashare-app/src/main/java/org/icij/datashare/tasks/BatchDownloadRunner.java
@@ -145,6 +145,7 @@ public class BatchDownloadRunner implements Callable<BatchDownloadRunnerResult>,
                 }
             }
         } catch (ElasticsearchException esEx) {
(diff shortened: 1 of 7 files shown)


pirhoo (approved):
Very neat, thanks!

Issue #2229, opened 2026-06-24
epic(artifacts): as a user I can generate and access per-document artifacts (raw, page, structure)


pirhoo:
An **artifact** is a derived representation of a document, produced alongside it, generatable on demand, and meant to be **stored** under `artifactDir` (keyed by content hash) rather than recomputed each time. Three types: **raw** (done), **page** (computed live today, persist on disk), **structure** (Tika XHTML+MD, from the [PoC](https://github.com/ICIJ/datashare/tree/poc/tika-document-structure-extraction)). All on a shared, extensible subsystem, served from a dedicated API. Layout/manifest follow the document-artifacts-convention spec; the API deviates to `/:project/artifacts/<type>/:id`.

```
<artifactDir>/<project>/<id[0:2]>/<id[2:4]>/<id>/
├── raw  
├── pages/page-0001.txt  
├── structure/page-0001.{xhtml,md}  
└── manifest.json
```

Decisions: 
* `manifest.json` replaces the PoC `.complete` marker 
* structure is per-page (`?format=md|xhtml`) 
* generation gated by `--artifacts` 
* serving on a dedicated resource. 

Out of scope: `thumbnail` (extension point), client viewer, full python alignment.

**Cross-repo dependency (not a sub-issue).** Full python alignment stays out of scope, but one piece of it blocks interop: datashare-python writes the structure entry's page count as a nested `pages: {type, total, byteRanges}`, while the convention and the Java side use entry-level `total` + `pagination`. Until [datashare-python#67](https://github.com/ICIJ/datashare-python/issues/67) lands, neither producer can read the other's page count. Java no longer destroys the fields it cannot read (the manifest is merged at tree level), so this is a read gap, not data loss.

Issue #2224, opened 2026-06-24
feat(artifacts): per-document manifest.json


pirhoo:
A `Manifest` class owning `manifest.json`, replacing the PoC `.complete` marker. One entry per type, extensible, concurrency-safe.

- [ ] Entry: `status` / `pages` / `paginated` / `taskInput` (+ optional `confidence`, `label`, `raw`'s `contentType`); tolerate unknown fields.
- [ ] Concurrency-safe write (lock or atomic temp+rename); each task touches only its own entry.
- [ ] Payload first, then entry last with `status: complete`; `skip-if-current` on matching `taskInput`.
- [ ] Remove `.complete` marker (`STRUCTURE_COMPLETE_MARKER`, `structureComplete`, `hasCompleteMarker`). Tests: concurrency, crash recovery, skip-vs-regenerate.

Refs: convention §Manifest / §Concurrency / §Completion.

Part of epic(artifacts).

pirhoo:
Manifest schema finalized in the **document-artifacts-convention** doc. Two changes vs. the original draft this issue referenced:

- **`total`** replaces the page-count field (was a self-referential `"pages": N` inside the `pages` entry).
- **`pagination`** object replaces the `paginated` boolean:
  - `{ "type": "filesystem" }`: one file per page, `page-%04d.<ext>` (the structure Tika case).
  - `{ "type": "byteRanges", "ranges": [[0,125],[125,39000]] }`: single `content.<ext>` split by half-open `[start, end)` offsets (contiguous, so `ranges[i][1] == ranges[i+1][0]`).
  - omitted: single document (`content.<ext>`, `total: 1`).

So an entry is `{ status, total, pagination?, taskInput, confidence?, label?, contentType?/filename? }`. The `Manifest` model should serialize these and tolerate unknown fields (extensibility AC unchanged).

Issue #2223, opened 2026-06-24
refactor(artifacts): type abstraction + registry


pirhoo:
Each artifact type becomes its own class behind a common `Artifact` interface, called by both stages and the API. Migrates the on-disk types that exist today (`raw`, `structure`); `page` lands in the page issue.

- [ ] `Artifact` interface: content-addressed path, `produce(doc, source)`, manifest entry.
- [ ] `RawArtifact` (wrap `SourceExtractor`), `StructureArtifact` (wrap `StructureMarkdownExtractor`).
- [ ] `ArtifactRegistry` + `--artifacts` selector (bare = all, CSV = subset, unknown = error).
- [ ] Move `ArtifactPath` `structure*` helpers behind the classes; keep generic `dir()`. Tests per class + selector.

Refactor only — no behavior change. Refs: `ArtifactPath`, `StructureMarkdownExtractor`, `ArtifactTask`.

Part of epic(artifacts).

PR #2231, opened 2026-06-25
feat: artifacts foundation

pirhoo:
Part of the artifacts subsystem epic #2229. Delivers the foundation it all builds on: the shared `Artifact` abstraction and registry (#2223) and the per-document `manifest.json` (#2224), with `RawArtifact` as the first live producer wired into the `ARTIFACT` stage.

This PR turns Datashare's per-document derived outputs into one shared, extensible artifact subsystem. A single `ArtifactRegistry` decides which types to produce (the `--artifacts` selector), skips anything already current, produces the rest, and records each in a per-document `manifest.json`. Types only produce payload files and describe their manifest entry; they never touch the manifest themselves.

## Design

```mermaid
flowchart TD
    Task["ArtifactTask (ARTIFACT stage):<br/>polls each document node"]
    Ctx["ArtifactContext<br/>(project, document, nodeDir, sources)"]
    Reg["ArtifactRegistry.run(selected, context)"]
    Cur{"isCurrent?<br/>entry complete AND taskInput matches"}
    Skip["skip (already current)"]
    Prod["Artifact.produce(context):<br/>writes payload files only"]
    Raw["RawArtifact: extractEmbeddedSources()<br/>writes raw / raw.json bytes"]
    Entry["returns ManifestEntry<br/>(or null: nothing to record)"]
    Store[("ManifestStore:<br/>the only reader/writer of manifest.json")]
    Manifest[/"manifest.json<br/>one entry per type, status=complete (written last)"/]

    Task --> Ctx --> Reg
    Reg -->|"for each selected type"| Cur
    Cur -->|yes| Skip
    Cur -->|no| Prod
    Prod --> Raw
    Prod --> Entry
    Entry -->|"non-null"| Store
    Store --> Manifest
    Cur -.->|"reads via get()"| Store
```

### manifest.json

```jsonc
// <artifactDir>/<project>/<id[0:2]>/<id[2:4]>/<id>/manifest.json
{
  "raw": {
    "status": "complete",
    "taskInput": {
      "type": "raw",
      "version": 1
    },
    "contentType": "application/pdf",
    "filename": "report.pdf"
  }
}
```

- `status: "complete"` (written last) is the only "ready" signal, replacing the PoC's `.complete` marker.
- Skip-if-current: a type is skipped only when its stored entry is `complete` and its `taskInput` (config plus an explicit version) deep-equals this run's, so a config or version change re-extracts.
- No producer name: an artifact's identity is its whole `taskInput`.

## CLI

This adds one new selector, `--artifacts`, to the existing `ARTIFACT` stage. It decides which types the registry produces for the run.

```bash
# All registered types (currently just "raw"). Bare flag and omitted flag are equivalent.
datashare stage run --stages ARTIFACT --artifacts
datashare stage run --stages ARTIFACT

# A specific subset (comma-separated, case-insensitive, trimmed)
datashare stage run --stages ARTIFACT --artifacts raw

# A typo fails loudly instead of silently producing nothing
datashare stage run --stages ARTIFACT --artifacts thumbnial
# => unknown artifact type 'thumbnial'; valid types: [raw]
```

What actually changes for someone running the stage:

| | Before | After |
| --- | --- | --- |
| Type selection | none: the stage always ran `extractEmbeddedSources` for every document | `--artifacts` picks types; bare or omitted means all registered types |
| Re-runs | re-extracted every document on every run | skip-if-current: a type is redone only when its `taskInput` (config plus version) changes |
| Record of work | none (PoC `.complete` marker) | per-document `manifest.json`, `status: "complete"` written last |
| Bad input | n/a | unknown or empty selection throws with the list of valid types |

The flag is optional and defaults to producing everything, so existing `--stages ARTIFACT` invocations keep working unchanged (now idempotent across re-runs).


datashare-index/src/main/java/org/icij/datashare/text/artifact/Artifact.java
```diff
@@ -0,0 +1,17 @@
+package org.icij.datashare.text.artifact;
+
+import java.util.Map;
+
+/** A derived representation of a document, produced alongside it and stored under artifactDir.
+ *  Implementations write payload files only and MUST NOT touch manifest.json. */
+public interface Artifact {
+    /** Stable type name: the --artifacts selector token AND the manifest key. */
+    String type();
```
ClemDoum: This could be an enum

datashare-index/src/main/java/org/icij/datashare/text/artifact/ManifestStore.java
```diff
@@ -0,0 +1,102 @@
...
+/** The only class that reads/writes a document's manifest.json. Writes are concurrency-safe
+ *  (in-JVM ReentrantLock + cross-process FileLock) and atomic (temp + ATOMIC_MOVE). */
+public class ManifestStore {
+    private static final String LOCK_FILE = ArtifactPath.MANIFEST_FILE + ".lock";
+    private static final long LOCK_TIMEOUT_MS = 30_000;
+    private static final ObjectMapper MAPPER = JsonObjectMapper.getMapper();
+    private static final TypeReference<LinkedHashMap<String, ManifestEntry>> MANIFEST_TYPE = new TypeReference<>() {};
+    private static final ConcurrentHashMap<String, ReentrantLock> JVM_LOCKS = new ConcurrentHashMap<>();
+
+    public ManifestEntry get(Path nodeDir, String type) throws IOException {
```
ClemDoum: I'm wondering about the `node`/`nodeDir` naming/vocabulary. It seems to be an addition to our domain vocab and it's not fully explicit. From what I read earlier the `nodeDir` is the document artifact directory. Can't we keep `artifactDir` or `docArtifactDir`. 
I used doc (and not file since at this point we already have the docId)

datashare-index/src/main/java/org/icij/datashare/text/artifact/ArtifactRegistry.java
```diff
@@ -0,0 +1,98 @@
...
+import java.io.IOException;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/** Holds the known artifact types, parses the --artifacts selector, and runs the
+ *  uniform skip-if-current -> produce -> record loop. */
+public class ArtifactRegistry {
```
ClemDoum: I'm wondering about the `Registry` naming here.

Currently in the DS codebase `Registry` is the suffix reserved for object in charge of persistence.

Here I feel like this objects is doing more than just persistence, I feel like we should split the persistence and asset production part and reserver the `Registry` name for the persistence layer

datashare-index/src/main/java/org/icij/datashare/text/artifact/ManifestStore.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.concurrent.locks.ReentrantLock;
+
+import static java.nio.file.StandardCopyOption.ATOMIC_MOVE;
+import static java.nio.file.StandardCopyOption.REPLACE_EXISTING;
+import static java.nio.file.StandardOpenOption.CREATE;
+import static java.nio.file.StandardOpenOption.WRITE;
+
+/** The only class that reads/writes a document's manifest.json. Writes are concurrency-safe
+ *  (in-JVM ReentrantLock + cross-process FileLock) and atomic (temp + ATOMIC_MOVE). */
+public class ManifestStore {
```
ClemDoum: I feel like `ManifestRegistry` would be more consistence with our naming where interfaces/objects in charge of persistence have the `Registry` suffix.

It would also be nice to define its interface and implement this as a `FilesystemManifestRegistry` (this way won't have to change a lot if we decided to persist manifest in a different maner)

datashare-index/src/main/java/org/icij/datashare/text/artifact/ArtifactRegistry.java
```diff
@@ -0,0 +1,98 @@
...
+        }
+    }
+
+    // An artifact is current when a completed entry already exists and was produced with the
+    // exact same task input (config + version) as this run; only then is regeneration skipped.
+    private boolean isCurrent(String type, Artifact artifact, ArtifactContext context) throws IOException {
+        ManifestEntry existing = store.get(context.nodeDir(), type);
+        return existing != null
+                && existing.isComplete()
+                && existing.taskInput().equals(artifact.taskInput());
```
ClemDoum: This is very unreliable, the task input doesn't really contain information about how the current doc was processed.
The current doc is often processed as part of a batch.
The same doc can be processed exactly the same way (with the same config etc etc) as part of 2 different batches.
Here because we directly compare the task input, this will very often evaluate to false unless even if the doc is already processed with exactly the same config

datashare-index/src/main/java/org/icij/datashare/text/artifact/Artifact.java
```diff
@@ -0,0 +1,17 @@
...
+
+    /** This run's config + an explicit version; compared by value for skip-if-current. */
+    Map<String, Object> taskInput();
```
ClemDoum, line 16: Several remarks here:
- task inputs can be very large (even if we try as much as possible to keep them small, but it could be a very very long list of IDs).  This mean we can't naïvely forward the task input. In Python i've defined a `TaskArg` with `as_manifest_entry` people can implement to handle the conversion to something which is not too large (https://github.com/ICIJ/datashare-python/pull/58/changes#diff-840c96a6e23f8013433e1899c15fcf0c2e0f87a4d194dbe8d5fe080899ab8631R275)
- I don't think it's a good idea to use the task inputs as skip processing (see my comment later), this mean we should force the task implementer to implement a function which turns task inputs into a task config hash. Task often take 2 inputs, 1 targeting the data, 1 defining how we should process it. Here we don't care about the data part, we just want to keep the config part, so documents which are already processed with the same config are no reprocessed (we don't care if they were processed as part of the same data set). This means we must find a way to force the task implementer to tell us what makes a processing unique in terms of config (not data). This also why I've added the `as_manifest_entry` method (actually I should have made it abstract to make it even more explicit that devs have to take time do define what should be part of the manifest and what shouldn't)

datashare-index/src/main/java/org/icij/datashare/text/artifact/Artifact.java
```diff
@@ -0,0 +1,17 @@
+package org.icij.datashare.text.artifact;
```
ClemDoum, line 1: I'm wondering if this is the write place for these interface vs the API.

For instance if we decide one day to persist artifacts manifest in the DB, we'll need to import the index into `datashare-db` in order to persist these objects and then the  `datashare-db` in here (defining them in the API would avoid the circular dep).

datashare-index/src/main/java/org/icij/datashare/text/artifact/ArtifactRegistry.java
```diff
@@ -0,0 +1,98 @@
...
+
+    public ArtifactRegistry(List<Artifact> artifacts, ManifestStore store) {
+        this.store = store;
+        for (Artifact artifact : artifacts) {
+            byType.put(artifact.type(), artifact);
+        }
+    }
+
+    /** bare/empty -> all types; CSV -> subset (case-insensitive, trimmed); unknown token -> error. */
+    public Set<String> select(String flagValue) {
```
ClemDoum: This seems to be more of an `datashare-app` responsibility rather than a core manifest level thing. I think the app should decided which artifact to produce and just ask the core to produce them.

datashare-index/src/main/java/org/icij/datashare/text/artifact/ArtifactRegistry.java
```diff
@@ -0,0 +1,98 @@
...
+        }
+        // A value made only of separators (e.g. ",") names no type; fail loudly rather than
+        // silently selecting nothing.
+        if (selected.isEmpty()) {
+            throw new IllegalArgumentException("no artifact types in '" + flagValue + "'; valid types: " + byType.keySet());
+        }
+        return selected;
+    }
+
+    public boolean run(Set<String> selected, ArtifactContext context) {
```
ClemDoum: we could take a set of enum as input for more robustness

datashare-index/src/main/java/org/icij/datashare/text/artifact/ArtifactRegistry.java
```diff
@@ -0,0 +1,98 @@
...
+        boolean allSucceeded = true;
+        for (String type : selected) {
+            if (!produceIfNeeded(type, context)) {
+                allSucceeded = false;
+            }
+        }
+        return allSucceeded;
+    }
```
ClemDoum: I feel like we could use the stream api here:

`return selected.stream().allMatch(t -> produce(t, context))` (and use the `rethrowConsumer`/`rethrowFunction` in case `produce throws Exception`)

datashare-index/src/main/java/org/icij/datashare/text/artifact/ArtifactRegistry.java
ClemDoum: actually I think we should even have a list of artifact (and why not deduplicate them by name) in here.

Take string or enum as input expose you to have and invalid artifact name or to fail resolve the enum into an artifact. This kind of errors should throw earlier (if possible in the app code not in here)

datashare-index/src/main/java/org/icij/datashare/text/artifact/ArtifactRegistry.java
```diff
@@ -0,0 +1,98 @@
...
+
+    // Returns true when the type was produced, skipped, or had nothing to record; false when a
+    // failure was caught and isolated, so the caller can avoid counting a partially-failed document.
+    private boolean produceIfNeeded(String type, ArtifactContext context) {
+        Artifact artifact = byType.get(type);
+        if (artifact == null) {
+            return true;
+        }
```
ClemDoum: To me this resolution logic should lie in the app, it doesn't really belong in here.

datashare-index/src/main/java/org/icij/datashare/text/artifact/ArtifactRegistry.java
```diff
@@ -0,0 +1,98 @@
...
+        try {
+            if (isCurrent(type, artifact, context)) {
+                return true;
+            }
+            // Write the payload first, then stamp the manifest entry complete last, so a crash
+            // mid-produce leaves no "ready" entry and the next run regenerates it.
+            ManifestEntry produced = artifact.produce(context);
+            // A producer may return null to signal there is nothing to record for this node.
```
ClemDoum: I'm wondering about this point. Not writing any manifest entry will mean we'll reprocess the file to get the same null entry once again.
I think we should still write in the manifest and find another way to say that the output was empty.

datashare-index/src/main/java/org/icij/datashare/text/artifact/ArtifactRegistry.java
```diff
@@ -0,0 +1,98 @@
...
+            if (produced != null) {
+                store.put(context.nodeDir(), type, produced.withStatus(ManifestEntry.STATUS_COMPLETE));
+            }
+            return true;
+        } catch (ArtifactException | IOException failure) {
+            LOGGER.error("failed to produce artifact '{}' for document {}", type, context.document().getId(), failure);
+            return false;
```
ClemDoum: I feel like we shouldn't swallow these exception but rather let them bubble up and let the caller decide what they want to do with them

datashare-index/src/main/java/org/icij/datashare/text/artifact/ManifestEntry.java
```diff
@@ -0,0 +1,43 @@
...
+        Integer total,
+        Map<String, Object> pagination,
+        String contentType,
+        String filename,
+        Double confidence,
+        String label) {
+
+    // The single status value that marks an entry as ready to serve. Anything else
+    // (including a missing status) means the artifact is not yet usable.
+    public static final String STATUS_COMPLETE = "complete";
```
ClemDoum: I think we would use an enum here

datashare-index/src/main/java/org/icij/datashare/text/artifact/ManifestEntry.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.Map;
+
+/** One artifact type's entry in a document's manifest.json. Single-file types use
+ *  contentType/filename; paginated types use total/pagination. */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public record ManifestEntry(
+        String status,
+        Map<String, Object> taskInput,
+        Integer total,
+        Map<String, Object> pagination,
```
ClemDoum: These two should be in a nested object

datashare-index/src/main/java/org/icij/datashare/text/artifact/FilesystemManifestRepository.java
```diff
@@ -0,0 +1,102 @@
...
+import static java.nio.file.StandardOpenOption.CREATE;
+import static java.nio.file.StandardOpenOption.WRITE;
+
+/** The only class that reads/writes a document's manifest.json. Writes are concurrency-safe
+ *  (in-JVM ReentrantLock + cross-process FileLock) and atomic (temp + ATOMIC_MOVE). */
+public class ManifestStore {
+    private static final String LOCK_FILE = ArtifactPath.MANIFEST_FILE + ".lock";
+    private static final long LOCK_TIMEOUT_MS = 30_000;
+    private static final ObjectMapper MAPPER = JsonObjectMapper.getMapper();
+    private static final TypeReference<LinkedHashMap<String, ManifestEntry>> MANIFEST_TYPE = new TypeReference<>() {};
```
ClemDoum, line 29: I think we should really have an enum as key to avoid typos

datashare-index/src/main/java/org/icij/datashare/text/artifact/ManifestStore.java
```diff
@@ -0,0 +1,102 @@
...
+            return null;
+        }
+        return read(manifest).get(type);
+    }
+
+    public void put(Path nodeDir, String type, ManifestEntry entry) throws IOException {
+        Files.createDirectories(nodeDir);
+        // Serialise writers within this JVM first: a FileLock is owned per-JVM (not per-thread),
+        // so two threads sharing the channel would otherwise collide on the same lock region.
+        ReentrantLock jvmLock = lockFor(nodeDir);
```
ClemDoum: Locking this way after we've written the artifacts prevents from having half written manifest but doesn't ensure artifacts are consistent with the manifest.

Say we start 2 extractions on the same file with a new config.
Both reach the manifest see that they should go ahead and process.

Process A writes its artifacts, then get stucks on something.
Process B writes its artifacts, then get in there and update the manifest.
Process A is up again and update the manifest.

We end up with a manifest describing process A artifacts while the artifact on the FS are the ones of process B.

I feel like the lock should encompass artifact write + manifest update for consistency.

datashare-index/src/main/java/org/icij/datashare/text/artifact/ArtifactRegistry.java
```diff
@@ -0,0 +1,98 @@
...
+        if (selected.isEmpty()) {
+            throw new IllegalArgumentException("no artifact types in '" + flagValue + "'; valid types: " + byType.keySet());
+        }
+        return selected;
+    }
+
+    public boolean run(Set<String> selected, ArtifactContext context) {
+        boolean allSucceeded = true;
+        for (String type : selected) {
+            if (!produceIfNeeded(type, context)) {
```
ClemDoum: I think we should really be able to force this otherwise it can easily be an operational nightmare.
Say we manage to have more or less reliable config hash from taskInputs (not including the data).

Then we'll systematically have to bump the producer version in a consistent manner in order to reprocess the file (no bump if we want don't want the files to be reprocessed, bump if we want to be reprocess).

This means workers maintainers will have to have 2 versions, their worker lib version (the normal one) + their "doc processing" version (which should only bump when we want to reprocess)

TBH as a worker lib maintainer,  I'd be afraid to forget bumping the "doc processing" version and will prefer bump it everytime so that DS can benefits from bug fixes. I'll probably end up using my lib version as the doc processing version (which will in the end more or less disable caching since we'll reprocess the docs at each of my worker lib bump).


I feel like caching is a diserable feature, but the operational caching strategies are really hard to derive into code. 

Sometimes we'll want to reprocess in case of bug fixes, for other bug fixes the balance between the improvement and the time of reprocessing will not be worth it. All of this operational logic which can't be encoded into versions or more generally task inputs.

II think we must be able to bypass caching and force reprocessing to allows us to for reprocess sets of documents in case of emergency. We don't want to be stuck with no reprocessing because the worker lib dev forget to correctly bump the version of the artifact producer.

datashare-index/src/main/java/org/icij/datashare/text/artifact/RawArtifact.java
```diff
@@ -0,0 +1,43 @@
...
+        Document document = context.document();
+        try {
+            // extract-lib writes the raw/raw.json bytes for the embedded subtree as a side effect.
+            context.sources().extractEmbeddedSources(context.project(), document);
+            // Only embedded nodes have their source bytes cached in their own node dir; a root
+            // document's source is the original on-disk file, served directly and never copied here.
+            // So a root records no raw entry — a raw entry must always mean a payload in this dir.
+            if (document.getExtractionLevel() <= 0) {
+                return null;
+            }
```
ClemDoum, line 36: returning null will trigger reprocessing of the file

ClemDoum (changes requested):

datashare-index/src/main/java/org/icij/datashare/text/artifact/Artifact.java
pirhoo: Agreed on the end state, and I think your last comment gets us there without an enum: the app now resolves the `--artifacts` selector to a concrete `List<Artifact>` (deduped by `type()`), and `ArtifactProducer.run(List<Artifact>, ...)` takes those objects. Unknown-type errors now throw in the app at selector-parse time, not deep in the loop. I guess it's the robustness you're after?

On `type()` itself being an enum: I'm not sure. The tension is extensibility because artifact types are meant to be provided by workers/plugins (extract-lib owns `raw`; the Python side, etc), and a closed enum in `datashare-index` can't name a type a plugin introduces at runtime. That's the same reason `Artifact` is an interface with a `String type()` rather than an enum.

So the question is really: are Java-side artifact types open (plugin-extensible, keep `String`) or closed (central enum, mirror Python's `ArtifactType`)?

datashare-index/src/main/java/org/icij/datashare/text/artifact/ManifestStore.java
pirhoo: Good point, I renamed to `docArtifactDir` to matches `artifacts_dir(doc_id)` on the Python side.

datashare-index/src/main/java/org/icij/datashare/text/artifact/ManifestStore.java
pirhoo: The "Registry = persistence" convention doesn't quite hold in the current codebase: `PipelineRegistry` is a `type -> object` holder (`get(Pipeline.Type)`, registers pipelines from jars), not a persistence class, and `DeliverableRegistry` is similar. So `Registry` here already means "catalog of typed things," and renaming the persistence layer to `...Registry` would conflict with that meaning rather than align with it.

I'd rather keep `Store` for the persistence layer and reserve `Registry` for the catalog.

datashare-index/src/main/java/org/icij/datashare/text/artifact/ManifestEntry.java
pirhoo: Correct!

datashare-index/src/main/java/org/icij/datashare/text/artifact/ManifestStore.java
pirhoo: You're right, the produce + record could run inside a single `store.inLock(...)` critical section.

datashare-index/src/main/java/org/icij/datashare/text/artifact/ArtifactRegistry.java
pirhoo: I will add a `--artifactsForce` flag to bypass skip-if-current, for the operational "reprocess this set now" case you raised.

datashare-index/src/main/java/org/icij/datashare/text/artifact/ArtifactRegistry.java
pirhoo: Done, I split it three ways: `ManifestStore` (persistence), `ArtifactProducer` (production run loop), `ArtifactRegistry` (catalog + selection).

datashare-index/src/main/java/org/icij/datashare/text/artifact/ArtifactRegistry.java
pirhoo: I made a narrower change than a full structural fix: 

The comparison mechanism is unchanged (`existing.isTerminal() && existing.taskInput().equals(artifact.taskInput())`), but the `Artifact.taskInput()` contract is now explicit that it is config-only:

> The config-only fingerprint of this run (e.g. {"type":..,"version":..}) compared by value for skip-if-current. MUST NOT include data (document ids, batch, queries): the same doc processed with the same config in two different batches must compare equal. Keep it compact (mirrors datashare-python TaskArgs.as_manifest_task_input).

Your concrete prediction is true, will very often evaluate to false with the same config, for the Python shape but not the Java code as written. In Python, `TaskArgs.as_manifest_task_input()` auto-dumps the whole args object including `docs`, so data leaks in unless you override to `.pop("docs")`. In Java there is no auto-dump: each `Artifact` hand-writes `taskInput()`, and `RawArtifact` returns a constant `{type, version}` with no ids or batch, so today it does compare equal across batches and the cache hits.

I'm aware this is a documented convention, not an enforced one. Nothing stops a future artifact author from putting `docs` into `taskInput()` and reintroducing exactly the bug. If you want true parity with your Python intent, the next step is to make the contract unavoidable, e.g. rename to `configFingerprint()` or an abstract `asManifestTaskInput()` that is clearly separate from any data-carrying input, so an implementer has to consciously decide config vs data. 

I did not do that yet, but I agree with your initial point. Want me to take that step?

datashare-index/src/main/java/org/icij/datashare/text/artifact/Artifact.java
pirhoo: Same as last comment, because Java hand-writes it (rather than dumping a whole args object), the size problem does not arise for `raw`. But as suggested before, this is just a convention yet. If you want the abstract-method forcing function that your Python `as_manifest_entry` gives, I'm happy to add it, and that same method is where the "strip the large/data part" logic would live.

datashare-index/src/main/java/org/icij/datashare/text/artifact/Artifact.java
pirhoo: Good call, done! The manifest persistence domain (`ManifestEntry`, `ManifestEntryStatus`, `Pagination`) and the `ManifestStore` interface now live in `datashare-api`, which both `datashare-index` and `datashare-db` already depend on.

A future DB-backed manifest impl needs only `datashare-api`, no `index -> db` or `db -> index` cycle. The production contracts that are coupled to `SourceExtractor` (an ES concern) stayed in `datashare-index`.

datashare-index/src/main/java/org/icij/datashare/text/artifact/ArtifactRegistry.java
pirhoo: Done too! So it lives in the app, which owns the catalog and the CLI flags.

datashare-index/src/main/java/org/icij/datashare/text/artifact/ArtifactRegistry.java
pirhoo: Done too, `ArtifactProducer.run` now takes a `List<Artifact>` (concrete objects), not strings or enums, so there is no name-to-artifact resolution left to fail inside the loop.

datashare-index/src/main/java/org/icij/datashare/text/artifact/Artifact.java
ClemDoum: Thanks for the update of the parsing of types.

I don't know for the enum I think plugins make sense for artifacts producer implementations but the don't really make sense for new artifacts types.

To me adding arbitrary artifacts types don't make sense. DS artifact make sense because they are somehow integrated with DS backend and frontend (otherwise we can just run Python script to produce them some assets which are mean to be consumed by other service/apps/scripts etc). Hence, for me artifacts types should always be known.

Known doesn't mean not extensible though. I agree it's important to support extension (say you want to add some functionality which isn't in this repo, then you can provided a new artifact type by registering it in an extension (probably the artifact producer extension), then you can also add a plugin to benefit from your artiract in the frontend).

So I think there are 2 problems here: 
- should we support unknown artifacts types ? -> I'd say know, unknown = unsupported by the backend and frontend
- should we allows artifacts types extensions -> Yes, this means an enum is not the right solution here (I mean it can be for the base types that we provide, but what we need is more something like a registry (in the design pattern sense, not the DS persistence sense). We could use an enum for the base type + java SPI (to allow people to dynamically extend artifact types), but that will work for Java only. If you do a frontend plugin for a full Python producer, then this java backend need to know about it since the code from this PR won't be called.

datashare-index/src/main/java/org/icij/datashare/text/artifact/ArtifactRegistry.java
ClemDoum: Perfect

datashare-index/src/main/java/org/icij/datashare/text/artifact/ArtifactRegistry.java
ClemDoum: Yes what about creating a `ManifestTaskInput` interface which inherits `Map<String, Object>` with a `String artifactFingerPrint()` +  `ManifestTaskInput fromTaskInput(taskInput Map<String, Object>)` method. 

Actually that would probably be simpler to has typed task inputs (inherits from `Map<String, Object>`) so we could to `MyManifestTaskInput MyTaskInput.toManifest()` (to drop not relevant task info), then `String artifactFingerPrint()` would just be a hash of the dict or something.

But introducing typed task input might break task serialization... So maybe we can leave that for a different PR. WDYT ?

datashare-index/src/main/java/org/icij/datashare/text/artifact/ArtifactProducer.java
```diff
@@ -0,0 +1,88 @@
...
+        return byType;
+    }
+
+    // true when produced, skipped, or empty-recorded; false when a failure was caught and isolated.
+    private boolean produce(Artifact artifact, ArtifactContext context, boolean force) {
+        ArtifactType type = artifact.type();
+        try {
+            // The whole critical section (skip check + payload write + manifest stamp) runs under one
+            // lock so a peer can never record a manifest entry describing another peer's payload.
+            return store.inLock(context.docArtifactDir(), () -> {
```
ClemDoum: I wonder if it's not too high in here.

My previous comment about the lock being too low was about avoiding desync/inconsistencies between the manifest and the artifacts actually written on the FS (so it was basically about writing on the FS or on the artifacts + manifest store)

Here the lock is much higher, it's before the artifacts is produced.
It locks: artifacts production + artifacts write + manifest update.

I'm not sure we should lock the artifacts production.

Locking the artifacts production has one advantage it prevent concurrent processing, hence it was launch twice the same processing, then `isCurrent` will return yes and the second processing will not run. However it will wait until the first one is done. So in the end you don't gain time. 

In other cases, it you start 2 processings with different configuration, then you'll have to wait the previous processing is done, before starting processing (to your idle while the other processing runs), so in the end the processing is longer than if you just lock artifacts write + manifest update.

So in the end I think we're systematically slower, for a bit less processing.

Maybe we should just lock the artifact + manifest writes so that they are consistent. 

Then if we want as an optimization, we could register and watch on the manifest, and if it gets updated by another worker with the current config we cancel the current processing and continue (but maybe we can take care of this later)

ClemDoum (changes requested):

datashare-index/src/main/java/org/icij/datashare/text/artifact/ManifestStore.java
ClemDoum: Sorry I meant `Repository` not `Registry`. Today we have `TaskRepository` (not `TaskStore`), `ApiKeyRepository`, `BatchSearchRepository` and so on

mvanzalu (approved):

PR #2233, opened 2026-06-26
feat: send Content-Length header for downloadableResult

lschim:


pirhoo (approved):
I'm approving, pending the CI checks!

However I think it would be nice to have a follow-up PR to also add the content-length header on the /api/:project/documents/src/:id endpoint (using ElasticSearch data).

PR #2240, opened 2026-07-01
feat: Add Content-Length header in endpoint to download original source

lschim:
It will only be available for content types that are not gzip compressed by the rest server (see recent changes by @bamthomas and I in fluent-http for more details), namely all types but :
     
```txt
    "text/html", "text/plain", "text/css",
    "text/javascript", "text/xml",
    "application/json", "application/javascript",
    "application/x-javascript", "application/xml",
    "image/svg+xml"
 ```

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/web/DocumentResource.java
+++ b/datashare-app/src/main/java/org/icij/datashare/web/DocumentResource.java
@@ -17,6 +17,8 @@ import net.codestory.http.Context;
 import net.codestory.http.annotations.*;
 import net.codestory.http.constants.HttpStatus;
 import net.codestory.http.errors.ForbiddenException;
+
+import static net.codestory.http.constants.Headers.CONTENT_LENGTH;
 import static org.icij.datashare.web.errors.ForbiddenException.requireGranted;
 import net.codestory.http.payload.Payload;
 import net.codestory.http.types.ContentTypes;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by pirhoo — APPROVED** [review]
Perfect, thanks!

File: datashare-app/src/main/java/org/icij/datashare/web/DocumentResource.java [hunk 1]
```diff
@@ -513,7 +515,7 @@ private Payload getPayload(Document doc, String index, boolean inline, boolean f
             if (contentType != null && contentType.startsWith("image/ocr-")) {
                 contentType = "image/" + contentType.substring("image/ocr-".length());
             }
-            Payload payload = new Payload(contentType, from);
+            Payload payload = new Payload(contentType, from).withHeader(CONTENT_LENGTH, String.valueOf(doc.getContentLength()));
```
pirhoo: important: on a `filter_metadata=true` download this advertises the wrong length. `SourceExtractor.getSource(..., filterMetadata=true)` returns `metadataCleaner.clean(...).getContent()` (root doc and embedded alike), which strips metadata and yields fewer bytes than `doc.getContentLength()`, while this header still promises the full stored length.

```suggestion
            Payload payload = filterMetadata
                    ? new Payload(contentType, from)
                    : new Payload(contentType, from).withHeader(CONTENT_LENGTH, String.valueOf(doc.getContentLength()));
```

PR #2241, opened 2026-07-01
feat: register subtypes of loaded extensions

mvanzalu:
This PR aims to allow extensions to register custom task result subtypes automatically at startup. Datashare now scans extension JARs for classes annotated with @JsonTypeName that implement Serializable, and registers them with the Jackson object mapper.

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/mode/CommonMode.java
+++ b/datashare-app/src/main/java/org/icij/datashare/mode/CommonMode.java
@@ -1,5 +1,6 @@
 package org.icij.datashare.mode;
 
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import com.fasterxml.jackson.annotation.JsonTypeName;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.jsontype.NamedType;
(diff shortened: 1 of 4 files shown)


datashare-app/src/main/java/org/icij/datashare/mode/CommonMode.java
```diff
@@ -465,6 +468,10 @@ public Routes addExtensionsConfiguration(Routes routes) {
          if (extensionLoader.extensionsDir != null) {
             try {
                 extensionLoader.load((Consumer<Class<?>>) routes::add, this::isEligibleForLoading);
+                extensionLoader.load(
+                        (Consumer<Class<?>>) cls -> JsonObjectMapper.registerSubtypes(new NamedType(cls)),
+                        cls -> cls.isAnnotationPresent(JsonTypeName.class) && Serializable.class.isAssignableFrom(cls)
+                );
```
pirhoo: crucial: looks like this reuses `ExtensionLoader.load`, which returns on the first predicate match and so registers at most one class per jar. Since the predicate deliberately matches many `@JsonTypeName` types, an extension bundling more than one subtype only gets its first-encountered one registered, and deserializing any of the others fails at runtime with `InvalidTypeIdException`. Could we add a find-all variant to `ExtensionLoader` (walk every jar entry, collect all matches instead of returning on the first) and register each one here?

pirhoo (changes requested):
It's a nice addition. The one thing holding it back is that it leans on `ExtensionLoader.findClassesInJar`, a helper built to find a single class per jar, so as written it only registers the first matching subtype and pays for a second full jar scan. If we teach `ExtensionLoader` a find-all path, all three notes below fall out of that one change and the feature does what it sets out to do.

datashare-app/src/main/java/org/icij/datashare/mode/CommonMode.java
mvanzalu: good catch, it really does retrieve just the first subtype. Redoing it with finding all types in the jar.

pirhoo (approved):
Thanks for the changes!

PR #2255, opened 2026-07-08
fix: carry Elasticsearch routing id through document queues

pirhoo:
Document queue entries carried only the document id, not its Elasticsearch routing id (the root document id). Consumers fetching embedded/child documents via `indexer.get(project, id)` routed by id instead of root id, missed the correct shard, and silently skipped those documents. **That means embedded documents never got NLP entities, content-type categories, or artifacts**. This PR introduces a `DocReference` value type that encodes both ids in queue entries and updates producers/consumers to use it.

## Changes

- Add `DocReference` (datashare-api, `org.icij.datashare.text`) encoding `<id>|<rootId>` for embedded docs, or bare `<id>` for root docs — backward compatible with legacy in-flight queue entries.
- Enqueue routing id alongside doc id from both producers: `ElasticsearchSpewer` (real-time SCAN → INDEX → NLP path) and `EnqueueFromIndexTask` (ENQUEUEIDX re-processing path).
- Update consumers (`ExtractNlpTask`, `CategorizeTask`, `ArtifactTask`) to parse `DocReference` entries and fetch documents with the correct routing via new `PipelineTask.getDocument(DocReference)` overloads.
- Add unit/integration test coverage for `DocReference` parsing and for routing-aware fetches across the affected tasks and the spewer.

## AI Disclosure

Done with superpowers, reviewed twice at high effort with Fable.

### DIFF
--- a/datashare-api/src/main/java/org/icij/datashare/text/DocReference.java
+++ b/datashare-api/src/main/java/org/icij/datashare/text/DocReference.java
@@ -2,7 +2,7 @@ package org.icij.datashare.text;
 
 
 /**
- * Reference to an indexed document inside a string {@link org.icij.extract.queue.DocumentQueue}:
+ * Reference to an indexed document inside a string {@code org.icij.extract.queue.DocumentQueue}:
  * the document id, optionally followed by its Elasticsearch routing id (root document id) for
  * embedded documents, separated by '|'. Bare ids (root documents and legacy queue entries)
  * parse to a null rootId, which keeps old in-flight entries working unchanged.
--- a/datashare-api/src/main/java/org/icij/datashare/text/DocReference.java
+++ b/datashare-api/src/main/java/org/icij/datashare/text/DocReference.java
@@ -8,7 +8,7 @@ package org.icij.datashare.text;
  * parse to a null rootId, which keeps old in-flight entries working unchanged.
  */
 public record DocReference(String id, String rootId) {
-    public static final String SEPARATOR = "|";
+    private static final String SEPARATOR = "|";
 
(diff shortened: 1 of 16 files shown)


datashare-api/src/main/java/org/icij/datashare/text/DocReference.java
```diff
@@ -0,0 +1,31 @@
...
+
+
+/**
+ * Reference to an indexed document inside a string {@code org.icij.extract.queue.DocumentQueue}:
+ * the document id, optionally followed by its Elasticsearch routing id (root document id) for
+ * embedded documents, separated by '|'. Bare ids (root documents and legacy queue entries)
+ * parse to a null rootId, which keeps old in-flight entries working unchanged.
+ */
+public record DocReference(String id, String rootId) {
+    public static final String SEPARATOR = "|";
```
ClemDoum: This one could be private since parsing is provided by `DocReference.parse`

datashare-api/src/main/java/org/icij/datashare/text/DocReference.java
```diff
@@ -0,0 +1,31 @@
...
+
+    public static DocReference parse(String entry) {
+        int separatorIndex = entry.indexOf(SEPARATOR);
+        return separatorIndex < 0 ? new DocReference(entry, null) :
```
ClemDoum, line 15: nice retrocompatibility

datashare-api/src/main/java/org/icij/datashare/text/DocReference.java
pirhoo: Done!

PR #2256, opened 2026-07-08
feat: ARTIFACT parallelism

pirhoo:
The ARTIFACT stage drained its document queue on a single thread and ignored `--parallelism`, so a large pre-caching run (e.g. 50k documents) used one core while the rest of the box sat idle. This makes `ArtifactTask` drain the queue across a `--parallelism`-sized worker pool, matching how `IndexTask` and `ExtractNlpTask` already scale.

Behavior is unchanged for the on-demand path: `--parallelism` defaults to 1 when the option is absent (server/on-demand tasks), while CLI batch runs pick up the value passed on the command line.

## Changes

- feat: drain the ARTIFACT queue across a `--parallelism`-sized worker pool
- fix: fail loudly when every worker dies, instead of returning a partial count as success
- fix: propagate cancellation so an interrupted run is recorded as cancelled rather than successful
- refactor: clamp `--parallelism` to a minimum of 1 and hoist `sourceExcludes` to a shared constant

## Notes

- Reuses the existing `--parallelism` option; no new CLI flag.
- Worker threads are daemons so a wedged native OCR call cannot block JVM exit.
- Per-document extraction cost is unchanged (OCR-produced documents are still re-parsed with OCR); the win is running those extractions in parallel.

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/tasks/ArtifactTask.java
+++ b/datashare-app/src/main/java/org/icij/datashare/tasks/ArtifactTask.java
@@ -94,8 +94,8 @@ public class ArtifactTask extends PipelineTask<String> {
                     nbFailures++;
                 }
             }
-            if (!futures.isEmpty() && nbFailures == futures.size()) {
-                throw new IllegalStateException(String.format("all %d artifact worker(s) terminated abnormally", futures.size()), firstCause);
+            if (nbFailures > 0) {
+                throw new IllegalStateException(String.format("%d of %d artifact worker(s) terminated abnormally", nbFailures, futures.size()), firstCause);
             }
         } catch (InterruptedException e) {
             // task was cancelled: force-stop the workers, give them a short grace period to
@@ -158,7 +158,6 @@ public class ArtifactTask extends PipelineTask<String> {
         return runnable -> {
             Thread thread = new Thread(runnable);
             thread.setName(prefix + "-" + counter.incrementAndGet());
-            thread.setDaemon(true);
             return thread;
         };
     }
--- a/datashare-app/src/main/java/org/icij/datashare/tasks/ArtifactTask.java
+++ b/datashare-app/src/main/java/org/icij/datashare/tasks/ArtifactTask.java
@@ -48,6 +48,7 @@ public class ArtifactTask extends PipelineTask<String> {
     private final Path artifactDir;
     private final int pollingInterval;
     private final int parallelism;
+    private final ExecutorService executor;
 
     @Inject
     public ArtifactTask(DocumentCollectionFactory<String> factory, Indexer indexer, PropertiesProvider propertiesProvider, @Assisted Task<Long> taskView, @Assisted final Function<Double, Void> updateCallback) {
@@ -57,6 +58,16 @@ public class ArtifactTask extends PipelineTask<String> {
         pollingInterval = Integer.parseInt(propertiesProvider.get(POLLING_INTERVAL_SECONDS_OPT).orElse(DEFAULT_POLLING_INTERVAL_SEC));
         parallelism = Math.max(1, propertiesProvider.get(PARALLELISM_OPT).map(Integer::parseInt).orElse(1));
         artifactDir = Path.of(propertiesProvider.get(ARTIFACT_DIR_OPT).orElseThrow(() -> new IllegalArgumentException(String.format("cannot create artifact task with empty %s", ARTIFACT_DIR_OPT))));
+        executor = Executors.newFixedThreadPool(parallelism, namedThreadFactory("artifact-worker"));
+    }
+
+    @Override
+    public void cancel(boolean requeue) {
+        // interrupt the task thread first (PipelineTask): that is what makes the blocking
+        // future.get() in call() throw InterruptedException and surface the cancellation.
+        // Then stop the worker pool so the workers themselves wind down promptly.
+        super.cancel(requeue);
+        executor.shutdownNow();
     }
 
     @Override
@@ -65,7 +76,6 @@ public class ArtifactTask extends PipelineTask<String> {
         logger.info("creating artifact cache in {} for project {} from queue {} with {} worker(s) and polling interval {}s", artifactDir, project, inputQueue.getName(), parallelism, pollingInterval);
         AtomicLong nbDocs = new AtomicLong(0);
         AtomicLong nbSkipped = new AtomicLong(0);
-        ExecutorService executor = Executors.newFixedThreadPool(parallelism, namedThreadFactory("artifact-worker"));
         try {
             List<Future<?>> futures = new ArrayList<>();
             for (int i = 0; i < parallelism; i++) {
@@ -92,7 +102,7 @@ public class ArtifactTask extends PipelineTask<String> {
             // actually stop (so we don't read counters/logs while they're still writing), then
             // rethrow so TaskWorkerLoop records this as a cancellation instead of a success.
             executor.shutdownNow();
-            awaitWorkersTermination(executor);
+            awaitWorkersTermination();
             Thread.currentThread().interrupt();
             throw e;
         } finally {
@@ -133,7 +143,7 @@ public class ArtifactTask extends PipelineTask<String> {
         return new SourceExtractor(propertiesProvider);
     }
 
-    private void awaitWorkersTermination(ExecutorService executor) {
+    private void awaitWorkersTermination() {
         try {
             if (!executor.awaitTermination(5, TimeUnit.SECONDS)) {
                 logger.warn("artifact worker(s) did not terminate within the grace period after cancellation");
--- a/datashare-app/src/main/java/org/icij/datashare/tasks/ArtifactTask.java
+++ b/datashare-app/src/main/java/org/icij/datashare/tasks/ArtifactTask.java
@@ -41,6 +41,7 @@ import static org.icij.datashare.cli.DatashareCliOptions.POLLING_INTERVAL_SECOND
 @TemporalSingleActivityWorkflow(name = "artifact", activityOptions = @ActivityOpts(timeout = "P1D"))
 @TaskGroup(TaskGroupType.Java)
 public class ArtifactTask extends PipelineTask<String> {
+    private static final List<String> SOURCE_EXCLUDES = List.of("content", "content_translated");
     private final Logger logger = LoggerFactory.getLogger(getClass());
     private final Indexer indexer;
     private final Project project;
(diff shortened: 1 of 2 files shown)


datashare-app/src/main/java/org/icij/datashare/tasks/ArtifactTask.java
```diff
@@ -19,60 +19,137 @@
...
     private final int pollingInterval;
+    private final int parallelism;
 
     @Inject
     public ArtifactTask(DocumentCollectionFactory<String> factory, Indexer indexer, PropertiesProvider propertiesProvider, @Assisted Task<Long> taskView, @Assisted final Function<Double, Void> updateCallback) {
         super(Stage.ARTIFACT, taskView.getUser(), factory, propertiesProvider, String.class);
         this.indexer = indexer;
         project = Project.project(propertiesProvider.get(DEFAULT_PROJECT_OPT).orElse(DEFAULT_DEFAULT_PROJECT));
         pollingInterval = Integer.parseInt(propertiesProvider.get(POLLING_INTERVAL_SECONDS_OPT).orElse(DEFAULT_POLLING_INTERVAL_SEC));
+        parallelism = Math.max(1, propertiesProvider.get(PARALLELISM_OPT).map(Integer::parseInt).orElse(1));
```
ClemDoum, line 59: This is nice for retrocompatibility, maybe in the next major we can align it with `IndexTask`: `Runtime.getRuntime().availableProcessors()`

datashare-app/src/main/java/org/icij/datashare/tasks/ArtifactTask.java
```diff
@@ -19,60 +19,137 @@
...
+                        firstCause = e.getCause();
+                    }
+                    nbFailures++;
                 }
-                extractor.extractEmbeddedSources(project, doc);
-                nbDocs++;
-            } catch (Throwable e) {
-                logger.error("error in ArtifactTask loop", e);
             }
+            if (!futures.isEmpty() && nbFailures == futures.size()) {
```
ClemDoum: I find this a bit inconsistent, as this will systematically trigger if you use a single thread and the processing fail on one data point, but not if you use 2.

I think the behavior should be consistent whatever the parallelism

datashare-app/src/main/java/org/icij/datashare/tasks/ArtifactTask.java
```diff
@@ -19,60 +19,137 @@
...
-        while ((queueEntry = inputQueue.poll(pollingInterval, TimeUnit.SECONDS)) != null) {
-            try {
-                Document doc = getDocument(indexer, project.name, DocReference.parse(queueEntry), sourceExcludes);
-                if (doc == null) {
-                    nbSkipped++;
-                    continue;
+        logger.info("creating artifact cache in {} for project {} from queue {} with {} worker(s) and polling interval {}s", artifactDir, project, inputQueue.getName(), parallelism, pollingInterval);
+        AtomicLong nbDocs = new AtomicLong(0);
+        AtomicLong nbSkipped = new AtomicLong(0);
+        ExecutorService executor = Executors.newFixedThreadPool(parallelism, namedThreadFactory("artifact-worker"));
```
bamthomas: Usually the executor is set at the instance level. Because it allows a better control on the threads allocated: here at each call to the task we are going to allocate `parallelism`  threads. So if we have 2 calls on the same instance twice as much threads are going to be used. 

Besides,  when we have exited the call() method there is no way to handle the thread pool. So it is harder to gracefully exit the task.

datashare-app/src/main/java/org/icij/datashare/tasks/ArtifactTask.java
```diff
@@ -19,60 +19,137 @@
...
+                throw new IllegalStateException(String.format("all %d artifact worker(s) terminated abnormally", futures.size()), firstCause);
+            }
+        } catch (InterruptedException e) {
+            // task was cancelled: force-stop the workers, give them a short grace period to
+            // actually stop (so we don't read counters/logs while they're still writing), then
+            // rethrow so TaskWorkerLoop records this as a cancellation instead of a success.
+            executor.shutdownNow();
+            awaitWorkersTermination(executor);
+            Thread.currentThread().interrupt();
+            throw e;
```
ClemDoum: I'm not sure to understand what failure mode it tackles exactly ?

datashare-app/src/main/java/org/icij/datashare/tasks/ArtifactTask.java
```diff
@@ -19,60 +19,137 @@
...
+            Thread.currentThread().interrupt();
+        }
+    }
+
+    private static ThreadFactory namedThreadFactory(String prefix) {
+        AtomicInteger counter = new AtomicInteger(0);
+        return runnable -> {
+            Thread thread = new Thread(runnable);
+            thread.setName(prefix + "-" + counter.incrementAndGet());
+            thread.setDaemon(true);
```
ClemDoum: Do we really want a daemon here ? 
It will prevent the app from closing

ClemDoum: I'm not sure we use daemon elsewhere

ClemDoum (changes requested):

datashare-app/src/main/java/org/icij/datashare/tasks/ArtifactTask.java
pirhoo: Hmmm you're right. I Moved the `ExecutorService` to a final instance field created in the constructor (sized by parallelism), like `IndexTask` does with its `DocumentConsumer`, so we no longer allocate a pool per call(). 

Also overrode cancel() to interrupt the task thread and then shutdownNow() the pool, so cancellation stops the workers directly instead of relying only on the thread interrupt propagating.

See https://github.com/ICIJ/datashare/pull/2256/commits/34f45668a1e1796eb8dfb6c18fa278cea5327273

datashare-app/src/main/java/org/icij/datashare/tasks/ArtifactTask.java
pirhoo: Yep, I kept the default at 1 for retro-compat with the current on-demand behavior; aligning it with `IndexTask` (`availableProcessors()`) makes sense as a next-major change.

datashare-app/src/main/java/org/icij/datashare/tasks/ArtifactTask.java
pirhoo: Small clarification on what this counts: a per-document processing failure is caught inside `runWorker` and is non-fatal at any parallelism. `nbFailures` only counts whole-worker deaths, i.e. the `Future` threw (e.g. `createSourceExtractor()` failing or the queue backend erroring). So a single failing data point never triggers this, with 1 thread or with 2.

You're right that the threshold is still asymmetric though: "throw only if all workers died" means 1 worker -> one death fails the task, N workers -> up to N-1 deaths tolerated silently.

I'll make it consistent by failing as soon as any worker dies abnormally (`nbFailures > 0`): per-document errors stay non-fatal, a worker death fails the task regardless of parallelism. That keeps the protection against "all workers die -> task silently reports 0 docs as success" while removing the count dependency.

datashare-app/src/main/java/org/icij/datashare/tasks/ArtifactTask.java
pirhoo: It's for cancellation. `PipelineTask.cancel()` interrupts the task thread while it's blocked in `future.get()`, which throws `InterruptedException`. We `shutdownNow()` the workers, wait a short grace period (so we don't read the counters / log "exiting after N documents" while workers are still writing), then rethrow so `TaskWorkerLoop` records the run as cancelled. 

Without the rethrow, `call()` returns a partial count and the runner marks the task DONE (a cancelled run then looks successful).

datashare-app/src/main/java/org/icij/datashare/tasks/ArtifactTask.java
pirhoo: Yeah you're right, I'll drop it. It's inconsistent with the rest of the codebase and now that the executor is shut down on `cancel()` and at the end of `call()` there's no orphaned-worker case left for daemon status to help with.

datashare-app/src/main/java/org/icij/datashare/tasks/ArtifactTask.java
pirhoo: Done in https://github.com/ICIJ/datashare/pull/2256/changes/76eb97acd200ef307338cce3fab8dcccfdb65d95

datashare-app/src/main/java/org/icij/datashare/tasks/ArtifactTask.java
pirhoo: Done in https://github.com/ICIJ/datashare/pull/2256/changes/76eb97acd200ef307338cce3fab8dcccfdb65d95

datashare-app/src/main/java/org/icij/datashare/tasks/ArtifactTask.java
ClemDoum: I think I still don't get it.
If we call `cancel()` the if already does: `public void cancel(boolean requeue) { ofNullable(taskThread).ifPresent(Thread::interrupt);}`.

Which will thrown an interruptedexception in call(). If we want to shutwown the executor we can either call shutdown in the `cancel` and catch the `call` and rethrow. Or just do nothing on this class's cancel, catch the interruption, cancel executor and rethrow.

I don't grace period and counter/log thing, since we don't have to get to that point if we catch the interruption and exit before the log. If we always rethrow when catching the interrupted exception I don't get how we could reach that log, hence to me there no risk of reading any irrelevant log

pirhoo: Yeah, you're right, that block is overthought. 

While `call()` is running, the only thing that can interrupt the thread is `cancel()`. The task runs synchronously on the loop thread in `handle()`, and the `loopThread.interrupt()` at the end of `mainLoop()` only fires once the loop is already done, never mid-task. So `cancel()` is the one and only path.

Which means the grace period was guarding against nothing: we rethrow before the "exiting after N documents" log, so there's no stale counter or log to worry about. Good catch, I'll drop it.

pirhoo: Done in https://github.com/ICIJ/datashare/pull/2256/changes/6e4230851f4fa0da723ceb2b9c11d261f7df808a

PR #2262, opened 2026-07-13
feat(artifacts): --artifacts generation in INDEX + ARTIFACT stages

### LINKED ISSUE #2225: feat(artifacts): --artifacts flag in INDEX + ARTIFACT stages
Opt-in, type-selectable generation, gated on `artifactDir`. `--artifacts` = all types; `--artifacts=raw,structure` = subset. INDEX default off; ARTIFACT default all.

## Already landed (foundation #2223 / #2224, PR #2231)

- [x] `ARTIFACTS_OPT` + `ARTIFACTS_FORCE_OPT` in `DatashareCliOptions` (JOpt surface) with parsers.
- [x] ARTIFACT stage wired: `ArtifactTask` -> `ArtifactRegistry.select()` (bare/`true` = all, CSV = subset, unknown = error), default all; `--artifactsForce` via `ArtifactProducer`.
- Note: during INDEX, extract-lib's `EmbedSpawner` already writes embedded `raw` bytes to `artifactDir` (existing "embedded caching"), same content-addressed layout as ARTIFACT. No manifest is written during INDEX today.

## Remaining

- [ ] **picocli surface**: add `--artifacts` and `--artifactsForce` to `PipelineOptions`, mirroring the JOpt `withOptionalArg` semantics (`arity = "0..1"`, `fallbackValue = "true"`); map through `toProperties()`.
- [ ] **INDEX inline (raw only), default off**: a manifest-only recorder wired into `ElasticsearchSpewer.writeDocument` (after `indexer.add`). It records the `raw` manifest entry for bytes extract-lib already wrote, **without re-extraction**. Skip-if-current + `force` honored. Parity with ARTIFACT guaranteed by sharing the raw-entry builder (root -> `empty`, embedded -> `singleFile`) between `RawArtifact.produce()` and the recorder.
- [ ] **Guard (fail fast, both stages)**: `--artifacts` present without `--artifactDir` -> `IllegalArgumentException("--artifacts requires --artifactDir")` in both `IndexTask` and `ArtifactTask` (each task checks; no central CLI validation).
- [ ] **Tests**: picocli parsing (bare/CSV/absent + force); guard on both tasks; per-stage defaults (INDEX off writes no manifest, ARTIFACT all); INDEX-with-`--artifacts` == INDEX-without + ARTIFACT stage (same on-disk `raw` bytes + `manifest.json`).

## Decisions / scope

- INDEX inline uses a **manifest-only hook** (not a full `ArtifactProducer` pass): extraction already ran during indexing, so re-extracting would be wasteful.
- Only `raw` is produced inline during INDEX. Structure/page have no bytes written by extract-lib during the parse, so their inline-INDEX production is **deferred** to when those producers land (#2226 / #2227).
- "Drop the PoC's unconditional structure write" is **moot**: nothing on main writes structure today.

Refs: `ARTIFACT_DIR_OPT`, `PipelineOptions`, `Stage`, `IndexTask`, `ArtifactTask`, `ElasticsearchSpewer`, `RawArtifact`.

Part of epic(artifacts) #2229.

pirhoo:
Opt-in per-document artifact generation (`raw`), gated by `--artifacts`, on both CLI surfaces and both the INDEX and ARTIFACT stages. Closes #2225.

- feat: `--artifacts` / `--artifactsForce` on the JOpt and picocli CLI surfaces
- feat: INDEX-stage inline manifest recording via `ManifestRecorder`, wired from `ElasticsearchSpewer`
- feat: fail-fast guard, `--artifacts` requires `--artifactDir`
- fix: no `--artifactsForce` default, so a settings-file value is not clobbered
- refactor: shared `RawArtifact.entryFor`, `ArtifactProducer.entryIsCurrent`, `ArtifactRegistry.withDefaults`, and `ArtifactStages` project/force resolution so INDEX and ARTIFACT stay in parity
- refactor: on-disk layout knowledge centralized in `ArtifactPath`

### DIFF
--- a/datashare-cli/src/main/java/org/icij/datashare/cli/command/PipelineOptions.java
+++ b/datashare-cli/src/main/java/org/icij/datashare/cli/command/PipelineOptions.java
@@ -19,6 +19,17 @@ public class PipelineOptions {
     @Option(names = {"--artifactDir"}, description = "Artifact directory for embedded caching")
     String artifactDir;
 
+    // arity = "0..1" + fallbackValue = "true" mirrors the JOpt withOptionalArg semantics: a bare
+    // --artifacts means "all types", --artifacts=raw,structure selects a subset. Left null when
+    // absent so toProperties omits the key and INDEX produces no artifacts.
+    @Option(names = {"--artifacts"}, arity = "0..1", fallbackValue = "true",
+            description = "Artifact types to produce, comma-separated (bare flag = all types); unknown types are rejected.")
+    String artifacts;
+
+    @Option(names = {"--artifactsForce"}, arity = "0..1", fallbackValue = "true", defaultValue = "false",
+            description = "Reprocess artifacts even when an up-to-date manifest entry exists (bypasses caching).")
+    boolean artifactsForce;
+
     @Option(names = {"--nlpPipeline"}, description = "NLP pipeline to be run", defaultValue = "CORENLP")
     Pipeline.Type nlpPipeline;
 
@@ -95,6 +106,8 @@ public class PipelineOptions {
     public Properties toProperties() {
         Properties props = new Properties();
         DatashareOptions.putIfNotNull(props, ARTIFACT_DIR_OPT, artifactDir);
+        DatashareOptions.putIfNotNull(props, ARTIFACTS_OPT, artifacts);
+        DatashareOptions.put(props, ARTIFACTS_FORCE_OPT, artifactsForce);
         DatashareOptions.putIfNotNull(props, NLP_PIPELINE_OPT, nlpPipeline);
         DatashareOptions.put(props, NLP_PARALLELISM_OPT, nlpParallelism);
         DatashareOptions.put(props, NLP_BATCH_SIZE_OPT, batchSize);
--- a/datashare-cli/src/main/java/org/icij/datashare/cli/command/PipelineOptions.java
+++ b/datashare-cli/src/main/java/org/icij/datashare/cli/command/PipelineOptions.java
@@ -26,7 +26,11 @@ public class PipelineOptions {
             description = "Artifact types to produce, comma-separated (bare flag = all types); unknown types are rejected.")
     String artifacts;
 
-    @Option(names = {"--artifactsForce"}, arity = "0..1", fallbackValue = "true", defaultValue = "false",
+    // Requires an explicit true/false (like --ocr / --followSymlinks) rather than a bare flag, so the
+    // JOpt (withRequiredArg) and picocli surfaces agree: absent -> false, --artifactsForce=false -> false,
+    // --artifactsForce=true -> reprocess. A bare flag is intentionally not allowed to avoid the two
+    // surfaces disagreeing on what "bare" means.
+    @Option(names = {"--artifactsForce"}, arity = "1", defaultValue = "false",
             description = "Reprocess artifacts even when an up-to-date manifest entry exists (bypasses caching).")
     boolean artifactsForce;
 
--- a/datashare-cli/src/main/java/org/icij/datashare/cli/command/PipelineOptions.java
+++ b/datashare-cli/src/main/java/org/icij/datashare/cli/command/PipelineOptions.java
@@ -16,7 +16,9 @@ import static org.icij.datashare.cli.DatashareCliOptions.*;
  */
 public class PipelineOptions {
 
-    @Option(names = {"--artifactDir"}, description = "Artifact directory for embedded caching")
+    // --artifactsDir is an alias: the singular name is public since 21814dfa6 and set in prod
+    // settings files, so it stays the property key while the plural matches --artifacts.
+    @Option(names = {"--artifactDir", "--artifactsDir"}, description = "Artifact directory for embedded caching")
     String artifactDir;
 
     // Mirrors the JOpt withOptionalArg semantics. Left null when absent so toProperties omits the
--- a/datashare-cli/src/main/java/org/icij/datashare/cli/command/PipelineOptions.java
+++ b/datashare-cli/src/main/java/org/icij/datashare/cli/command/PipelineOptions.java
@@ -19,17 +19,14 @@ public class PipelineOptions {
     @Option(names = {"--artifactDir"}, description = "Artifact directory for embedded caching")
     String artifactDir;
 
-    // arity = "0..1" + fallbackValue = "true" mirrors the JOpt withOptionalArg semantics: a bare
-    // --artifacts means "all types", --artifacts=raw,structure selects a subset. Left null when
-    // absent so toProperties omits the key and INDEX produces no artifacts.
+    // Mirrors the JOpt withOptionalArg semantics. Left null when absent so toProperties omits the
+    // key and INDEX produces no artifacts.
     @Option(names = {"--artifacts"}, arity = "0..1", fallbackValue = "true",
             description = "Artifact types to produce, comma-separated (bare flag = all types); unknown types are rejected.")
     String artifacts;
 
-    // Requires an explicit true/false (like --ocr / --followSymlinks) rather than a bare flag, so the
-    // JOpt (withRequiredArg) and picocli surfaces agree: absent -> false, --artifactsForce=false -> false,
-    // --artifactsForce=true -> reprocess. A bare flag is intentionally not allowed to avoid the two
-    // surfaces disagreeing on what "bare" means.
+    // Explicit true/false like --ocr, no bare flag: the JOpt surface uses withRequiredArg, and a bare
+    // flag here would make the two surfaces disagree on what "bare" means.
     @Option(names = {"--artifactsForce"}, arity = "1", defaultValue = "false",
             description = "Reprocess artifacts even when an up-to-date manifest entry exists (bypasses caching).")
     boolean artifactsForce;
--- a/datashare-cli/src/main/java/org/icij/datashare/cli/command/PipelineOptions.java
+++ b/datashare-cli/src/main/java/org/icij/datashare/cli/command/PipelineOptions.java
@@ -26,10 +26,13 @@ public class PipelineOptions {
     String artifacts;
 
     // Explicit true/false like --ocr, no bare flag: the JOpt surface uses withRequiredArg, and a bare
-    // flag here would make the two surfaces disagree on what "bare" means.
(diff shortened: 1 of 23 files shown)


datashare-index/src/main/java/org/icij/datashare/text/artifact/ArtifactProducer.java
```diff
@@ -59,7 +59,7 @@ private boolean produce(Artifact artifact, ArtifactContext context, boolean forc
 
     // Producers return EMPTY as-is; any other (status-less) entry is stamped COMPLETE only after
     // its payload has been written, so a crash mid-produce never leaves a "ready" lie.
-    private ManifestEntry stampTerminal(ManifestEntry produced) {
+    static ManifestEntry stampTerminal(ManifestEntry produced) {
```
caro3801: hint: it's preferable to avoid static function since they are harder to test. This could be moved to the ManifestEntry class maybe?

datashare-index/src/main/java/org/icij/datashare/text/artifact/ArtifactProducer.java
```diff
@@ -68,6 +68,12 @@ private ManifestEntry stampTerminal(ManifestEntry produced) {
     // regeneration skipped.
     private boolean isCurrent(ArtifactType type, Artifact artifact, ArtifactContext context) throws IOException {
         ManifestEntry existing = repository.get(context.docArtifactDir(), type.token());
-        return existing != null && existing.isTerminal() && existing.taskInput().equals(artifact.taskInput());
+        return entryIsCurrent(existing, artifact.taskInput());
+    }
+
+    // Shared with the INDEX-time ManifestRecorder. Compares from the always-non-null taskInput side so
+    // an entry with an absent taskInput does not NPE.
+    static boolean entryIsCurrent(ManifestEntry existing, Map<String, Object> taskInput) {
```
caro3801: hint: (same as above) it's preferable to avoid static function since they are harder to test. This could be moved to the ManifestEntry class maybe?

datashare-cli/src/main/java/org/icij/datashare/cli/command/PipelineOptions.java
```diff
@@ -19,6 +19,18 @@ public class PipelineOptions {
     @Option(names = {"--artifactDir"}, description = "Artifact directory for embedded caching")
     String artifactDir;
 
+    // Mirrors the JOpt withOptionalArg semantics. Left null when absent so toProperties omits the
+    // key and INDEX produces no artifacts.
+    @Option(names = {"--artifacts"}, arity = "0..1", fallbackValue = "true",
```
caro3801, line 26: suggestion(consistency) : the option is now plural "artifacts", whereas we used singular for "artifactDir". Can we rename one or the other so it's easier to remember/document ?

pirhoo: Plural because it takes a list of types: `--artifacts raw,structure` (bare flag = all), same shape as `--stages`. The odd one out is `--artifactDir`, but that one has been public since 21814dfa6 (Aug 2024) and is set in prod settings files, so renaming it would break existing configs. I would rather keep both names and document them together; happy to add an `--artifactsDir` alias in a follow-up if you think it helps.

datashare-index/src/main/java/org/icij/datashare/text/artifact/ArtifactProducer.java
pirhoo: Done in c0bd0119f: it is now `ManifestEntry.withTerminalStatus()`, so `ArtifactProducer` just calls `produced.withTerminalStatus()`. Covered directly by `ManifestEntryTest`.

datashare-index/src/main/java/org/icij/datashare/text/artifact/ArtifactProducer.java
pirhoo: Same, c0bd0119f: now `ManifestEntry.isCurrentFor(taskInput)`. The null check stays at the two call sites (the repository read can return null), and the comparison is still done from the argument side so an entry without taskInput does not NPE. Four cases tested in `ManifestEntryTest`.

pirhoo:
Thanks @caro3801, useful pass. Fixed here:

- **`JVM_LOCKS` growth** (a450e64b): weak values now, so a lock lives only while a thread holds it. Kept one lock per dir rather than striping, because the reentrancy shortcut in `inLock` skips the cross-process `FileLock` and two dirs sharing a stripe would make that unsound.
- **`manifestRecorder` volatile** (129c7ec2).
- **`Optional.get()` coupling** (2f68cf21): now `.orElse(null)`, same as `ArtifactTask`.
- **`mergeWith` doc nit** (555e23b4): comment names `CommonMode.overrideWith` instead.

Not changed:

- **Settings-file parity**: real, but pre-existing and ARTIFACT is the outlier rather than INDEX, since every other stage reads its task args. Filed as #2298.
- **`catch (Exception)`**: narrowing to `IOException | RuntimeException` catches the same set in practice, `Error` already propagates.

Unit tests plus the artifact integration tests are green.

datashare-cli/src/main/java/org/icij/datashare/cli/command/PipelineOptions.java
caro3801: I agree with the retrocompat argument. Adding `--artifactsDir` makes sense to me

datashare-cli/src/main/java/org/icij/datashare/cli/command/PipelineOptions.java
```diff
@@ -19,6 +19,18 @@ public class PipelineOptions {
...
+            description = "Artifact types to produce, comma-separated (bare flag = all types); unknown types are rejected.")
+    String artifacts;
+
+    // Explicit true/false like --ocr, no bare flag: the JOpt surface uses withRequiredArg, and a bare
+    // flag here would make the two surfaces disagree on what "bare" means.
+    @Option(names = {"--artifactsForce"}, arity = "1", defaultValue = "false",
```
caro3801: important: This reintroduces the bug just fixed on the JOpt CLI surface: `artifactsForce` has `defaultValue = "false"`, so it's always `false` unless the user passes `--artifactsForce`. `toProperties()` writes it unconditionally (`DatashareOptions.put`, not `putIfNotNull`), so `artifactsForce=false` always ends up in the properties — even with no flag at all. Since `CommonMode` merges CLI properties over the settings file via `putAll`, this silently overrides a settings-file `artifactsForce=true` back to `false`.

Suggest the same fix as the JOpt side: drop the default, make `artifactsForce` a nullable `Boolean`, and use `putIfNotNull`.

datashare-cli/src/main/java/org/icij/datashare/cli/command/PipelineOptions.java
```diff
@@ -95,6 +107,8 @@ public class PipelineOptions {
     public Properties toProperties() {
         Properties props = new Properties();
         DatashareOptions.putIfNotNull(props, ARTIFACT_DIR_OPT, artifactDir);
+        DatashareOptions.putIfNotNull(props, ARTIFACTS_OPT, artifacts);
+        DatashareOptions.put(props, ARTIFACTS_FORCE_OPT, artifactsForce);
```
caro3801: linked to https://github.com/ICIJ/datashare/pull/2262/changes#r3690870263

datashare-index/src/main/java/org/icij/datashare/text/artifact/ManifestRecorder.java
```diff
@@ -0,0 +1,55 @@
...
+
+    /** Record the raw entry for a document written during indexing. No-op when raw was not among the
+     *  selected types, or when a matching terminal entry already exists (unless force). */
+    public void record(Document document) throws IOException {
+        if (!rawSelected) {
+            return;
+        }
+        Path docArtifactDir = ArtifactPath.dir(projectRoot, document.getId());
+        if (!force) {
+            ManifestEntry existing = repository.get(docArtifactDir, ArtifactType.RAW.token());
```
caro3801, line 41: Minor: `record()` does a `repository.get(...)` check and a later `repository.put(...)` without holding the JVM lock across both, so there's a TOCTOU (time-of-check to time-of-use aka race condition) gap between the "is current" check and the write. Not a new issue - it mirrors the existing `ArtifactProducer.isCurrent`/`produce` pattern - just flagging for awareness, not blocking.
suggested fix: move the read-check-write sequence inside a single `inLock` call instead of two separate `repository.get()` / `repository.put() `calls.

caro3801 (changes requested):

datashare-cli/src/main/java/org/icij/datashare/cli/command/PipelineOptions.java
pirhoo: Right, only the JOpt surface was fixed. Done in 4fb1b4b53: `Boolean artifactsForce`, no `defaultValue`, `putIfNotNull`, so no flag means no key (absent -> `false` in `ArtifactStages.force`). `StageRunCommandTest` now asserts the key is absent instead of `false`.

datashare-cli/src/main/java/org/icij/datashare/cli/command/PipelineOptions.java
pirhoo: Added in 5c3fc3403: both surfaces now accept `--artifactsDir` as an alias, and `artifactDir` stays the property key (JOpt derives it from the first long flag), so existing settings files keep working.

datashare-index/src/main/java/org/icij/datashare/text/artifact/ManifestRecorder.java
pirhoo: Deliberate, same shape as `ArtifactProducer`: the entry is a pure function of the document and the type version (no timestamps), and `put` merges under the lock, so a racing writer writes byte-identical content. Wrapping the pre-check in `inLock` would add a `FileLock` acquire per indexed document on the INDEX hot path for no correctness gain, so I would keep the unlocked read.

caro3801 (approved):
thanks for all the changes and explanations! it's a great addition to the project
We should consider updating the doc when it'll be marked as stable

pirhoo:
Thanks for the great review!

PR #2274, opened 2026-07-16
fix(batch-search): JSON-escape query when substituting into query template

pirhoo:
Batch search substituted the raw query into a JSON template (`"query":"<query>"`) without escaping, so a query containing a double quote (e.g. `title:"source.zip"`) broke the JSON and failed with a `JsonParsingException`. Interactive search was unaffected. This JSON-escapes the query before substitution.

- fix: JSON-escape the query and use literal `replace()` instead of regex `replaceAll()` when filling the query template
- test: cover template queries containing double quotes (execute + scroll + batch paths)


caro3801 (approved):
good catch! receiving an ElasticsearchException instead of breaking the JSON parser is a good improvement

PR #2277, opened 2026-07-17
fix: cli mode awaits for its own spawned tasks

mvanzalu:
This PR aims to fix cli mode hanging after finishing tasks. Instead of waiting for the 'RUNNING' tasks of the repository to finish, it now waits for it's own created tasks.

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/CliApp.java
+++ b/datashare-app/src/main/java/org/icij/datashare/CliApp.java
@@ -41,13 +41,13 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.time.Instant;
+import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.function.Supplier;
 
-import static java.util.concurrent.TimeUnit.SECONDS;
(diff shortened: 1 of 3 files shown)


datashare-app/src/main/java/org/icij/datashare/CliApp.java
```diff
@@ -174,45 +174,65 @@ private static void runTaskWorker(CommonMode mode, Properties properties) throws
...
+        awaitTasks(taskManager, taskIds);
         taskManager.shutdown();
     }
 
+    static void awaitTasks(TaskManager taskManager, List<String> taskIds) throws IOException {
+        int pollingInterval = taskManager.getTerminationPollingInterval();
+        while (true) {
+            boolean allDone = taskIds.stream().allMatch(taskId -> {
+                try {
+                    return taskManager.getTask(taskId).getState().isFinal();
```
ClemDoum: I would avoid doing `getTask` + `getState` on the task since `getTask` deserialized potentially large results.
I would avoid doing a `allMatch` to avoid polling again task we know are already finished.
Probably updating `waitTasksToBeDone` is a better option it uses `getTaskIds` (no input/results deserialization) + filter for non final states. We just have to check that no task ID of unfinished task matches our task IDs

ClemDoum (changes requested):

PR #2289, opened 2026-07-28
refactor(pipeline): remove the POISON pill mechanism

pirhoo:
Removes the in-band POISON sentinel from the pipeline queues. Consumers now drain until their input queue is empty and their producer's task is terminal, read from `TaskRepository`, which works for any number of consumers, in-process or across processes, and leaves nothing in the queue between runs. `STRING_POISON` is gone; a transitional skip guard for queues written by pre-upgrade runs stays until 21.18.

- refactor(pipeline): drain every stage queue until empty instead of terminating on a sentinel
- refactor(artifact): delete the poison relay, the residual-poison startup drain and the null-poll retry budget
- feat(tasks): gate queue drains on the upstream producer task state, so an empty queue only means "done" when the producer is finished
- feat(pipeline): chain stage task ids in `CliApp.runPipeline` and the `/batchUpdate/index` and `/findNames` endpoints
- feat(index): gate `IndexTask` through `DocumentQueueDrainer`'s existing `SealableLatch`, no extract-lib change
- fix(tasks): make cancellation stop a drain instead of silently draining to the end and reporting DONE
- fix(tasks): survive interrupts that Redisson and the Elasticsearch rest-client rethrow wrapped in a RuntimeException
- fix(tasks): clear the interrupt flag when signalling cancellation, so it does not leak onto the next task on an AMQP worker
- fix(artifact): restore the `searcher.clearScroll()` dropped with the sentinel block
- fix(index): stop counting a skipped legacy sentinel as an indexed document
- fix(cli): exit non-zero when a stage fails, run each configured stage once, reject `NLP` + `CREATENLPBATCHESFROMIDX`
- refactor: delete `PipelineHelper.has()`, the unreachable null-task-id branch and dead `pollingInterval` test properties
- docs(cli): describe what `pollingInterval` actually does
- test: cover concurrent producer/consumer stages, cancellation, the legacy sentinel and the stage barrier

Notes for review: the sentinel used to inflate progress denominators, so two progress assertions now pin `contains(1.0)`; `CREATENLPBATCHESFROMIDX` now runs after `NLP` in enum order, which `PipelineHelper` rejects as a pairing rather than documenting.

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/CliApp.java
+++ b/datashare-app/src/main/java/org/icij/datashare/CliApp.java
@@ -196,6 +196,9 @@ class CliApp {
      * Consumers stop as soon as their input queue is empty (there is no end-of-stream sentinel), so
      * this barrier is what makes "empty" mean "the producer is done" for any number of workers.
      * PipelineHelper.stages is already sorted by Stage.comparator, which is pipeline order.
+     * That enum order runs CREATENLPBATCHESFROMIDX after NLP, so configuring both stages together
+     * is not a supported pairing: CreateNlpBatchesFromIndex searches .without(nlpPipeline), and by
+     * the time it runs, NLP has already marked every document it would otherwise have selected.
      * ponytail: launcher-side barrier, so a producer and its consumer stage no longer overlap. If
      * that overlap ever measures, gate the consumer on the upstream task state (pass the producer's
      * task id in the consumer's args) rather than reintroducing a sentinel.
@@ -208,10 +211,6 @@ class CliApp {
                 continue;
             }
             String taskId = taskManager.startTask(taskClass, nullUser(), propertiesToMap(properties));
-            if (taskId == null) {
-                logger.error("stage {} could not be started, stopping the pipeline", stage);
-                return;
-            }
             taskManager.awaitTermination(Integer.MAX_VALUE, SECONDS, Set.of(taskId));
             Task.State state = taskManager.getTask(taskId).getState();
             if (state != Task.State.DONE) {
--- a/datashare-app/src/main/java/org/icij/datashare/CliApp.java
+++ b/datashare-app/src/main/java/org/icij/datashare/CliApp.java
@@ -27,7 +27,7 @@ import org.icij.datashare.tasks.DeduplicateTask;
 import org.icij.datashare.tasks.EnqueueFromIndexTask;
 import org.icij.datashare.tasks.ExtractNlpTask;
 import org.icij.datashare.tasks.IndexTask;
-import org.icij.datashare.tasks.PipelineTask;
+import org.icij.datashare.tasks.UpstreamGate;
 import org.icij.datashare.tasks.ScanIndexTask;
 import org.icij.datashare.tasks.ScanTask;
 import org.icij.datashare.text.indexing.Indexer;
@@ -200,7 +200,7 @@ class CliApp {
 
     /**
      * Starts every configured stage, then awaits them all at once. Each stage carries the previous
-     * stage's task id in its args under {@link PipelineTask#UPSTREAM_TASK_ID}, so a consumer knows
+     * stage's task id in its args under {@link UpstreamGate#UPSTREAM_TASK_ID}, so a consumer knows
      * which producer feeds its input queue and keeps polling while that producer runs. Stages
      * therefore overlap, which a bounded in-memory queue needs to make progress on a large corpus:
      * what makes termination correct is that gate, not the launch sequencing.
@@ -220,7 +220,7 @@ class CliApp {
             }
             Map<String, Object> args = propertiesToMap(properties);
             if (upstreamTaskId != null) {
-                args.put(PipelineTask.UPSTREAM_TASK_ID, upstreamTaskId);
+                args.put(UpstreamGate.UPSTREAM_TASK_ID, upstreamTaskId);
             }
             upstreamTaskId = taskManager.startTask(taskClass, nullUser(), args);
             stagesByTaskId.put(upstreamTaskId, stage);
--- a/datashare-app/src/main/java/org/icij/datashare/CliApp.java
+++ b/datashare-app/src/main/java/org/icij/datashare/CliApp.java
@@ -43,6 +43,7 @@ import java.io.IOException;
 import java.nio.file.Path;
 import java.time.Instant;
 import java.util.Date;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -176,8 +177,13 @@ class CliApp {
 
         PipelineHelper pipeline = new PipelineHelper(new PropertiesProvider(properties));
         logger.info("executing {}", pipeline);
-        runPipeline(taskManager, pipeline, properties);
+        boolean completed = runPipeline(taskManager, pipeline, properties);
         taskManager.shutdown();
+        if (!completed) {
+            // a partial run must not look like a success: `datashare ... && post-process.sh` would
+            // otherwise post-process a corpus that was never fully indexed
+            System.exit(EXIT_RUNTIME);
+        }
     }
 
     static final Map<Stage, Class<?>> TASK_CLASSES = Map.of(
@@ -195,16 +201,16 @@ class CliApp {
      * Runs the configured stages one at a time, awaiting each stage's task before starting the next.
      * Consumers stop as soon as their input queue is empty (there is no end-of-stream sentinel), so
      * this barrier is what makes "empty" mean "the producer is done" for any number of workers.
-     * PipelineHelper.stages is already sorted by Stage.comparator, which is pipeline order.
-     * That enum order runs CREATENLPBATCHESFROMIDX after NLP, so configuring both stages together
-     * is not a supported pairing: CreateNlpBatchesFromIndex searches .without(nlpPipeline), and by
-     * the time it runs, NLP has already marked every document it would otherwise have selected.
+     * PipelineHelper.stages is already sorted by Stage.comparator, which is pipeline order, and is
+     * de-duplicated here so a repeated --stages entry runs its stage once.
      * ponytail: launcher-side barrier, so a producer and its consumer stage no longer overlap. If
      * that overlap ever measures, gate the consumer on the upstream task state (pass the producer's
      * task id in the consumer's args) rather than reintroducing a sentinel.
+     *
+     * @return true when every configured stage completed, false when the pipeline aborted
      */
-    static void runPipeline(TaskManager taskManager, PipelineHelper pipeline, Properties properties) throws Exception {
-        for (Stage stage : pipeline.stages) {
+    static boolean runPipeline(TaskManager taskManager, PipelineHelper pipeline, Properties properties) throws Exception {
+        for (Stage stage : new LinkedHashSet<>(pipeline.stages)) {
             Class<?> taskClass = TASK_CLASSES.get(stage);
             if (taskClass == null) {
                 logger.warn("no task class for stage {}, skipping it", stage);
@@ -215,9 +221,10 @@ class CliApp {
             Task.State state = taskManager.getTask(taskId).getState();
             if (state != Task.State.DONE) {
                 logger.error("stage {} ended in state {}, stopping the pipeline: a downstream stage would read its partial queue as complete", stage, state);
-                return;
+                return false;
             }
         }
+        return true;
     }
 
     static int handleUserCreate(UserAdminService service, Properties properties) {
--- a/datashare-app/src/main/java/org/icij/datashare/CliApp.java
+++ b/datashare-app/src/main/java/org/icij/datashare/CliApp.java
@@ -27,6 +27,7 @@ import org.icij.datashare.tasks.DeduplicateTask;
 import org.icij.datashare.tasks.EnqueueFromIndexTask;
 import org.icij.datashare.tasks.ExtractNlpTask;
 import org.icij.datashare.tasks.IndexTask;
+import org.icij.datashare.tasks.PipelineTask;
 import org.icij.datashare.tasks.ScanIndexTask;
 import org.icij.datashare.tasks.ScanTask;
 import org.icij.datashare.text.indexing.Indexer;
@@ -43,10 +44,10 @@ import java.io.IOException;
 import java.nio.file.Path;
 import java.time.Instant;
 import java.util.Date;
+import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
 import java.util.Properties;
 import java.util.function.Supplier;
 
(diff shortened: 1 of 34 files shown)


datashare-app/src/main/java/org/icij/datashare/tasks/EnqueueFromIndexTask.java
```diff
@@ -90,15 +90,6 @@ public Long call() throws Exception {
                 docsToProcess.forEach(doc -> outputQueue.add(DocReference.fromDocument((Document) doc).toQueueEntry()));
                 docsToProcess = searcher.scroll(scrollDuration).toList();
             } while (!docsToProcess.isEmpty());
-            searcher.clearScroll();
```
lschim, line 93: Question : Why do you remove the call to clearScroll() ?

EDIT : Remark : I see it was revert in the very next commit. I think this kind of very small mistakes reverted in a following commit could be squashed, it preserves the history and ease code reviews

datashare-app/src/main/java/org/icij/datashare/tasks/PipelineTask.java
```diff
@@ -73,6 +73,34 @@ protected Document getDocument(Indexer indexer, String projectName, DocReference
...
+    }
+
+    // ponytail: transitional. Redis queue keys survive upgrades, so a pre-21.16 run can leave a
+    // "POISON" entry in a String queue. Skip it instead of resolving it as a doc reference.
+    // Delete in 21.18, together with IndexTask's PATH_POISON guard.
+    protected boolean isLegacySentinel(String queueEntry) {
+        if (!"POISON".equals(queueEntry)) {
+            return false;
+        }
+        LoggerFactory.getLogger(getClass()).warn("skipping legacy POISON sentinel in queue {}", inputQueue.getName());
```
lschim: Remark : It would probably be better to declare the Logger at the top level to make it available to the full class if needed. Also, the log line shouldn't be here, it should be at the place where the skip is actually done

datashare-app/src/main/java/org/icij/datashare/web/TaskResource.java
```diff
@@ -535,10 +538,15 @@ public Payload extractNlp(@Parameter(name = "pipeline", description = "name of t
         properties.put(NLP_PIPELINE_OPT, pipelineName);
         syncModels(parseBoolean(properties.getProperty(SYNC_MODELS_OPTION, "true")));
         List<String> tasks = new LinkedList<>();
+        Map<String, Object> nlpArgs = propertiesToMap(properties);
```
caro3801, line 541: important(security): `upstreamTaskId` here comes straight from client-supplied options with no ownership check. A user could point this at any task ID they control and keep it running, causing this task to poll forever and tie up a worker - repeatable to exhaust the whole worker pool (DoS). 
Suggest validating that the referenced task belongs to the caller, or stripping this key from client input and only setting it server-side.

datashare-app/src/main/java/org/icij/datashare/tasks/PipelineTask.java
```diff
@@ -77,6 +89,68 @@ protected Document getDocument(Indexer indexer, String projectName, DocReference
...
+    /**
+     * True while the producer feeding this stage may still enqueue. An absent upstream id, an
+     * unknown task, or a repository failure all mean "no upstream to wait for", so the consumer
+     * falls back to exiting on an empty queue, which is the pre-gate behaviour.
+     */
+    protected boolean upstreamRunning(TaskRepository taskRepository) {
+        return upstreamTaskId().map(upstreamTaskId -> {
+            try {
+                return !taskRepository.getTask(upstreamTaskId).getState().isFinal();
+            } catch (UnknownTask | IOException e) {
```
caro3801: remark: `upstreamRunning()` only catches `UnknownTask`/`IOException`. The Jooq repo can throw unchecked `DataAccessException` on DB errors, which isn't caught here - could silently mark the producer "finished" too early (data loss) or crash the consumer on an exception that could be tolerated.

datashare-app/src/main/java/org/icij/datashare/tasks/ExtractNlpTask.java
```diff
@@ -74,25 +71,47 @@ public ExtractNlpTask(Indexer indexer, PipelineRegistry registry, final Document
...
-                    nbMaxPolls--;
+                queueEntry = inputQueue.poll();
+            } catch (RuntimeException e) {
+                if (causedByInterrupt(e)) {
+                    Thread.currentThread().interrupt();
+                    break;
                 }
+                throw e;
+            }
+            if (queueEntry == null) {
```
caro3801, line 83: remark: removing the old `NB_MAX_POLLS` retry means a task started without upstream wiring (e.g. via `PUT /api/task/:id`) now exits as DONE on the very first empty poll instead of getting a grace period - risk of false "0 documents processed" results.

datashare-app/src/main/java/org/icij/datashare/CliApp.java
```diff
@@ -176,44 +178,63 @@ private static void runTaskWorker(CommonMode mode, Properties properties) throws
...
+        Map<String, Stage> stagesByTaskId = new LinkedHashMap<>();
+        String upstreamTaskId = null;
+        for (Stage stage : new LinkedHashSet<>(pipeline.stages)) {
+            Class<?> taskClass = TASK_CLASSES.get(stage);
+            if (taskClass == null) {
+                logger.warn("no task class for stage {}, skipping it", stage);
+                continue;
+            }
+            Map<String, Object> args = propertiesToMap(properties);
+            if (upstreamTaskId != null) {
```
caro3801, line 222: hint: each stage is gated on the *previous started* stage's id, not its actual producer. For non-adjacent `--stages` subsets (e.g. skipping a stage) this wires the wrong upstream - fails safe but can needlessly serialize stages that should run concurrently.

datashare-app/src/main/java/org/icij/datashare/tasks/DeduplicateTask.java
```diff
@@ -50,13 +53,34 @@ long transferToOutputQueue() throws Exception {
...
+                    throw e;
+                }
+                if (path == null) {
+                    if (drained(taskRepository)) {
+                        break;
+                    }
+                    Thread.sleep(UPSTREAM_POLL_INTERVAL_MS);
+                    continue;
+                }
                 if (filter.test(path)) {
```
caro3801, line 71: suggestion(for consistency): unlike the other tasks, this task has no legacy `POISON` sentinel skip ( `isLegacySentinel`) - a stale sentinel from a pre-21.16 upgrade would pass through here and affect dedup counts before being caught downstream.

datashare-app/src/main/java/org/icij/datashare/tasks/DeduplicateTask.java
caro3801, line 40: important: 
Before this PR, SCAN presumably finished before `DEDUPLICATE` started, so a single `removeDuplicates()` at the top was sufficient.
Since this PR's concurrency change (commit ac81d3128) specifically lets `DEDUPLICATE` start draining while `SCAN` is still running, any duplicate `SCAN` enqueues after that single `removeDuplicates()` call sails straight through with zero deduplication.

`DeduplicateTask.call()` runs `removeDuplicates()` exactly once, then drains via `poll()` in a loop, forwarding everything to the output queue unconditionally (p -> true, no dedup check per-item). 

suggested fix: either re-run `removeDuplicates()` periodically during the drain loop, or track seen paths incrementally (e.g. a set checked in the `filter.test(path)` call) instead of relying on one upfront pass.

mvanzalu (approved):

caro3801 (changes requested):
thanks this is a big and solid refactor; replacing the poison-pill sentinel with an upstream-task-id gating mechanism is a cleaner way to let pipeline stages overlap.

With the assistance of Claude code here are the findings:  

### Security
- **`TaskResource.java:541`** — `upstreamTaskId` is taken from client-supplied options with no ownership/authorization check, and forwarded unfiltered into task args (also reachable via the generic `PUT /api/task/:id`). An authenticated user can point it at a task they control and keep alive, causing the started task to poll forever and pin a worker — repeatable to exhaust the whole fixed-size worker pool. This needs validation before merge.

### Correctness
- **`PipelineTask.java:126`** - `upstreamRunning()` only catches checked `UnknownTask`/`IOException`; the real `JooqTaskRepository` can throw unchecked `DataAccessException` on transient DB errors, which isn't handled and can either mask a still-running producer as finished (silent data loss) or crash the consumer outright.
- **`DeduplicateTask.java:43`** - confirmed real bug: `removeDuplicates()` is a one-shot scan and the queue is a plain Redis list with no insert-time uniqueness. Now that this PR lets DEDUPLICATE overlap with SCAN, any duplicate SCAN enqueues after that single call skips dedup entirely and gets indexed twice.
- **`ExtractNlpTask.java:86`** - the old `NB_MAX_POLLS` retry/grace-period (removed in `2fcdfda60`) was never restored for tasks with no upstream wired (e.g. started via the generic `PUT /api/task/:id`, or standalone). These now exit as DONE on the very first empty poll instead of getting any grace period — see inline comment for a suggested fix.
- **`CliApp.java:222`** - `runPipeline()` gates each stage on the *previous started* stage's id rather than its actual producer, so a `--stages` subset where order skips a stage wires the wrong upstream. Fails safe (just over-serializes) but defeats the overlap this PR is meant to enable.
- **`DeduplicateTask.java:74`** - minor/nice-to-have: unlike every sibling stage in this PR, `DeduplicateTask` has no legacy `"POISON"` sentinel skip. `IndexTask` catches it downstream so it's not a functional bug, but worth adding for consistency and cleaner dedup-count logging.

### Suggested priority
The security issue and the `DeduplicateTask` dedup gap are the two blocking for merging; the others are worth fixing but lower risk.

datashare-app/src/main/java/org/icij/datashare/web/TaskResource.java
pirhoo: Fixed. `UPSTREAM_TASK_ID` is now stripped in `applyProjectTo`, which covers `/findNames`, `/batchUpdate/index` and `/batchUpdate/index/:path:` in one place; both launchers still set it server-side afterwards. Covered by two new `TaskResourceTest` cases.

Note the three endpoints are all `checkAllowedMode(LOCAL, EMBEDDED)`, so the worker-pool story really lands via `PUT /api/task/:id`, which is a separate pre-existing hole: it carries no `@TaskPolicy` and already accepts an arbitrary task class with arbitrary args. An ownership check there would not close the DoS either, since you can point at a long-running task you own. Worth its own issue rather than a partial guard here.

datashare-app/src/main/java/org/icij/datashare/tasks/PipelineTask.java
pirhoo: Fixed, and confirmed: jOOQ throws `DataAccessException` unchecked, so it escaped the lambda entirely.

`UnknownTask` now means finished, while `IOException`/`RuntimeException` mean "still running": retrying costs one poll interval, whereas calling it finished can end a drain on a transient blip and strand whatever the producer enqueues next. Covered by a new `UpstreamGateTest` case that fails on the old code.

Side effect: a wrapped-interrupt `RuntimeException` from `getTask` used to escape `call()` as-is and report ERROR. It now falls through to `Thread.sleep`, which throws `InterruptedException`, so cancellation during an upstream read reports CANCELLED.

datashare-app/src/main/java/org/icij/datashare/tasks/ExtractNlpTask.java
pirhoo: Pushing back on this one. The old budget was `NB_MAX_POLLS = 3` x `pollingInterval` (CLI default 60s), after which the task exited DONE having returned 0 anyway, so the false "0 documents processed" already existed before this PR: it just arrived three minutes later.

A consumer with no upstream wired, polling an empty queue, genuinely has nothing to wait for. Restoring the budget re-adds the timing heuristic that the producer-state gate is meant to replace, and it never actually distinguished "producer is slow" from "producer is done".

datashare-app/src/main/java/org/icij/datashare/CliApp.java
pirhoo: I do not think this one holds. `PipelineHelper.getOutputQueueNameFor(stage)` returns the queue named after the next stage **in the configured list**, not in a fixed graph (`PipelineHelper.java:37` and `getNextStage` at `:43`).

So for `--stages SCAN,NLP`, SCAN's output queue is `extract:queue:nlp`, which is exactly NLP's input queue. Every task builds its `PipelineHelper` from the same `stages` arg, so the chain stays consistent: the previously started stage is the producer by construction, for any subset.

datashare-app/src/main/java/org/icij/datashare/tasks/DeduplicateTask.java
pirhoo: Agreed, with a detail: this queue is a `DocumentQueue<Path>`, so `isLegacySentinel(String)` does not apply here. The stale entry is `Paths.get("POISON")`, currently skipped downstream at `IndexTask.java:84`.

Folded into #2296 alongside the dedup gap, since the fix there is the same predicate.

datashare-app/src/main/java/org/icij/datashare/tasks/DeduplicateTask.java
pirhoo: Real bug, but not a regression from this PR, so I moved it to #2296.

Pre-PR, `CliApp` started `DeduplicateTask` **before** `ScanTask` (the DEDUPLICATE block came first) and it blocked on `inputQueue.take()` until `PATH_POISON`. The stages already overlapped fully, and `removeDuplicates()` already ran on an effectively empty queue, so the gap predates the poison-pill removal and was arguably worse there.

The fix in #2296 replaces the one-shot `removeDuplicates()` with a per-entry seen-set predicate, which is less code than today and has the same memory profile since `removeDuplicates()` already builds a full `HashMap` of the queue.

pirhoo:
Verified each finding. Three land, three I would push back on: details in the inline threads.

**Fixed**
- `TaskResource` security issue: `UPSTREAM_TASK_ID` is stripped in `applyProjectTo`, covering `/findNames`, `/batchUpdate/index` and `/batchUpdate/index/:path:` at once. `PUT /api/task/:id` is a separate pre-existing hole (no `@TaskPolicy`, arbitrary task class + args) and needs its own issue.
- `PipelineTask.upstreamRunning`: jOOQ`s `DataAccessException` is unchecked and escaped. Now `UnknownTask` means finished, `IOException`/`RuntimeException` means still running.

**Moved to #2296**
- The `DeduplicateTask` dedup gap and the missing `PATH_POISON` skip. The dedup gap is real but predates this PR: pre-PR `CliApp` started `DeduplicateTask` before `ScanTask` and it blocked on `take()`, so the stages already overlapped and `removeDuplicates()` already ran on an empty queue.

**Disagree**
- `CliApp:222`: `getOutputQueueNameFor` names the queue after the next stage in the configured list, so the previously started stage is the producer by construction, for any `--stages` subset.
- `ExtractNlpTask:86`: the old budget exited DONE with 0 after 3 x 60s anyway, so the false result predates this PR; restoring it re-adds the timing heuristic the gate replaces.

3 new tests, all confirmed failing on the unfixed source.

caro3801 (approved):
Thanks for making the changes and creating the issue. LGTM!

datashare-app/src/main/java/org/icij/datashare/tasks/PipelineTask.java
pirhoo: Good point, it's fixed as part of https://github.com/ICIJ/datashare/pull/2289/changes/d2e891080ae4e35caf1cd57f085418f6f702f4be

datashare-app/src/main/java/org/icij/datashare/tasks/DeduplicateTask.java
pirhoo: Here is the follow up PR https://github.com/ICIJ/datashare/pull/2297

PR #2306, opened 2026-08-03
feat(artifacts): dedicated ArtifactResource

### LINKED ISSUE #2228: feat(artifacts): dedicated ArtifactResource
Serve all types at `/:project/artifacts/<type>/:id` (deviates from the spec's `/documents/structure/:id`); migrate the existing document artifact endpoints and update the convention doc.

- [ ] `ArtifactResource`: `raw/:id` · `page/:id` + `page/:id/:page` · `structure/:id` (`{pages, formats}`) + `structure/:id/:page?format=md|xhtml`. Shared access control (403/404), `status: complete` + contiguous gating.
- [ ] Migrate `/documents/{src,pages,content/pages}` → `/artifacts/{raw,page}` with deprecation aliases; remove PoC `/documents/structure/*`.
- [ ] Update convention §Serving to `/artifacts/…`; coordinate datashare-client. Tests: 403/404 matrix, `?format=`, aliases.

Refs: PoC `DocumentResource` structure endpoints; existing `/documents/src|pages`.

Part of epic(artifacts).

pirhoo:
Serves per-document artifacts from disk at `/:project/artifacts/<type>/:id`, driven by each document's `manifest.json` under `artifactDir`. Closes #2228.

- feat: add `ArtifactReader`, serving both filesystem and byte-range pagination, with guards for malformed manifests
- feat: add `ArtifactResource` with `raw`, `page` and `structure` routes (`?format=md|xhtml`)
- refactor: extract the source download gate into `DocumentSourceAccess`, shared with `/documents/src`
- refactor: collapse the `structure/` and `pages/` path helpers onto one type-keyed `payloadDir`/`payloadPage`/`payloadContent`
- fix: read pagination through the sealed schemes `ManifestEntry` models, on the unpadded `page-N` names
- docs: deprecate `/documents/content/pages` in favour of the page artifact route

Suggested follow-ups (not in this diff): stream page payloads instead of buffering a `byte[]` bounded only by `Integer.MAX_VALUE`; cache the immutable `COMPLETE` manifest so paging an N-page document is not N Elasticsearch GETs plus N manifest parses; unify the remaining copies of the artifact-dir resolution; add a route-enumerating test so a future route added without a membership gate cannot ship green; decide whether a `pages` block with no `pagination` should read as the convention's single `content.<ext>` rather than falling through to `page-N`; reject a page whose manifest ranges no longer describe the live `content.txt`, the post-swap staleness `PageArtifact` defers to this issue.

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/web/ArtifactResource.java
+++ b/datashare-app/src/main/java/org/icij/datashare/web/ArtifactResource.java
@@ -184,7 +184,7 @@ public class ArtifactResource {
                     DatashareCliOptions.ARTIFACT_DIR_OPT, id, project);
             return null;
         }
-        return ArtifactPath.dir(Path.of(artifactDir.get()).resolve(project), document.getId());
+        return ArtifactPath.dir(ArtifactPath.projectRoot(Path.of(artifactDir.get()), project), document.getId());
     }
 
     private static Map<String, String> structureContentTypes() {
--- a/datashare-app/src/main/java/org/icij/datashare/web/ArtifactResource.java
+++ b/datashare-app/src/main/java/org/icij/datashare/web/ArtifactResource.java
@@ -122,6 +122,9 @@ public class ArtifactResource {
     @Get("/:project/artifacts/structure/:id/:page?routing=:routing&format=:format")
     public Payload structurePage(final String project, final String id, final String page, final String routing,
                                  final String format, final Context context) throws IOException {
+        // Membership gates before format validation, same as every other route here: a non-member
+        // must see 403 regardless of what else is wrong with the request.
+        requireGranted(context, project);
         String extension = ofNullable(format).filter(value -> !value.isBlank()).orElse("md");
         String contentType = STRUCTURE_CONTENT_TYPES.get(extension);
         if (contentType == null) {
--- a/datashare-app/src/main/java/org/icij/datashare/web/ArtifactResource.java
+++ b/datashare-app/src/main/java/org/icij/datashare/web/ArtifactResource.java
@@ -9,6 +9,7 @@ import io.swagger.v3.oas.annotations.responses.ApiResponse;
 import net.codestory.http.Context;
 import net.codestory.http.annotations.Get;
 import net.codestory.http.annotations.Prefix;
+import net.codestory.http.constants.HttpStatus;
 import net.codestory.http.payload.Payload;
 import org.icij.datashare.PropertiesProvider;
 import org.icij.datashare.cli.DatashareCliOptions;
@@ -38,6 +39,14 @@ import static org.icij.datashare.web.errors.ForbiddenException.requireGranted;
 @Singleton
 @Prefix("/api")
 public class ArtifactResource {
+    // Extensions are fixed by type (convention §71). The map is the ?format= whitelist and content-type
+    // lookup; the list is the order of record for disk probing and for the error message, because
+    // Map.of has no defined iteration order.
+    private static final Map<String, String> STRUCTURE_CONTENT_TYPES = Map.of(
+            "md", "text/markdown;charset=UTF-8",
+            "xhtml", "application/xhtml+xml;charset=UTF-8");
+    private static final List<String> STRUCTURE_FORMATS = List.of("md", "xhtml");
+
     private final Indexer indexer;
     private final PropertiesProvider propertiesProvider;
     private final DocumentSourceAccess sources;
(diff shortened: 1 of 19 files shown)


datashare-app/src/main/java/org/icij/datashare/web/ArtifactResource.java
```diff
@@ -0,0 +1,244 @@
...
+public class ArtifactResource {
+    private static final Logger LOGGER = LoggerFactory.getLogger(ArtifactResource.class);
+    // Extensions are fixed by type (convention §71). One ordered map, so adding a format is one
+    // edit: its key set is the ?format= whitelist, the disk-probe order and the error-message order.
+    private static final Map<String, String> STRUCTURE_CONTENT_TYPES = structureContentTypes();
+
+    private final Indexer indexer;
+    private final PropertiesProvider propertiesProvider;
+    private final DocumentSourceAccess sources;
+    private final ArtifactReader reader = new ArtifactReader(new FilesystemManifestRepository());
```
caro3801, line 59: ArtifactReader is manually wired (`new ArtifactReader(new FilesystemManifestRepository())`) instead of injecting ArtifactReader through the Guice constructor like other params. 

question: Is there a particular reason it by passes the DI  since `ArtifactReader` already takes `ManifestRepository` as a constructor param? 

Two costs: it's the only untestable seam in the class (forces `ArtifactResourceTest` to use real `TemporaryFolder` files instead of a stub), and it's inconsistent with this same branch's own refactor pattern.

suggested fix: add `ArtifactReader` to `ArtifactResource's` constructor, bind `FilesystemManifestRepository` to `ManifestRepository` in the Guice module  (`LocalMode.java`/`ServerMode.java`). That would improve testability and consistency

datashare-index/src/main/java/org/icij/datashare/text/artifact/ArtifactReader.java
```diff
@@ -0,0 +1,126 @@
...
+    /** The type's entry when it exists and is servable, else null. Callers map null to 404 and
+     *  never inspect status themselves, so "servable" has one definition. A corrupt manifest.json
+     *  (bad syntax, or an unrecognised status) is not servable either: manifest.json is written by
+     *  other producers, including datashare-python, so this is the boundary where a malformed one
+     *  reads as "not found" rather than propagating as a 500 to whoever asks for a serving read. */
+    public ManifestEntry servableEntry(Path docArtifactDir, ArtifactType type) throws IOException {
+        ManifestEntry entry;
+        try {
+            entry = manifests.get(docArtifactDir, type.token());
+        } catch (JsonProcessingException malformed) {
```
caro3801, line 35: important: [FilesystemManifestRepository.java:110](https://github.com/ICIJ/datashare/blob/fix/2298-artifact-task-args/datashare-index/src/main/java/org/icij/datashare/text/artifact/FilesystemManifestRepository.java#L110) throws plain `IOException` for "manifest is not a JSON object," but `servableEntry` only catches `JsonProcessingException`. 
Result: this corruption shape slips past the intended 404 handling and surfaces as an uncaught 500 - contradicting the branch's own stated intent. 
suggested fix: throw `JsonProcessingException` in `FilesystemManifestRepository.java` (or a dedicated subtype) so it's caught alongside the syntax-error case.

datashare-app/src/main/java/org/icij/datashare/web/ArtifactResource.java
```diff
@@ -0,0 +1,244 @@
...
+    @ApiResponse(responseCode = "400", description = "if format is not one of md, xhtml")
+    @ApiResponse(responseCode = "403", description = "forbidden if the user doesn't have access to the project")
+    @ApiResponse(responseCode = "404", description = "if the document, the artifact, the page, or that format is not found")
+    @Get("/:project/artifacts/structure/:id/:page?routing=:routing&format=:format")
+    public Payload structurePage(final String project, final String id, final String page, final String routing,
+                                 final String format, final Context context) throws IOException {
+        // Membership gates before format validation, same as every other route here: a non-member
+        // must see 403 regardless of what else is wrong with the request.
+        requireGranted(context, project);
+        String extension = ofNullable(format).filter(value -> !value.isBlank()).orElse("md");
```
caro3801: nitpick: format like "md" and "xhtml" could be refactored as enum or constant

caro3801 (changes requested):
Looks good to me the implementation is solid (extraction, edge-case coverage, exception handling).   
I put 2 comments about DI and exception-type mismatch that might worth changes before merging

datashare-index/src/main/java/org/icij/datashare/text/artifact/ArtifactReader.java
pirhoo: Confirmed, fixed in 4242aab0d. `JsonProcessingException` has only protected constructors, so it now throws `MismatchedInputException.from(...)`, a subtype that lands in the same catch. Covered by `test_servable_entry_is_null_when_manifest_json_is_not_an_object` (red before the fix, with the `is not a JSON object` IOException escaping as a 500).

datashare-app/src/main/java/org/icij/datashare/web/ArtifactResource.java
pirhoo: There is no Guice binding for `ManifestRepository` anywhere: `ArtifactTask:141`, `IndexTask:137` and `ArtifactCoverageChecker:45` all hand-wire `new FilesystemManifestRepository()` too, so this follows the artifact code's existing convention. Binding it just for the web read path would make 1 of 4 injected.

On testability, `ArtifactResourceTest` already builds the resource directly, so a param would be free, but it uses `TemporaryFolder` on purpose: what it checks is the on-disk contract (pages written through `ArtifactPath`, ranges sliced out of a real `content.txt`). The reader's own rules are covered by `ArtifactReaderTest`.

Happy to inject all four call sites in a follow-up if we want DI here.

datashare-app/src/main/java/org/icij/datashare/web/ArtifactResource.java
pirhoo: Extracted `MARKDOWN` and `XHTML` constants in f4351d7fd. Kept the ordered map rather than an enum: its key set is already the `?format=` whitelist, the disk-probe order and the error-message order in one place.

datashare-app/src/main/java/org/icij/datashare/web/ArtifactResource.java
caro3801: Fair - not worth partial DI for this alone

caro3801 (approved):
Thanks for the changes

pirhoo:
Thanks for the review 🚀

PR #2308, opened 2026-08-03
refactor(tasks): ArtifactTask reads its configuration from the task args

### LINKED ISSUE #2298: refactor: ARTIFACT is the only pipeline stage reading process-wide properties instead of its task args
Spotted by @caro3801 while reviewing #2262.

`ArtifactTask` reads the process-wide `PropertiesProvider` (settings file overridden by CLI), while every other pipeline stage builds one from its own task args. On the CLI those args never include the settings file, so config set only in `datashare.properties` is honoured by ARTIFACT and ignored by every other stage: `artifacts=raw` there makes `--stages ARTIFACT` produce artifacts and `--stages INDEX` produce nothing. REST is unaffected, it folds the settings file into the args already.

Fix in one of two directions: make ARTIFACT args-based like its siblings (consistent, but it stops honouring settings-file-only values), or fold the settings file into the CLI properties before they become task args (what users expect, but CLI defaults then override settings values, as `--artifactsForce` already hit).

pirhoo:
Closes #2298.

ARTIFACT was the only pipeline stage reading process-wide properties instead of its task args. Moving it to task args exposed four bugs in how a `-s` settings file resolves against CLI option defaults, on both the legacy jopt parser and the picocli one, so those are fixed here too.

- refactor: `ArtifactTask` reads its configuration from the task args instead of the injected process-wide `PropertiesProvider`, like every other stage already does
- fix: substitute a settings-file value before skipping an option that declares no default, so a settings-only `artifactDir` reaches the ARTIFACT stage instead of failing it
- fix: consult the settings file only when the operator passed `-s`, so a classpath `datashare.properties` no longer outranks every option default
- fix: exclude arity-0 booleans from the picocli default value provider, so a settings file setting `resume=true` no longer makes `-r` mean "do not resume"
- fix: read the settings file alone via the new `PropertiesProvider.getFileProperties()`, so `DS_DOCKER_*` env vars keep the precedence `CommonMode` gives them instead of beating an option default only when `-s` happens to be passed
- fix: substitute the settings value as env-over-file, so a `DS_DOCKER_*` var keeps outranking the settings file it substitutes
- fix: build the CLI pipeline task args from the merged provider, so an env-only value still reaches a stage
- test: cover the precedence bugs in `DatashareCliTest`, `SettingsDefaultsTest` and `PropertiesProviderTest`

## Not covered

- `-s<path>` (attached short form) is still not parsed by `settingsPathFrom`, so the settings file is ignored on the picocli path for that invocation style.
- `ArtifactTask` reads task args through `java.util.Properties`, so a non-String arg from the API or Temporal is silently dropped.
- Starting a stage through the generic `PUT /task/:id` with a minimal body needs the full config in `args`: `ScanTask` and `IndexTask` already did, so ARTIFACT is now consistent with them rather than falling back to the server's `datashare.conf`.
- A settings file cannot set `noDigestProject` or `browserOpenLink` (#2339): pre-existing, both write `false` unconditionally.
- `DS_DOCKER_*` env vars still lose to any option that declares a default, because `CommonMode` layers the CLI properties on top. That contradicts the precedence documented on `SettingsResource.patchSettings`, but it predates this branch and changing it would affect every Docker deployment, so it is left alone here.

### DIFF
--- a/datashare-app/src/test/java/org/icij/datashare/tasks/ArtifactTaskFixture.java
+++ b/datashare-app/src/test/java/org/icij/datashare/tasks/ArtifactTaskFixture.java
@@ -3,13 +3,11 @@ package org.icij.datashare.tasks;
 import org.icij.datashare.asynctasks.Task;
 import org.icij.datashare.user.User;
 
-import java.util.HashMap;
 import java.util.Map;
 
 /** Builds the Task an ArtifactTask reads its configuration from. Shared by the three artifact test classes. */
 class ArtifactTaskFixture {
-    /** Task args are mutable in production (the launcher adds the upstream id), so copy into a HashMap. */
     static Task<Long> taskWith(Map<String, Object> args) {
-        return new Task<>(ArtifactTask.class.getName(), User.local(), new HashMap<>(args));
+        return new Task<>(ArtifactTask.class.getName(), User.local(), args);
     }
 }
--- a/datashare-app/src/test/java/org/icij/datashare/tasks/ArtifactTaskFixture.java
+++ b/datashare-app/src/test/java/org/icij/datashare/tasks/ArtifactTaskFixture.java
@@ -1,0 +1,15 @@
+package org.icij.datashare.tasks;
+
+import org.icij.datashare.asynctasks.Task;
+import org.icij.datashare.user.User;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/** Builds the Task an ArtifactTask reads its configuration from. Shared by the three artifact test classes. */
+class ArtifactTaskFixture {
+    /** Task args are mutable in production (the launcher adds the upstream id), so copy into a HashMap. */
+    static Task<Long> taskWith(Map<String, Object> args) {
+        return new Task<>(ArtifactTask.class.getName(), User.local(), new HashMap<>(args));
+    }
+}
--- a/datashare-app/src/test/java/org/icij/datashare/tasks/ArtifactTaskTest.java
+++ b/datashare-app/src/test/java/org/icij/datashare/tasks/ArtifactTaskTest.java
@@ -61,7 +61,26 @@ public class ArtifactTaskTest {
 
     @Test(expected = IllegalArgumentException.class)
     public void test_missing_artifact_dir() {
-        new ArtifactTask(factory, mockEs, new PropertiesProvider(Map.of()), new UpstreamGate.Factory(taskRepository), new Task<>(ArtifactTask.class.getName(), User.local(), Map.of()), null);
+        new ArtifactTask(factory, mockEs, new UpstreamGate.Factory(taskRepository),
+                ArtifactTaskFixture.taskWith(Map.of()), null);
+    }
+
+    @Test(timeout = 10000)
+    public void test_configuration_comes_from_the_task_args() throws Exception {
+        indexEmbeddedDoc();
+        DocumentQueue<String> queue = factory.createQueue("extract:queue:artifact", String.class);
+        queue.add(EMBEDDED_DOC_SHA256);
+
+        // args carry the whole configuration, exactly as IndexTask and every other stage expects.
+        Long numberOfDocuments = new ArtifactTask(factory, mockEs,
+                new UpstreamGate.Factory(taskRepository),
+                ArtifactTaskFixture.taskWith(Map.of(
+                        "artifactDir", artifactDir.getRoot().toString(),
+                        "defaultProject", "prj")), null)
+                .call();
+
+        assertThat(numberOfDocuments).isEqualTo(1);
+        assertThat(artifactDir.getRoot().toPath().resolve("prj/6a/bb").toFile()).isDirectory();
     }
 
     @Test(timeout = 10000)
@@ -145,13 +164,14 @@ public class ArtifactTaskTest {
         queue.add(secondId);
 
         CountDownLatch bothInFlight = new CountDownLatch(2);
-        PropertiesProvider props = new PropertiesProvider(Map.of(
+        Map<String, Object> config = Map.of(
                 "artifactDir", artifactDir.getRoot().toString(),
                 "defaultProject", "prj",
-                "parallelism", "2"));
-        Task<Long> task = new Task<>(ArtifactTask.class.getName(), User.local(), new HashMap<>());
+                "parallelism", "2");
+        PropertiesProvider props = new PropertiesProvider(config);
+        Task<Long> task = ArtifactTaskFixture.taskWith(config);
 
-        ArtifactTask artifactTask = new ArtifactTask(factory, mockEs, props, new UpstreamGate.Factory(taskRepository), task, null) {
+        ArtifactTask artifactTask = new ArtifactTask(factory, mockEs, new UpstreamGate.Factory(taskRepository), task, null) {
             @Override
             protected SourceExtractor createSourceExtractor() {
                 return new SourceExtractor(props) {
@@ -188,13 +208,13 @@ public class ArtifactTaskTest {
         queue.add(EMBEDDED_DOC_SHA256);
         queue.add(secondId);
 
-        PropertiesProvider props = new PropertiesProvider(Map.of(
+        Map<String, Object> config = Map.of(
                 "artifactDir", artifactDir.getRoot().toString(),
                 "defaultProject", "prj",
-                "parallelism", "2"));
-        Task<Long> task = new Task<>(ArtifactTask.class.getName(), User.local(), new HashMap<>());
+                "parallelism", "2");
+        Task<Long> task = ArtifactTaskFixture.taskWith(config);
 
-        ArtifactTask artifactTask = new ArtifactTask(factory, mockEs, props, new UpstreamGate.Factory(taskRepository), task, null) {
+        ArtifactTask artifactTask = new ArtifactTask(factory, mockEs, new UpstreamGate.Factory(taskRepository), task, null) {
             @Override
             protected SourceExtractor createSourceExtractor() {
                 throw new IllegalStateException("no extractor");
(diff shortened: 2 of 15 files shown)


datashare-cli/src/main/java/org/icij/datashare/cli/command/DatashareCommand.java
```diff
@@ -51,4 +53,47 @@ public Properties collectProperties() {
...
+        String settingsPath = settingsPathFrom(args);
+        if (settingsPath == null) {
+            // Not "no file": PropertiesProvider(null) resolves a classpath datashare.properties, which
+            // is not what -s asked for.
+            return;
+        }
+        // The file alone, not getProperties(), which also folds in DS_DOCKER_* env vars: those are a
+        // separate tier that CommonMode ranks, and promoting them here would make them beat an option
+        // default only when -s happens to be passed.
+        Properties settings = new PropertiesProvider(settingsPath).getFileProperties();
```
caro3801, line 73: suggestion: It looks like the settings file now gets read from disk twice per CLI invocation -
once here (or in `DatashareCli.asProperties`) to build the CLI `Properties`, and again inside
`CommonMode`'s constructor to build the fully-merged provider. 
Could this be consolidated into a single read, e.g. by threading the already-parsed `Properties` into `CommonMode` instead of having it re-read the path?

datashare-cli/src/main/java/org/icij/datashare/cli/command/DatashareCommand.java
```diff
@@ -51,4 +53,47 @@ public Properties collectProperties() {
...
+        CommandLine.IDefaultValueProvider fromSettings = new CommandLine.PropertiesDefaultProvider(settings);
+        // Arity-0 booleans are excluded: picocli sets a matched flag to !defaultValue, so taking that
+        // default from the file inverts the flag, and "resume=true" in the file would make -r mean
+        // "do not resume". They keep their declared default here and CommonMode's overrideWith fold-in
+        // is what applies the file's value, as it did before this provider existed.
+        commandLine.setDefaultValueProvider(
```
caro3801, line 81: question: The comment says `overrideWith` still lets the settings file's value through for these excluded booleans - but that's only true if `toProperties()` skips the key on `false` (the `putIfTrue` convention). `noDigestProject` (`GlobalOptions.java:150`) and `browserOpenLink` (`ServerOptions.java:147`) don't follow it - they always write `false` explicitly, so a `true` in the settings file can never win; the CLI's default silently overrides it every time.

For reference, `PipelineOptions.resume` is the same shape (arity-0, default `false`) and correctly uses `putIfTrue`, so these two really do look like outliers rather than an intentional second pattern. (`ocr`/`followSymlinks` also call `put` unconditionally, but those are `arity = "1"` - the user must state a value explicitly - so they're a different case and not part of this gap.)

Is this known/accepted, or worth a follow-up to switch those two to `putIfTrue`?

datashare-cli/src/main/java/org/icij/datashare/cli/DatashareCli.java
```diff
@@ -223,12 +225,25 @@ OptionParser createParser() {
...
-            if (options.has(spec) || !entry.getValue().isEmpty()) {
-                properties.setProperty(
-                        asPropertyKey(prefix, spec),
-                        asPropertyValue(entry.getValue()));
+            String key = asPropertyKey(prefix, spec);
+            boolean typed = options.has(spec);
+            // Substituting the file's value, rather than omitting the key, matters for the consumers
+            // that read these raw properties without CommonMode's settings fold-in: Main.startApplication
+            // for logLevel, CliApp.start for pluginsDir/extensionsDir, ArtifactTask for artifactDir.
+            if (!typed && settings.containsKey(key)) {
```
caro3801, line 251: crucial: This can break the `DS_DOCKER_*` > settings-file precedence that `CommonMode` is supposed to enforce.

When an option like `artifactDir` has no jopt default and wasn't typed on the CLI, `asProperties()` now pulls its value from the settings file into the returned CLI `Properties`. Later, `CommonMode.overrideWith()` does a blind `Properties.putAll`, so that settings-file value clobbers the correctly env-merged value.

Concrete case: operator sets `DS_DOCKER_ARTIFACT_DIR=/env/path` and also has `artifactDir=/file/path` in the settings file, but never types `--artifactDir`. The env value silently loses.

Could we scope this substitution to only the options that actually need it, or skip it when an env var already supplied the value, instead of applying it whenever `!typed && settings.containsKey(key)`?

datashare-app/src/test/java/org/icij/datashare/tasks/ArtifactTaskFixture.java
```diff
@@ -0,0 +1,15 @@
...
+import org.icij.datashare.user.User;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/** Builds the Task an ArtifactTask reads its configuration from. Shared by the three artifact test classes. */
+class ArtifactTaskFixture {
+    /** Task args are mutable in production (the launcher adds the upstream id), so copy into a HashMap. */
+    static Task<Long> taskWith(Map<String, Object> args) {
+        return new Task<>(ArtifactTask.class.getName(), User.local(), new HashMap<>(args));
```
caro3801: hint: `taskWith()` copies `args` into a new `HashMap` before handing it to the `Task` constructor,
but `Task`'s own constructor chain already copies internally. The extra copy here is redundant, and the comment above it ("Task args are mutable in production... so copy into a HashMap") could mislead a future reader into thinking `Task` doesn't already do this itself.

Suggested fix: drop the new HashMap<>(...) wrapper and pass args straight through, and remove/fix the misleading comment.

datashare-app/src/test/java/org/icij/datashare/tasks/ArtifactTaskTest.java
caro3801: hint: `ArtifactTask` extends `PipelineTask`, which exposes a protected `propertiesProvider` field built from the same `task` config the test already passes in - no need for an extra local `props` variable L242.

```suggestion
                return new SourceExtractor(propertiesProvider) {
```

caro3801 (changes requested):
Nice work getting `ArtifactTask` out of `PropertiesProvider` to explicit task args, that's a great improvement for how config works

Most of the comments below fall into two buckets:

- **CLI config precedence** - a few spots where settings-file values can now sneak in ahead of `DS_DOCKER_*` env vars or get read twice, plus a couple of small gaps (a missing warn log, one option parsing path getting out of sync).
- **The generic task endpoint** - starting an `ArtifactTask` via `PUT /task/:id` with a minimal body now fails instead of falling back to server defaults, since the task no longer reads `datashare.conf`. Worth deciding if that should be handled or just documented.

The rest are small test cleanups (some redundant `PropertiesProvider`/`HashMap` copies, a couple of stale comments) that don't affect production behavior.

Here are few comments on files impacted by the changes out of the modified files: 

* In [`datashare-app/src/main/java/org/icij/datashare/web/TaskResource.java:164`](https://github.com/ICIJ/datashare/blob/main/datashare-app/src/main/java/org/icij/datashare/web/TaskResource.java#L164)

important: Since `ArtifactTask` now reads all its configuration from `taskView.args`
(`ArtifactTask.java:58-63`) instead of the injected process-wide `PropertiesProvider`, a client that starts one through the generic `PUT /task/:id` endpoint with a minimal body (e.g. `{"name":"org.icij.datashare.tasks.ArtifactTask","args":{}}`) now fails at construction with `IllegalArgumentException`, where before it would fall back to the server's own `datashare.conf`.
Could we either merge server-side defaults into `taskView.args` before dispatch here, or document that callers of this generic endpoint must now supply the full config explicitly?

* In [`datashare-cli/src/test/java/org/icij/datashare/cli/command/RetroCompatibilityTest.java:36`
](https://github.com/ICIJ/datashare/blob/main/datashare-cli/src/test/java/org/icij/datashare/cli/command/RetroCompatibilityTest.java#L36)

hint: `parseNew()` duplicates the picocli `CommandLine` bootstrap from `Main.runPicocli` but wasn't updated to call `DatashareCommand.applySettingsDefaults(commandLine, args)` before `execute(args)`, unlike `Main.java` and `AbstractDatashareCommandTest.configure()`. No current test exercises `-s`/`--settings` here so nothing breaks today, but a future settings-file test added to this file would silently diverge from real app behavior. Could we route this through the same bootstrap helper to keep the two in sync?

datashare-cli/src/main/java/org/icij/datashare/cli/command/DatashareCommand.java
pirhoo: Leaving as is: `CommonMode` needs the path anyway (`SettingsResource:78` saves through that provider), so threading the parsed `Properties` in would move the read, not remove it. Each provider reads the file once.

datashare-cli/src/main/java/org/icij/datashare/cli/command/DatashareCommand.java
pirhoo: Right, the comment was too broad. Narrowed it in 76abeed0f, and filed #2339 for switching the two to `putIfTrue`: pre-existing on main, and it changes the effective value of both flags for anyone who set them in a settings file, so not in this PR.

datashare-cli/src/main/java/org/icij/datashare/cli/DatashareCli.java
pirhoo: Fixed in 5cc706311: the substituted value is now env-over-file (`PropertiesProvider.getEnvProperties()`), so `DS_DOCKER_ARTIFACT_DIR` wins and the substitution stays scoped to the keys the file declares. Test: `test_a_docker_env_var_outranks_the_settings_file_value`.

You also pointed at the other half: with the env var set and *no* file entry the key was dropped entirely, so ARTIFACT failed where main read the merged provider. b5e67ecb9 builds the pipeline task args from `mode.properties()` instead of the raw CLI properties.

datashare-app/src/test/java/org/icij/datashare/tasks/ArtifactTaskFixture.java
pirhoo: Done in ad32a757a. `Task.addTo` (`Task.java:277`) copies into a `LinkedHashMap` and `Task.java:101` stores it unmodifiable, so the wrapper and the comment were both wrong.

datashare-app/src/test/java/org/icij/datashare/tasks/ArtifactTaskTest.java
pirhoo: Done in ad32a757a, and removed the same leftover local from the two other tests that already used `propertiesProvider`.

pirhoo:
Thanks, all handled. The two items from the review body:

**`TaskResource:164`** (generic `PUT /task/:id`): documented, not defaulted. `ScanTask:36` already NPEs on `args:{}` and `IndexTask:76` reads args the same way, so ARTIFACT is now consistent with the other stages, with a named `IllegalArgumentException` instead of an NPE, and nothing but `CliApp` starts it. Added to *Not covered*.

**`RetroCompatibilityTest:36`**: fixed in ad32a757a, `parseNew()` now calls `applySettingsDefaults` like `Main` does. I kept the one line rather than extracting a shared bootstrap helper for a single test.

caro3801 (approved):
Perfect 💯

Issue #2194, opened 2026-06-16
epic: import structured entities from tabular data using a data schema


pirhoo:
Mode A is tabular column-mapping. A user supplies a CSV/XLSX in their project plus a mapping (hand-authored JSON/YAML) that says how the table's columns fit the data schema. A background task executes the mapping, materializes entities into a statement store (the system of record), and projects them into a searchable `<project>.entities` index. The mapping is persisted so the extraction is reproducible.

The only target model is FollowTheMoney, but it is built on a model-neutral core: the `TargetModel` and exporter seams are defined as interfaces now, with FtM the sole implementation. Adding Schema.org or Wikidata later means a new provider class, no core rework.

Out of MVP scope: Mode B (regex over extracted text), cross-source entity resolution (fuzzy merging of records that refer to the same real-world entity, a separate later layer with expert validation, distinct from the deterministic key-based identity the mapping already gives us), Schema.org and Wikidata adapters, the Python `ingest-file` sidecar, Tika un-flattening, and any UI (the mapping is hand-authored).

The work is broken into the sub-issues below; each carries its own subtask checklist and a `Depends on:` line. Suggested order: superseded by the [2026-08-19 state check](https://github.com/ICIJ/datashare/issues/2194#issuecomment-5346473298), which records the wave order, what parallelizes, and five corrections to the plan above (notably that the FtM Java library supplies no runtime model, so #2200 is a build item and ships with #2197).

## Architecture

```mermaid
flowchart TD
    USER(["User / API client"])
    REST["REST + CLI trigger · #2207"]
    TASK["StructuredEntityExtractionTask<br/>+ factory wiring · #2206"]

    SRC[("Tabular source<br/>CSV · XLS/XLSX · JSON/JSONL")]
    READER["RowSource SPI<br/>commons-csv · POI · Jackson · #2198"]
    MAPPING[("extraction_mapping<br/>FtM mapping-YAML · #2201")]

    subgraph CORE["Model-neutral core · #2197"]
        EXEC["Mapping executor<br/>rows + mapping → statements · #2204"]
        REG["TargetModel registry<br/>+ statement atom"]
        FTM["FtmTargetModel<br/>in-JVM defaultModel.json · #2200"]
        EXEC --> REG
        REG --> FTM
    end

    subgraph PERSIST["Persistence — system of record"]
        STORE[("statement table<br/>SHA1 ids · idempotent upsert · #2202")]
    end

    subgraph DERIVED["Derived / disposable projections"]
[...]

PR #2311, opened 2026-08-04
feat(index): authorize a "<project>.entities" index against its base project

### LINKED ISSUE #2199: refactor: proxy authorization for namespaced indices
Part of #2194.

Let a user who can access a project also reach that project's new entities index. Today the search proxy grants access only when the index name exactly equals a project the user can see, so a name like "myproject.entities" is refused.

- [ ] In the access check, map an index name back to its project: strip a known suffix (only "entities" for now) and check access to the base project.
- [ ] Do not switch to prefix matching, or access to one project would leak into similarly named ones.
- [ ] Make sure index-name validation allows a dot in the name.
- [ ] Add tests: a user can reach their own ".entities" index but not another project's.

This only touches the search proxy. Project-management permissions are unchanged.

pirhoo:
Lets a user who can access a project also reach that project's `<project>.entities` index through the search proxy. Part of #2194.

- refactor: allow a dot between index-name segments, never leading, and refuse a leading underscore, so the elasticsearch system indices and `_all` stay unreachable
- refactor: map an index back to its project by stripping a known `.entities` suffix before the grant check, exact match on the remainder
- fix: use `matches()` rather than `find()` in the index-name check, so a trailing line terminator no longer validates
- fix: require a grant on the index's base project to create, close or open it, per index in a comma-separated list
- fix: record async-search ownership against the base project, so a poll re-checks the string its submit authorized
- fix: mode-gate `createIndex` the way `_close` and `_open` already are
- fix: validate a new project name as an index name on both create paths, so a row can't be saved under a name elasticsearch then refuses
- test: a user reaches their own `.entities` index, and is refused another project's, an unknown suffix, a prefix match, and a foreign index piggybacked on a granted one

Known gaps: `esHead`, `esOptions` and `_snapshot/*` still forward caller input no grant check has seen (#2312), and the suffix rule authorizes on name shape rather than proven ownership, so a third-party `<project>.entities` in a shared cluster is readable by anyone granted `<project>`.

Closes #2199

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/web/ProjectResource.java
+++ b/datashare-app/src/main/java/org/icij/datashare/web/ProjectResource.java
@@ -94,7 +94,7 @@
                 requestBody = @RequestBody(content = @Content(mediaType = "application/json", schema = @Schema(implementation = Project.class)))
         )
         @ApiResponse(responseCode = "201", description = "if project and index have been created")
-        @ApiResponse(responseCode = "400", description = "if project name is empty or contains a dot")
+        @ApiResponse(responseCode = "400", description = "if project name is empty, contains a dot, or starts with an underscore")
         @ApiResponse(responseCode = "400", description = "if project path is not allowed for the project")
         @ApiResponse(responseCode = "409", description = "if project exists")
         @ApiResponse(responseCode = "500", description = "project creation in DB or index creation failed")
@@ -107,6 +107,8 @@
                 return PayloadFormatter.error("`name` field is required.", HttpStatus.BAD_REQUEST);
             } else if (project.getName().contains(".")) {
                 return PayloadFormatter.error("`name` must not contain a dot.", HttpStatus.BAD_REQUEST);
+            } else if (project.getName().startsWith("_")) {
+                return PayloadFormatter.error("`name` must not start with an underscore.", HttpStatus.BAD_REQUEST);
             }
             Project effectiveProject = isProjectSourcePathNull(project) ? withDefaultSourcePath(project) : project;
             if (!dataDirVerifier.allowed(effectiveProject.getSourcePath())) {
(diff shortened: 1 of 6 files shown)


datashare-app/src/main/java/org/icij/datashare/web/ProjectResource.java
```diff
@@ -56,6 +57,9 @@
     @Singleton
     @Prefix("/api/project")
     public class ProjectResource {
+        private static final Pattern INDEX_SAFE_NAME = Pattern.compile("[a-z0-9][-a-z0-9_]{0,254}");
```
caro3801: important: The pattern between the HTTP client and the CLI are different: a name like `my_project` is accepted here yet fails `Project.NAME_PATTERN` in `Validators.projectName` (`datashare-cli/src/main/java/org/icij/datashare/cli/Validators.java:79-83`) and in `ProjectAdminServiceImpl.validate` (`datashare-app/src/main/java/org/icij/datashare/project/admin/ProjectAdminServiceImpl.java:419-422`) 
So such a project can be created through the web API but never granted, revoked, deleted through
the CLI, or recreated through the admin service.

Could we harmonize this so every project created here stays manageable everywhere else ?

caro3801: suggestion: Could we cap the project name length so the derived `<name>.entities` index name stays under Elasticsearch's 255-byte index name limit? 

Right now a name of up to 255 chars passes this regex, but appending the 9-char `.entities` suffix can push the companion index name past that limit, and nothing here catches it before it fails later at index-creation time.

caro3801 (changes requested):
Solid PR overall, the suffix-strip logic, the `matches()` vs `find()` fix, the per-index grant loop and the mode-gating on `createIndex` all checked out cleanly against their tests and callers.
Worth looking at the project name name validation before merging, it could later leave CLI management unreachable for underscore-named projects. The other comment is a small suggestion to keep the 255-char elasticsearch cap compatible with the `.entities` suffix.

datashare-app/src/main/java/org/icij/datashare/web/ProjectResource.java
pirhoo: Harmonized in d6191a3eb: both creation paths now use `Project.NAME_PATTERN`, so an underscore, a single-char or a >64-char name is refused here too. Existing projects stay updatable, the guard is create-only.

pirhoo: Covered by the same change: `NAME_PATTERN` caps the name at 64 chars, so `<name>.entities` is at most 73 bytes. Test added for the 255-char case.

PR #2319, opened 2026-08-12
feat(artifacts): search a document's structure pages

pirhoo:
Adds an in-document search over a document's persisted structure pages, counting a query's occurrences per page the way `/documents/searchContent` counts them over the indexed content.

- feat: add `GET /:project/artifacts/structure/search/:id`, answering `{count, pages, scanned, hits}`
- feat: port the `searchOccurrences.painless.java` folding rules to Java, with a test pinning the script so the two counters cannot drift apart
- fix: render structure pages with the OCR the document was indexed with, and record it in the artifact fingerprint
- fix: bound the page total a manifest can advertise, so a total written by another producer cannot become an unbounded loop
- fix: count an unreadable page as unscanned rather than failing the whole search, and log it once per scan instead of once per page
- chore: the OCR fingerprint gains `ocr` and `ocrStrategy`, so every existing structure artifact is stale and the next ARTIFACT run re-extracts

### How it works, and what it does not do

The scan reads one `page-N.md` at a time and folds it with the rules ported from `searchOccurrences.painless.java`, so the endpoint and `/documents/searchContent` fold text identically. Occurrences are counted per page with `indexOf` stepping by the raw query length, and summed. Work is linear in pages, capped by a 10s budget: `scanned` below `pages` means the count is a floor, not a total.

Known limits, all deliberate:

- Counts the stored Markdown, not the indexed content, so flexmark escaping, table pipes and link URLs are part of the searched text. XHTML is rejected for this reason.
- A match straddling a page break counts zero here and one in Elasticsearch, which scans the concatenated `content` field.
- Original language only: structure pages are rendered from the source bytes, so there is no `targetLanguage`.
- A scanned PDF renders empty unless the run passes `--ocrStrategy`, which is the same condition under which INDEX extracted nothing from it.

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/web/ArtifactResource.java
+++ b/datashare-app/src/main/java/org/icij/datashare/web/ArtifactResource.java
@@ -149,33 +149,38 @@ public class ArtifactResource {
                 : payload;
     }
 
-    @Operation(description = "Counts a query's occurrences in a document's structure pages, per page.",
+    @Operation(description = "Counts a query's occurrences in a document's Markdown structure pages, per page. "
+            + "Markdown only: the XHTML the page route serves is markup, so counting over it reports tag "
+            + "names, class names and link targets as occurrences of the user's term. Counts the document's "
+            + "original language: structure pages are rendered from the source bytes, so unlike "
+            + "/documents/searchContent this route has no targetLanguage and never counts a translation.",
             parameters = {
                     @Parameter(name = "project", description = "the project id", in = ParameterIn.PATH),
                     @Parameter(name = "id", description = "the document id", in = ParameterIn.PATH),
                     @Parameter(name = "query", description = "the term to count, matched literally", in = ParameterIn.QUERY),
-                    @Parameter(name = "routing", description = "routing key if not a root document", in = ParameterIn.QUERY),
-                    @Parameter(name = "format", description = "md (default) or xhtml", in = ParameterIn.QUERY)
+                    @Parameter(name = "routing", description = "routing key if not a root document", in = ParameterIn.QUERY)
             }
     )
-    @ApiResponse(responseCode = "200", description = "JSON {\"count\": N, \"hits\": [{\"page\": P, \"count\": N}]}")
-    @ApiResponse(responseCode = "400", description = "if the query is blank, or format is not one of md, xhtml")
+    @ApiResponse(responseCode = "200", description = "JSON {\"count\": N, \"pages\": P, \"scanned\": S, "
+            + "\"hits\": [{\"page\": P, \"count\": N}]}. scanned below pages means the artifact lost pages "
+            + "between its manifest and disk, so the counts are a floor rather than a total.")
+    @ApiResponse(responseCode = "400", description = "if the query is blank")
     @ApiResponse(responseCode = "403", description = "forbidden if the user doesn't have access to the project")
-    @ApiResponse(responseCode = "404", description = "if the document, the artifact, or that format is not found")
-    @Get("/:project/artifacts/structure/:id/search?query=:query&routing=:routing&format=:format")
-    public Payload structureSearch(final String project, final String id, final String query, final String routing,
-                                   final String format, final Context context) throws IOException {
+    @ApiResponse(responseCode = "404", description = "if the document, the artifact, or its markdown is not found")
+    // The verb before the id, as /documents/searchContent/:id has it, so this cannot collide with the
+    // :page route. Under that route's shape it would win only on a tie-break: fluent-http's UriParser
+    // orders on the count of :params across the whole pattern, query string included, and both hold
+    // five. A sixth would sort this second, :page would match "search", and every search would 404.
+    @Get("/:project/artifacts/structure/search/:id?query=:query&routing=:routing")
+    public Payload structureSearch(final String project, final String id, final String query,
+                                   final String routing, final Context context) throws IOException {
         requireGranted(context, project);
         if (query == null || query.isBlank()) {
             // Not the 404 /documents/searchContent gives an empty query: a missing parameter is a
             // malformed request, and a whitespace query would scan every page to match every space.
             return PayloadFormatter.error("a non-blank query is required", HttpStatus.BAD_REQUEST);
         }
-        String extension = structureExtension(format);
-        if (extension == null) {
-            return unsupportedFormat(format);
-        }
-        return search(docArtifactDir(project, id, routing), extension, query);
+        return search(project, id, routing, query);
     }
 
     @Operation(description = "Fetches a document's raw (embedded or source) bytes. Same access rules as /documents/src: project membership, the project's download restriction, and the root-document size limit.",
@@ -276,11 +281,14 @@ public class ArtifactResource {
         return new Payload(contentType, bytes).withHeader("X-Content-Type-Options", "nosniff");
     }
 
-    private Payload search(final Path docArtifactDir, final String extension, final String query) throws IOException {
+    // Same (project, id, routing) shape as manifest() and payload(), so all three routes read alike.
+    private Payload search(final String project, final String id, final String routing,
+                           final String query) throws IOException {
+        Path docArtifactDir = docArtifactDir(project, id, routing);
         if (docArtifactDir == null) {
             return Payload.notFound();
         }
-        StructureSearch.Hits hits = new StructureSearch(reader, docArtifactDir, extension).search(query);
+        StructureSearch.Hits hits = new StructureSearch(reader, docArtifactDir, MARKDOWN).search(query);
         return hits == null ? Payload.notFound() : PayloadFormatter.json(hits);
     }
 
--- a/datashare-app/src/main/java/org/icij/datashare/web/ArtifactResource.java
+++ b/datashare-app/src/main/java/org/icij/datashare/web/ArtifactResource.java
@@ -235,7 +235,9 @@ public class ArtifactResource {
     }
 
     // The extension a ?format= value asks for, defaulted, or null when it is not one this resource
-    // serves. Shared by the page route and the search route so their whitelists cannot drift.
+    // serves. The page route's whitelist alone: the search route takes no ?format= and counts the
+    // Markdown deliberately, since counting over the XHTML would report tag names, class names and
+    // link targets as occurrences of the user's term (see structureSearch).
     private static String structureExtension(final String format) {
         String extension = ofNullable(format).filter(value -> !value.isBlank()).orElse(MARKDOWN);
         return STRUCTURE_CONTENT_TYPES.containsKey(extension) ? extension : null;
--- a/datashare-app/src/main/java/org/icij/datashare/web/ArtifactResource.java
+++ b/datashare-app/src/main/java/org/icij/datashare/web/ArtifactResource.java
@@ -18,6 +18,7 @@ import org.icij.datashare.text.artifact.ArtifactReader;
 import org.icij.datashare.text.artifact.ArtifactType;
 import org.icij.datashare.text.artifact.FilesystemManifestRepository;
 import org.icij.datashare.text.artifact.ManifestEntry;
+import org.icij.datashare.text.artifact.StructureSearch;
 import org.icij.datashare.text.indexing.Indexer;
 import org.icij.datashare.text.indexing.elasticsearch.ArtifactPath;
 import org.icij.datashare.utils.DocumentSourceAccess;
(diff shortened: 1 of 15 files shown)


datashare-app/src/main/java/org/icij/datashare/web/ArtifactResource.java
```diff
@@ -196,6 +234,18 @@ private static Map<String, String> structureContentTypes() {
         return unmodifiableMap(contentTypes);
     }
 
+    // The extension a ?format= value asks for, defaulted, or null when it is not one this resource
+    // serves. Shared by the page route and the search route so their whitelists cannot drift.
```
caro3801: Search route hardcodes markdown and ignores `?format=` (L295), so this whitelist does not cover it.
question: is it on purpose? if yes is it possible to reword the comment to state that it's intentionnal?

datashare-index/src/main/java/org/icij/datashare/text/ContentOccurrences.java
```diff
@@ -0,0 +1,93 @@
...
+    private static final String[] MEMO = new String[MEMO_LIMIT];
+
+    private ContentOccurrences() {}
+
+    /**
+     * Occurrences of {@code query} in {@code content}, both folded first. Stepping by the raw query's
+     * length, as the script does, means matches may overlap when the fold expands the query (a
+     * ligature): {@code count("afffb", "ﬀ")} is 2, and the two share a character.
+     */
+    public static int count(String content, String query) {
```
caro3801, line 34: suggestion: `count()` folds `query` fresh on every call, and `StructureSearch.countPage()` now
calls it once per page with the same unchanging query.
 A thousand-page document currently reallocates the same fold a thousand times.

Could we add an overload that takes an already-folded query, so `StructureSearch.scan()` folds it once up front and reuses it across the whole scan?

```suggestion
    public static int count(String content, String query) {
        // On the folded query, which is what drives the loop: indexOf never returns -1 for an empty
        // needle, so an empty one spins forever rather than miscounting. Nothing folds to empty today,
        // so guarding the raw query would leave termination resting on that, one edit away.
        return count(content, fold(query), query.length());
    }

    /**
     * Same as {@link #count(String, String)}, but for a query already folded and its original
     * (unfolded) length — for a caller that folds once and reuses it across many calls, such as
     * {@link org.icij.datashare.text.artifact.StructureSearch} scanning a document page by page.
     */
    public static int count(String content, String foldedQuery, int queryLength) {
        if (foldedQuery.isEmpty()) {
            return 0;
        }
        return occurrences(fold(content), foldedQuery, queryLength);
    }

    // was: `static String fold(String input)` — widen to public so StructureSearch can fold once.
    public static String fold(String input) {
```

The suggested code has not been tested

datashare-index/src/main/java/org/icij/datashare/text/artifact/ArtifactReader.java
```diff
@@ -17,6 +19,12 @@
  *  servable, where a page lives, and what a missing payload means. */
 public class ArtifactReader {
     private static final Logger LOGGER = LoggerFactory.getLogger(ArtifactReader.class);
+
+    /** Beyond this a total is a corrupt manifest, not a long document. It matters because {@link
+     *  StructureSearch} turns a total written by another producer into a loop bound, and unbounded,
+     *  {@code Integer.MAX_VALUE} never terminates at all: the counter wraps to MIN_VALUE, still in range. */
+    public static final int MAX_SERVABLE_PAGES = 100_000;
```
caro3801: important: `MAX_SERVABLE_PAGES` is enforced here in `servableTotal()`, which every artifact route
goes through (manifest, single-page fetch, and structure search) - but the doc comment on the
constant says its only purpose is bounding `StructureSearch`'s scan loop from running away on a
corrupt total.

Could we move the cap into `StructureSearch.scan()` instead (clamp its own loop
bound to `min(total, MAX_SERVABLE_PAGES)`) and let `servableTotal()` accept any positive total? As
written, a legitimately huge document (a merged archive, a bulk-exported log) with more than
100,000 pages now 404s on every artifact route, not just search.

datashare-index/src/main/java/org/icij/datashare/text/artifact/ArtifactReader.java
```diff
@@ -68,16 +78,21 @@ public byte[] page(Path docArtifactDir, ArtifactType type, ManifestEntry entry,
...
-        if (!Files.isReadable(file)) {
-            // In range per the manifest but absent on disk: the two disagree, which is worth
-            // seeing in the logs rather than reshaping the advertised page count silently.
-            LOGGER.warn("manifest advertises {} page(s) for '{}' but {} is missing or unreadable", total, type.token(), file);
+        // Read and let it fail, rather than stat then read: on the shared artifactDir this is
+        // deployed on, each is a network round trip, and StructureSearch walks every page, so the
+        // pre-check doubled the round trips of a whole scan to learn what the read reports anyway.
+        try {
+            return Files.readAllBytes(file);
+        } catch (NoSuchFileException | AccessDeniedException unreadable) {
```
caro3801: crucial: This narrowed from the old `Files.isReadable()` pre-check to catching only
`NoSuchFileException`/`AccessDeniedException`

Could we catch the broader `IOException` here instead, so a directory-where-a-file-was-expected or a symlink-loop `FileSystemException` still reads as a clean 404 (as it did before) rather than propagating out of `page()`/`payload()` and surfacing as a 500 to `/artifacts/page/:id/:page` and `/artifacts/structure/:id/:page`?

```suggestion
        } catch (IOException unreadable) {
```

datashare-index/src/main/java/org/icij/datashare/text/artifact/StructureArtifact.java
```diff
@@ -42,24 +53,62 @@ public ArtifactType type() {
...
         return Map.of("pipeline", "tika", "version", TIKA_VERSION,
+                "ocr", ocr.images(), "ocrStrategy", ocr.pdfStrategy().name(),
                 "extract", BuildVersions.EXTRACT, "datashare", BuildVersions.DATASHARE);
     }
 
+    // What the INDEX stage applied to this document, so the pages hold the text the content field holds.
+    // getOcrParser() is set only when OCR actually ran for it, and it is one-way here as disableOcr() is
+    // in PageArtifact: the run can turn OCR off, it cannot turn it on for a document indexed without it.
+    private OcrSettings ocrSettings(Document document) {
+        return document.getOcrParser() == null ? OcrSettings.NONE : runOcrSettings();
```
caro3801, line 84: question: Could we store the OCR strategy used at index time in document metadata and have `StructureArtifact` read it later?

Right now it only knows whether OCR ran, then uses the current ARTIFACT run’s `--ocrStrategy`, which can make structure pages differ from the content searched by `/documents/searchContent`.

datashare-index/src/main/java/org/icij/datashare/text/artifact/StructureArtifact.java
```diff
@@ -42,24 +53,62 @@ public ArtifactType type() {
...
+    }
+
+    private OcrSettings runOcrSettings() {
```
caro3801, line 87: hint: `runOcrSettings()`'s OCR-enabled resolution duplicates `PageArtifact.ocrEnabled()` verbatim
Could we pull it into one shared helper so a future change to how `OCR_OPT` is parsed or defaulted doesn't have to be made twice?

datashare-index/src/main/java/org/icij/datashare/text/artifact/StructureSearch.java
```diff
@@ -0,0 +1,127 @@
...
+        this.scanBudget = scanBudget;
+    }
+
+    /** Per-page occurrence counts in page order, pages with none omitted, or null when this
+     *  document has nothing servable to search. Callers map null to 404 the way they map the
+     *  reader's own nulls, so "not found" keeps one definition across the artifact routes. */
+    public Hits search(String query) throws IOException {
+        ManifestEntry entry = reader.servableEntry(docArtifactDir, TYPE);
+        Integer total = entry == null ? null : reader.servableTotal(docArtifactDir, TYPE, entry);
+        // Without the formats probe, a document whose page-N.md files are gone would answer "no
```
caro3801: suggestion(perf): `search()`'s `formats()` probe stats page 1 to confirm the format exists, and then
`scan()`'s first loop iteration reads that same page 1 file again a few lines later.
This is the exact stat-then-read round trip the sibling `ArtifactReader.page()` change in this PR was written to eliminate, just reintroduced once per search request. 

Could we have the probe read page 1's bytes directly (instead of just checking readability) and hand them to `scan()` so the first iteration reuses them instead of re-reading?

datashare-index/src/main/java/org/icij/datashare/text/artifact/StructureSearch.java
```diff
@@ -0,0 +1,127 @@
...
+                continue;
+            }
+            scanned++;
+            if (count > 0) {
+                hits.add(new PageHits(page, count));
+            }
+        }
+        reportIncompleteScan(scanned, page - 1, total);
+        return new Hits(hits.stream().mapToInt(PageHits::count).sum(), total, scanned, hits);
+    }
```
caro3801, line 109: **suggestion:** `scan()` computes the total count with a second pass (`hits.stream().mapToInt(PageHits::count).sum()`) after the loop, even though each page's count is
already in hand at the `count > 0` check inside it. 

Could we accumulate it there instead and drop the extra traversal?

```suggestion
    private Hits scan(ManifestEntry entry, int total, String query) {
        List<PageHits> hits = new ArrayList<>();
        // Subtraction rather than a plain comparison, so a nanoTime wrap cannot end the scan early.
        long deadline = System.nanoTime() + scanBudget.toNanos();
        int scanned = 0;
        int matches = 0;
        int page = 1;
        for (; page <= total && System.nanoTime() - deadline < 0; page++) {
            Integer count = countPage(entry, page, query);
            if (count == null) {
                continue;
            }
            scanned++;
            if (count > 0) {
                matches += count;
                hits.add(new PageHits(page, count));
            }
        }
        reportIncompleteScan(scanned, page - 1, total);
        return new Hits(matches, total, scanned, hits);
    }
```

caro3801 (changes requested):
Great feature and solid implementation easy to follow through
There are the two blocking correctness issues before merge: 
* the page-count cap currently affects all artifact routes, 
* and narrowed read-error handling may turn some 404 cases into 500s. 

I also put a question about storing the OCR strategy used at index time to avoid future content/rendering mismatches. The rest is mostly cleanup, clarification, and small structure-search efficiency improvements.

datashare-index/src/main/java/org/icij/datashare/text/artifact/ArtifactReader.java
pirhoo: Moved in 414807ea2: `servableTotal()` now takes any positive total, and `StructureSearch` clamps its own loop to `min(total, MAX_SCANNED_PAGES)`. `pages` still reports the manifest total, so past the cap `scanned < pages` marks the count as a floor.

datashare-index/src/main/java/org/icij/datashare/text/artifact/ArtifactReader.java
pirhoo: Done in 33c911d50, with a test staging a directory where the page file belongs: a `FileSystemException`, so it escaped the narrower catch as a 500.

datashare-index/src/main/java/org/icij/datashare/text/artifact/StructureSearch.java
pirhoo: Fixed in aba5c506b, one step further: the probe is gone. `search()` counts page 1 through `ArtifactReader.page()` and seeds the scan, which starts at page 2. One consequence: under byte-range pagination a malformed range 1 now 404s the search, a scheme no structure artifact uses.

datashare-index/src/main/java/org/icij/datashare/text/artifact/StructureSearch.java
pirhoo: Done in aba5c506b, `matches` accumulates in the loop.

datashare-index/src/main/java/org/icij/datashare/text/artifact/StructureArtifact.java
pirhoo: Pulled into `Artifact.ocrEnabled(PropertiesProvider)` in 9719b952d, both call it. `--ocrStrategy` stays in `StructureArtifact`, since `PageArtifact` has no PDF strategy to resolve.

datashare-app/src/main/java/org/icij/datashare/web/ArtifactResource.java
pirhoo: Intentional, and the comment was wrong: `structureExtension()` has one caller, the page route. Reworded in 5f05df78a to say Markdown-only is deliberate.

datashare-index/src/main/java/org/icij/datashare/text/ContentOccurrences.java
pirhoo: Leaving this one. Folding the query is O(query length), folding a page is O(page size): over a thousand pages that is ~10k character operations against tens of millions for the content, so it would not be measurable. Not worth a second public entry point plus widening `fold()`.

datashare-index/src/main/java/org/icij/datashare/text/artifact/StructureArtifact.java
pirhoo: `--ocrStrategy` only reaches `PDFParserConfig`, so the drift needs a scanned PDF plus INDEX and ARTIFACT runs passing different strategies; a standalone image uses the `--ocr` boolean, already gated on `getOcrParser()`. This PR also puts `ocrStrategy` in the fingerprint, so a mismatch goes stale and re-extracts instead of sticking. Recording the index-time strategy means writing it from INDEX into document metadata, a datashare + extract-lib change for a case passing the same flag already covers.

datashare-index/src/main/java/org/icij/datashare/text/artifact/StructureArtifact.java
```diff
@@ -42,24 +52,62 @@ public ArtifactType type() {
...
+    // Unset or unparseable means NO_OCR, both as extract-lib resolves --ocrStrategy: a scanned PDF
+    // renders empty here exactly when the INDEX stage extracted nothing from it either.
+    private PDFParserConfig.OCR_STRATEGY pdfOcrStrategy() {
+        try {
+            return PDFParserConfig.OCR_STRATEGY.valueOf(
+                    propertiesProvider.get(OCR_STRATEGY_OPT).orElse("NO_OCR").toUpperCase(Locale.ROOT));
+        } catch (IllegalArgumentException unknown) {
+            return PDFParserConfig.OCR_STRATEGY.NO_OCR;
+        }
+    }
```
caro3801, line 105: hint: do you want to reuse the same enum parsing/mapping as in datashare cli and map it into Tika's so the two don't drift independently if a member is ever renamed?

```java
    private PDFParserConfig.OCR_STRATEGY pdfOcrStrategy() {
        try {
            OcrStrategy strategy = OcrStrategy.valueOf(
                    propertiesProvider.get(OCR_STRATEGY_OPT).orElse("NO_OCR").toUpperCase(Locale.ROOT));
            return PDFParserConfig.OCR_STRATEGY.valueOf(strategy.name());
        } catch (IllegalArgumentException unknown) {
            return PDFParserConfig.OCR_STRATEGY.NO_OCR;
        }
    }
```

datashare-index/src/main/java/org/icij/datashare/text/artifact/ArtifactReader.java
caro3801: suggestion: `slice()` does a `Files.isReadable()` presence check (stat) before opening the file, the exact stat-then-read pattern this same PR removed from the sibling filesystem-page branch a few lines up, with a comment calling it out as a doubled round trip on a shared/network `artifactDir`. 
Could we drop the pre-check here too and read-then-catch instead, the way `page()`'s filesystem branch now does?

remark(TOCTOU): this pattern can be sometimes racy, especially on network/shared filesystems. It’s often better to just try the read/open directly and handle IOException, because the file could change between the check and the read anyway.

datashare-index/src/main/java/org/icij/datashare/text/artifact/StructureSearch.java
caro3801: thanks for making the changes. It introduced an asymmetric semantic behavior between search and scan now .

hint(low impact): `search()` now counts page 1 directly and treats a null result as "no structure here at all," returning 404 for the whole document. Every later page read failure just decrements `scanned` and lets the scan continue, so page 1 alone can abort the whole search and hide matches on pages 2..N. 

One concrete edge case is byte-range pagination with a malformed first range: page 1 returns `null`, but later ranges may still be readable. 

Could `search()` fall through to `scan()` on a null first page too, treating it like any other unreadable page and letting `scanned` / `reportIncompleteScan` carry the degraded-result signal? 

This is probably low-impact today since structure artifacts do not use byte-range pagination, but it would keep the failure semantics consistent.

datashare-index/src/main/java/org/icij/datashare/text/artifact/ArtifactReader.java
pirhoo: Done in 05f176ba0, and the `Files.size()` bound check went too: `readFully` raising `EOFException` says the same thing without a second stat. `slice()` now catches `IOException` and returns null, so the byte-range branch 404s where the filesystem one does.

datashare-index/src/main/java/org/icij/datashare/text/artifact/StructureArtifact.java
pirhoo: Done in e4763f4a6. Both enums carry the same four members today, so no behavior change. Note it does not catch a rename at compile time, it routes it into the `NO_OCR` fallback; an exhaustive `switch` would, if you prefer that.

datashare-index/src/main/java/org/icij/datashare/text/artifact/StructureSearch.java
pirhoo: Fixed in e3216800a: the scan decides, `search()` 404s only when `scanned == 0`, so page 1 is read like any other page. Test: `test_an_unreadable_first_page_does_not_hide_the_pages_after_it`.

datashare-index/src/main/java/org/icij/datashare/text/artifact/StructureArtifact.java
caro3801: no need thanks!

PR #2322, opened 2026-08-13
refactor(temporal): extract workflow registration into WorkflowRegistry and TemporalWorkers

lschim:
## Why

The discovery of Workflows was hardcoded on detecting single activity workflows annotated with @WorkflowInterface, the PR offers a way to register different types of Workflows.

Also, `TemporalInterlocutor` carried three responsibilities: talking to Temporal, knowing the naming convention that maps a workflow interface to its implementation and activity, and owning the lifecycle of the workers serving them. This PR extracts the last two; the interlocutor is a Temporal client again.

It is also the occasion to widen the possibilities of queue management and ease of starting workers.

## The two extracted classes

- **`WorkflowRegistry`** : what should be served, and on which queue. Manual registration, discovery by naming convention, read views. **No Temporal dependency at all**, no client.
- **`TemporalWorkers.start(client, registry, listeningQueues, options)`** — starts the workers and returns the `Closeable`. The only part that knows about `WorkerFactory`.

## The substantive change: when the routing decision is made

Discovery used to drop every workflow that did not belong to the worker's group (`makeWorkflowFilter(routingStrategy, group)`), so a process could only ever serve one queue.

The registry now registers everything it finds, keyed by the queue the routing strategy resolves, and `TemporalWorkers.start` picks which of those queues to poll. A distributed deployment can serve a subset; `CommonMode` serves everything it discovered. This is what unblocks the fan-out.

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/mode/CommonMode.java
+++ b/datashare-app/src/main/java/org/icij/datashare/mode/CommonMode.java
@@ -565,7 +565,7 @@ public abstract class CommonMode extends AbstractModule implements Closeable {
                     Utils.getRoutingStrategy(propertiesProvider), new Group(TaskGroupType.Java));
             // this process serves everything it discovered; a distributed deployment would pass a subset of the queues
             addCloseable(TemporalWorkers.start(temporal.getClient(), registry, registry.registeredQueues(),
-                    new TemporalWorkerOptions(getTaskWorkersNb(), getTaskWorkersNb())));
+                    new TemporalWorkerOptions(getTaskWorkersNb())));
         }
         return executorService;
     }
--- a/datashare-app/src/main/java/org/icij/datashare/mode/CommonMode.java
+++ b/datashare-app/src/main/java/org/icij/datashare/mode/CommonMode.java
@@ -21,6 +21,9 @@ import org.icij.datashare.asyncsearch.MemoryAsyncSearchStore;
 import org.icij.datashare.asyncsearch.RedisAsyncSearchStore;
 import org.icij.datashare.asynctasks.*;
 import org.icij.datashare.asynctasks.temporal.TemporalInterlocutor;
+import org.icij.datashare.asynctasks.temporal.TemporalWorkerOptions;
+import org.icij.datashare.asynctasks.temporal.TemporalWorkers;
+import org.icij.datashare.asynctasks.temporal.WorkflowRegistry;
 import org.icij.datashare.batch.BatchSearchRepository;
 import org.icij.datashare.cli.Mode;
 import org.icij.datashare.cli.QueueType;
@@ -554,12 +557,15 @@ public abstract class CommonMode extends AbstractModule implements Closeable {
     private ExecutorService runTemporalWorkers() {
         if (getTaskWorkersNb() > 0) {
             TemporalInterlocutor temporal = get(TemporalInterlocutor.class);
-            executorService.submit(() -> {
-                closeables.add(temporal.discoverWorkflows(getTaskWorkersNb(), get(DatashareTaskFactory.class),
-                                get(TaskRepository.class),
-                                Utils.getRoutingStrategy(propertiesProvider),
-                                new Group(TaskGroupType.Java)));
-            });
+            DatashareTaskFactory taskFactory = get(DatashareTaskFactory.class);
+            TaskRepository taskRepository = get(TaskRepository.class);
+            WorkflowRegistry registry = new WorkflowRegistry();
+            registry.discoverWorkflows("org.icij.datashare.tasks",
+                    activityClass -> WorkflowRegistry.activityFactoryForSingleActivitiesWorkflow(activityClass, taskFactory, temporal.getClient(), taskRepository,  1d).getThrows(),
+                    Utils.getRoutingStrategy(propertiesProvider), new Group(TaskGroupType.Java));
+            // this process serves everything it discovered; a distributed deployment would pass a subset of the queues
+            addCloseable(TemporalWorkers.start(temporal.getClient(), registry, registry.registeredQueues(),
+                    new TemporalWorkerOptions(getTaskWorkersNb(), getTaskWorkersNb())));
         }
         return executorService;
     }
--- a/datashare-tasks/src/main/java/org/icij/datashare/asynctasks/TaskManagerTemporal.java
+++ b/datashare-tasks/src/main/java/org/icij/datashare/asynctasks/TaskManagerTemporal.java
@@ -39,6 +39,7 @@ public class TaskManagerTemporal implements TaskManager {
     @Override
     public <V extends Serializable> String startTask(Task<V> taskView, Group group) throws IOException, TaskAlreadyExists {
         String taskId = taskView.id;
+        logger.atInfo().log("Starting task {} of type {}", taskView.id, taskView.type);
         try {
             taskRepository.insert(taskView, group);
             temporal.createWorkflow(taskId, taskView.name, resolveWfTaskQueue(taskView.name ,group),
--- a/datashare-tasks/src/main/java/org/icij/datashare/asynctasks/temporal/TemporalWorkers.java
+++ b/datashare-tasks/src/main/java/org/icij/datashare/asynctasks/temporal/TemporalWorkers.java
@@ -1,0 +1,84 @@
+package org.icij.datashare.asynctasks.temporal;
+
+
+import io.temporal.client.WorkflowClient;
+import io.temporal.worker.Worker;
+import io.temporal.worker.WorkerFactory;
+import io.temporal.worker.WorkerOptions;
+import io.temporal.worker.WorkflowImplementationOptions;
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+
+/**
+ * Starts the Temporal workers serving what a {@link WorkflowRegistry} holds.
+ */
+public class TemporalWorkers {
+    private static final WorkflowImplementationOptions WF_IMPLEMENTATION_DEFAULT_OPTIONS = WorkflowImplementationOptions.newBuilder()
+            .setFailWorkflowExceptionTypes(Error.class) // Unregistered workflows
+            .build();
+
+    private TemporalWorkers() {
+    }
+
+    /**
+     * Starts a worker for each of the given queues, serving the workflows and activities the registry holds for it.
+     *
+     * @param client the client the workers poll Temporal with
+     * @param registry what to serve, keyed by task queue
+     * @param listeningQueues the queues the workers will poll; registrations bound to any other queue are ignored
+     * @param options the concurrency limits applied to every started worker
+     * @return a handle closing the started worker factory
+     */
+    public static Closeable start(
+            WorkflowClient client, WorkflowRegistry registry,
+            Collection<String> listeningQueues, TemporalWorkerOptions options) {
+        Objects.requireNonNull(client, "client");
+        Objects.requireNonNull(registry, "registry");
+
+        WorkerFactory workerFactory = WorkerFactory.newInstance(client);
+        WorkerOptions workerOptions = WorkerOptions.newBuilder()
+                .setMaxConcurrentWorkflowTaskExecutionSize(options.maxConcurrentWorkflowSize())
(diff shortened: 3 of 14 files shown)


datashare-tasks/src/main/java/org/icij/datashare/asynctasks/temporal/TemporalWorkers.java
```diff
@@ -0,0 +1,85 @@
...
+            Set<Class<?>> workflowClasses = registry.registeredWorkflows(queue);
+            if (!workflowClasses.isEmpty()) {
+                workers.computeIfAbsent(queue, q -> workerFactory.newWorker(q, workerOptions))
+                        .registerWorkflowImplementationTypes(WF_IMPLEMENTATION_DEFAULT_OPTIONS, workflowClasses.toArray(Class<?>[]::new));
+            }
+            Set<Object> activities = registry.registeredActivities(queue);
+            if (!activities.isEmpty()) {
+                workers.computeIfAbsent(queue, q -> workerFactory.newWorker(q, workerOptions))
+                        .registerActivitiesImplementations(activities.toArray());
+            }
```
mvanzalu: suggestion: would it be worth logging what we start on each queue here? Something like:
```
logger.info("starting temporal worker on queue <{}> with {} workflow(s) and {} activity(ies)", queue, workflowClasses.size(), activities.size());
```
I'm not sure how many queues we end up with in practice, if it's more than a handful, a single line after the loop would be preferable

mvanzalu:
LGTM !

datashare-tasks/src/main/java/org/icij/datashare/asynctasks/temporal/WorkflowRegistry.java
```diff
@@ -0,0 +1,219 @@
...
+
+        Reflections reflections = new Reflections(packageName);
+        // We rely on naming convention rather than on inspection, that's OK as code is generated
+        reflections.getTypesAnnotatedWith(WorkflowInterface.class)
+                .stream()
+                .filter(workflowInterfaceFilter())
+                .forEach(c -> {
+                    String taskQueue = resolveWfTaskQueue(routingStrategy, buildWorkflowKeyFrom(c), group);
+                    registerWorkflow(workflowImplementation(c), taskQueue);
+                    registerActivity(activityInstance(c, activityInstantiator), taskQueue);
```
pirhoo, line 119: crucial: the activity lands on the routing-derived queue, but every generated `*WorkflowImpl` pins its stub to `ActivityOpts.taskQueue()`, default `"default-java"` and never overridden. Under GROUP or NAME the activity is scheduled on a queue nobody polls and the task sits RUNNING forever; UNIQUE only works because `WORKFLOWS_DEFAULT` happens to match. Could `TemporalWorkflowGenerator` emit
`Workflow.getInfo().getTaskQueue()` instead of baking the literal in?

datashare-tasks/src/main/java/org/icij/datashare/asynctasks/temporal/TemporalWorkers.java
```diff
@@ -0,0 +1,85 @@
...
+    public static Closeable start(
+            WorkflowClient client, WorkflowRegistry registry,
+            Collection<String> listeningQueues, TemporalWorkerOptions options) {
+        Objects.requireNonNull(client, "client");
+        Objects.requireNonNull(registry, "registry");
+
+        WorkerFactory workerFactory = WorkerFactory.newInstance(client);
+        WorkerOptions workerOptions = WorkerOptions.newBuilder()
+                .setMaxConcurrentWorkflowTaskExecutionSize(MAX_CONCURRENT_WORKFLOW_TASK_EXECUTION)
+                .setMaxConcurrentActivityExecutionSize(options.maxConcurrentActivitySize())
```
pirhoo, line 49: important: `workerOptions` used to be discarded, since the workflow line created the worker first without options, so activities ran at the SDK default of 200. It applies now, so the default `taskWorkers=1` lets a long `IndexTask` serialize every other task on the queue. Is sizing activity concurrency by `taskWorkers`
the intent here?

datashare-tasks/src/main/java/org/icij/datashare/asynctasks/temporal/TemporalWorkers.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.Set;
+
+/**
+ * Starts the Temporal workers serving what a {@link WorkflowRegistry} holds.
+ */
+public class TemporalWorkers {
+    private static final WorkflowImplementationOptions WF_IMPLEMENTATION_DEFAULT_OPTIONS = WorkflowImplementationOptions.newBuilder()
+            .setFailWorkflowExceptionTypes(Error.class) // Unregistered workflows
+            .build();
+    private static final int MAX_CONCURRENT_WORKFLOW_TASK_EXECUTION = 200; //This value is hardcoded for now as workflow tasks are not very costly but it is necessary to keep it high enough
```
pirhoo, line 26: question: 200 is per worker, but all workers share one `WorkerFactory` capped at 600 workflow threads. NAME routing would ask for 16 x 200 against that. Should the cap be divided across workers, or raised on the
factory?

datashare-tasks/src/main/java/org/icij/datashare/asynctasks/temporal/WorkflowRegistry.java
```diff
@@ -0,0 +1,219 @@
...
+        workflowsByQueue.computeIfAbsent(queue, q -> new LinkedHashSet<>()).add(workflowClass);
+    }
+
+    /**
+     * Register an activity instance to be served on the given queue name.
+     */
+    public void registerActivity(Object activity, String queue) {
+        Objects.requireNonNull(activity, "activity");
+        Objects.requireNonNull(queue, "queue");
+        activitiesByQueue.computeIfAbsent(queue, q -> new LinkedHashSet<>()).add(activity);
```
pirhoo, line 66: suggestion: workflows dedupe by `Class`, activities by instance identity, and `TemporalActivityImpl` has no `equals`. Discovery run twice would then register the same activity class twice and die with `TypeAlreadyRegisteredException`, the case `TemporalWorkers.start` says is handled:

```suggestion
        Set<Object> activities = activitiesByQueue.computeIfAbsent(queue, q -> new LinkedHashSet<>());
        if (activities.stream().noneMatch(a -> a.getClass().equals(activity.getClass()))) {
            activities.add(activity);
        }
```

datashare-app/src/main/java/org/icij/datashare/mode/CommonMode.java
```diff
@@ -554,12 +557,15 @@ private ExecutorService runTaskWorkerLoop() {
...
-                                get(TaskRepository.class),
-                                Utils.getRoutingStrategy(propertiesProvider),
-                                new Group(TaskGroupType.Java)));
-            });
+            DatashareTaskFactory taskFactory = get(DatashareTaskFactory.class);
+            TaskRepository taskRepository = get(TaskRepository.class);
+            WorkflowRegistry registry = new WorkflowRegistry();
+            registry.discoverWorkflows("org.icij.datashare.tasks",
+                    activityClass -> WorkflowRegistry.activityFactoryForSingleActivitiesWorkflow(activityClass, taskFactory, temporal.getClient(), taskRepository,  1d).getThrows(),
+                    Utils.getRoutingStrategy(propertiesProvider), new Group(TaskGroupType.Java));
```
pirhoo, line 565: suggestion: running discovery on the calling thread is the right call, but `WebApp` calls `runWorkers()` after the port is bound, so a `WorkflowRegistrationException` leaves the server up and skips `requeueDatabaseBatchSearches`. Could the worker startup move ahead of `.start(port)`?

datashare-tasks/src/main/java/org/icij/datashare/asynctasks/TaskManagerTemporal.java
```diff
@@ -39,6 +39,7 @@ public TaskManagerTemporal(TemporalInterlocutor temporal, TaskRepository taskRep
     @Override
     public <V extends Serializable> String startTask(Task<V> taskView, Group group) throws IOException, TaskAlreadyExists {
         String taskId = taskView.id;
+        logger.atInfo().log("Starting task {} of type {}", taskView.id, taskView.type);
```
pirhoo, line 42: suggestion: this INFO line looks unrelated to the split, and sitting before `insert` it also logs a start for tasks that immediately throw `TaskAlreadyExists`. Drop it, or move it after the insert succeeds:

```suggestion
```

datashare-tasks/src/main/java/org/icij/datashare/asynctasks/temporal/TemporalWorkers.java
```diff
@@ -0,0 +1,85 @@
...
+            if (!workflowClasses.isEmpty()) {
+                workers.computeIfAbsent(queue, q -> workerFactory.newWorker(q, workerOptions))
+                        .registerWorkflowImplementationTypes(WF_IMPLEMENTATION_DEFAULT_OPTIONS, workflowClasses.toArray(Class<?>[]::new));
+            }
+            Set<Object> activities = registry.registeredActivities(queue);
+            if (!activities.isEmpty()) {
+                workers.computeIfAbsent(queue, q -> workerFactory.newWorker(q, workerOptions))
+                        .registerActivitiesImplementations(activities.toArray());
+            }
+        });
```
pirhoo, line 65: hint: a listening queue with nothing registered creates no worker and logs nothing, so `start` returns a
healthy-looking `Closeable` for a process that polls an empty set. A warning there would make a mistyped
queue name visible at startup.

pirhoo (changes requested):
Really nice split! The one thing holding this up is that the queue the registry binds activities to does not match the queue the generated workflow stubs call, so the fan-out this unblocks would hang before it runs.

ClemDoum (approved):

datashare-tasks/src/main/java/org/icij/datashare/asynctasks/temporal/WorkflowRegistry.java
lschim: Good catch! Currently nothing works outside of the UNIQUE naming, but I can start to handle the other strategies too.

lschim: Fixed

datashare-tasks/src/main/java/org/icij/datashare/asynctasks/temporal/TemporalWorkers.java
lschim: Yes, that's right. I added back a configuration value that by default is set to 200. It could become a user configurable entry when needed

datashare-tasks/src/main/java/org/icij/datashare/asynctasks/temporal/WorkflowRegistry.java
lschim: Done !

datashare-app/src/main/java/org/icij/datashare/mode/CommonMode.java
lschim: Done !

datashare-tasks/src/main/java/org/icij/datashare/asynctasks/TaskManagerTemporal.java
lschim: Done

datashare-tasks/src/main/java/org/icij/datashare/asynctasks/temporal/TemporalWorkers.java
lschim: I made the change to reduce to 20 per worker, but compute the value at the factory level to be workers * maxValuePerWorker. The value 20 might seem a bit low, but it could be increased in the future, and I think we will also remove completely routing strategies and decide by ourselves what should be the queue name of each Task, Workflow, worker etc... to be able to do the optimization we need

datashare-tasks/src/main/java/org/icij/datashare/asynctasks/temporal/TemporalWorkers.java
lschim: I don't think it is a good idea to implement that warning, because the idea will be to have different workers (in term of processes this time, so multiple executions of ./datashare --mode WORKER) which will poll different queues. And scaling will be done on those queues, so the result will be that not every worker will listen to all the queues, and the warning you suggest would add erroneous noise

pirhoo (approved):
Thanks for the changes!

Issue #2343, opened 2026-08-19
fix: bundled commons-beanutils, protobuf-java, and handlebars have known CVEs in 21.18.0


neoReuters:
## Summary

The `icij/datashare:21.18.0` image bundles three libraries with known CVEs, and one library with unresolvable licence metadata. The affected versions are `commons-beanutils` 1.8.3, `protobuf-java` 2.5.0, `handlebars` 4.2.1, and `org.webjars:coffee-script` 1.11.0.

Other bundled libraries in the same image are current. This suggests the four versions above are set by a transitive dependency rather than by choice.

## Affected artifact

`icij/datashare:21.18.0`, file `/home/datashare/lib/datashare-dist-21.18.0-all.jar`.

## High severity

| Library | Version | CVE | Fixed in |
| --- | --- | --- | --- |
| `commons-beanutils:commons-beanutils` | 1.8.3 | [CVE-2025-48734](https://nvd.nist.gov/vuln/detail/CVE-2025-48734) — unsafe reflection | 1.11.0 |
| `commons-beanutils:commons-beanutils` | 1.8.3 | [CVE-2014-0114](https://nvd.nist.gov/vuln/detail/CVE-2014-0114) — unsafe reflection | 1.9.4 |
| `com.github.jknack:handlebars` | 4.2.1 | [CVE-2026-55760](https://nvd.nist.gov/vuln/detail/CVE-2026-55760) — directory traversal | 4.5.2 |
| `com.google.protobuf:protobuf-java` | 2.5.0 | [CVE-2024-7254](https://nvd.nist.gov/vuln/detail/CVE-2024-7254) — stack-based buffer overflow | 3.25.5 |

## Medium severity

| Library | Version | CVE | Fixed in |
| --- | --- | --- | --- |
| `commons-beanutils:commons-beanutils` | 1.8.3 | [CVE-2019-10086](https://nvd.nist.gov/vuln/detail/CVE-2019-10086) — deserialization of untrusted data | 1.9.4 |
| `com.google.protobuf:protobuf-java` | 2.5.0 | [CVE-2015-5237](https://nvd.nist.gov/vuln/detail/CVE-2015-5237) — integer overflow | 3.4.0 |
| `com.google.protobuf:protobuf-java` | 2.5.0 | [CVE-2021-22569](https://nvd.nist.gov/vuln/detail/CVE-2021-22569) — denial of service | 3.16.1 |
[...]

PR #2344, opened 2026-08-19
fix(deps): pin commons-beanutils and handlebars up for known CVEs

neoReuters:
Fixes part of #2343.

Both libraries arrive transitively at versions with published CVEs:

| Library | Current | CVEs | Pinned to |
| --- | --- | --- | --- |
| `commons-beanutils` | 1.8.3 | CVE-2014-0114, CVE-2019-10086, CVE-2025-48734 | 1.9.4 |
| `com.github.jknack:handlebars` | 4.2.1 | CVE-2026-55760 | 4.5.2 |

The pins go in `dependencyManagement`, so they apply at any tree depth. This follows the existing pattern for the ASM family.

`commons-beanutils.version` already existed as a property, but no `pom.xml` referenced it. This reuses it rather than adding a second one.

## Not included

- `protobuf-java` 2.5.0 has four CVEs, but the fix is 3.25.5. The generated-code API changed between 2.x and 3.x, so that upgrade needs its own change.
- `org.webjars:coffee-script` 1.11.0 reports as an unknown licence. The licence string is wrong in the artifact metadata, so no version pin corrects it.

## What I verified

- The parent POM validates and installs.
- Both properties resolve to the intended versions.
- A resolved dependency tree shows `handlebars:4.5.2` with `commons-beanutils:1.9.4` beneath it.

I could not run the full reactor. `datashare-db` runs jOOQ code generation, which needs a live PostgreSQL connection, so the build stops there in a plain checkout. Your CI covers that path.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -116,7 +116,11 @@
              "IllegalArgumentException: Unsupported api 589824" (ASM9). ASM 9.x is backward compatible
              with Nashorn's ASM7-level usage, so pinning up is safe. -->
         <asm.version>9.9.1</asm.version>
-        <commons-beanutils.version>1.7.0</commons-beanutils.version>
+        <!-- Pinned up for CVE-2014-0114, CVE-2019-10086 and CVE-2025-48734. The image currently
+             ships 1.8.3 transitively; 1.9.4 is API compatible with it. -->
+        <commons-beanutils.version>1.9.4</commons-beanutils.version>
+        <!-- Pinned up for CVE-2026-55760 (directory traversal). The image currently ships 4.2.1. -->
+        <handlebars.version>4.5.2</handlebars.version>
         <jna.version>5.19.1</jna.version>
         <bouncycastle.version>1.70</bouncycastle.version>
         <httpcomponents.version>4.5.13</httpcomponents.version>
@@ -247,6 +251,19 @@
                 <groupId>org.ow2.asm</groupId>
                 <artifactId>asm-util</artifactId>
                 <version>${asm.version}</version>
+            </dependency>
+
+            <!-- commons-beanutils and handlebars pinned up for the CVEs noted beside their
+                 version properties. Both arrive transitively, so pinning here applies at any depth. -->
+            <dependency>
+                <groupId>commons-beanutils</groupId>
+                <artifactId>commons-beanutils</artifactId>
+                <version>${commons-beanutils.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>com.github.jknack</groupId>
+                <artifactId>handlebars</artifactId>
+                <version>${handlebars.version}</version>
             </dependency>
 
             <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -117,8 +117,9 @@
              with Nashorn's ASM7-level usage, so pinning up is safe. -->
         <asm.version>9.9.1</asm.version>
         <!-- Pinned up for CVE-2014-0114, CVE-2019-10086 and CVE-2025-48734. The image currently
-             ships 1.8.3 transitively; 1.9.4 is API compatible with it. -->
-        <commons-beanutils.version>1.9.4</commons-beanutils.version>
+             ships 1.8.3 transitively; 1.11.0 is API compatible with it. 1.9.4 fixes only the first
+             two: GHSA-wxr5-93ph-8wr9 records the CVE-2025-48734 fix at 1.11.0. -->
+        <commons-beanutils.version>1.11.0</commons-beanutils.version>
         <!-- Pinned up for CVE-2026-55760 (directory traversal). The image currently ships 4.2.1. -->
         <handlebars.version>4.5.2</handlebars.version>
         <jna.version>5.19.1</jna.version>
--- a/datashare-app/pom.xml
+++ b/datashare-app/pom.xml
@@ -13,7 +13,16 @@
     <artifactId>datashare-app</artifactId>
     <name>datashare-app</name>
 
+    <!-- central is declared first on purpose. JitPack synthesizes a POM for any coordinate it can
+         resolve from a GitHub tag, and for com.github.jknack:handlebars.java it returns one without
+         <packaging>pom</packaging>. Maven then rejects it as a parent, drops handlebars' declared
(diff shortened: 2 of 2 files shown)


pom.xml
```diff
@@ -116,7 +116,11 @@
              "IllegalArgumentException: Unsupported api 589824" (ASM9). ASM 9.x is backward compatible
              with Nashorn's ASM7-level usage, so pinning up is safe. -->
         <asm.version>9.9.1</asm.version>
-        <commons-beanutils.version>1.7.0</commons-beanutils.version>
+        <!-- Pinned up for CVE-2014-0114, CVE-2019-10086 and CVE-2025-48734. The image currently
+             ships 1.8.3 transitively; 1.9.4 is API compatible with it. -->
+        <commons-beanutils.version>1.9.4</commons-beanutils.version>
+        <!-- Pinned up for CVE-2026-55760 (directory traversal). The image currently ships 4.2.1. -->
+        <handlebars.version>4.5.2</handlebars.version>
```
pirhoo, line 124: important: pinning handlebars to 4.5.2 makes Maven resolve its parent POM from the jitpack.io repo declared in `datashare-app/pom.xml`, and JitPack answers with a 2 kB synthesized POM that has no `<packaging>pom</packaging>`. Maven then logs `The POM for com.github.jknack:handlebars:jar:4.5.2 is invalid, transitive dependencies (if any) will not be available` and drops all of handlebars' declared deps, so `commons-text` lands at 1.10.0 from webjars-locator instead of the 1.15.0 handlebars asks for. 4.2.1 escaped this because JitPack 404s on it and it fell through to Central. Could we declare `central` ahead of `jitpack.io` in the `<repositories>` block of `datashare-app/pom.xml`? With that in place the warning is gone and `commons-text:1.15.0` resolves under handlebars.

```xml
<repositories>
    <repository>
        <id>central</id>
        <url>https://repo.maven.apache.org/maven2</url>
    </repository>
    <repository>
        <id>jitpack.io</id>
        <url>https://jitpack.io</url>
    </repository>
</repositories>
```

pom.xml
```diff
@@ -116,7 +116,11 @@
              "IllegalArgumentException: Unsupported api 589824" (ASM9). ASM 9.x is backward compatible
              with Nashorn's ASM7-level usage, so pinning up is safe. -->
         <asm.version>9.9.1</asm.version>
-        <commons-beanutils.version>1.7.0</commons-beanutils.version>
+        <!-- Pinned up for CVE-2014-0114, CVE-2019-10086 and CVE-2025-48734. The image currently
+             ships 1.8.3 transitively; 1.9.4 is API compatible with it. -->
+        <commons-beanutils.version>1.9.4</commons-beanutils.version>
```
pirhoo: important: looks like 1.9.4 does not close CVE-2025-48734. GHSA-wxr5-93ph-8wr9 records the fix at 1.11.0, so a rescan of the image will still flag it and the ticket stays partly open. The other two CVEs are genuinely fixed in 1.9.4. The only consumer is `less4j:1.17.2`, whose bytecode references just `PropertyUtils.getProperty`, `setProperty` and `getPropertyDescriptors`, all still present in 1.11.0, so bumping looks safe. Worth a quick check that less4j never resolves a `class` or `declaringClass` property, since suppressing those is what 1.11.0 changed.

```suggestion
        <!-- Pinned up for CVE-2014-0114, CVE-2019-10086 and CVE-2025-48734. The image currently
             ships 1.8.3 transitively; 1.11.0 is API compatible with it. -->
        <commons-beanutils.version>1.11.0</commons-beanutils.version>
```

pirhoo (changes requested):
Thanks for chasing these CVEs down. Two things to sort out before this lands.

pom.xml
neoReuters: Done, bumped to 1.11.0. Same coordinates, so it is a plain version change.

On the `class` / `declaringClass` question: `ASTManipulator` is the only class in less4j that references beanutils, and `replace()` is the only method that resolves properties. It enumerates the descriptors of the parent's class, keeps the ones `isAstProperty` accepts (a read method that is not annotated `@NotAstProperty`), then reads a property only if one of two gates passes:

- `propertyType.isInstance(replacementNode)`
- `propertyType.isAssignableFrom(List.class)`

The `class` property has type `java.lang.Class`. An `ASTCssNode` is not a `Class`, and `List` is not assignable to `Class`, so both gates are false and the property is never passed to `PropertyUtils.getProperty`. `ASTCssNode` has no `declaringClass` property.

Also checked it functionally. Compiled the same LESS source — nesting, a mixin, a variable, an arithmetic operation and `:extend` — against all three versions:

| commons-beanutils | Result | Output SHA | Warnings |
| --- | --- | --- | --- |
| 1.8.3 (currently shipped) | ok | `c03b49bc1c4f` | 1 |
| 1.9.4 | ok | `c03b49bc1c4f` | 1 |
| 1.11.0 | ok | `c03b49bc1c4f` | 1 |

Byte-identical output in all three.

pom.xml
neoReuters: Done, `central` now precedes `jitpack.io`.

Confirmed the diagnosis on a machine with no `settings.xml` at all, so the repository ordering in `datashare-app/pom.xml` is the only factor:

```
[WARNING] The POM for com.github.jknack:handlebars:jar:4.5.2 is invalid, transitive dependencies (if any) will not be available
[ERROR] Invalid packaging for parent POM com.github.jknack:handlebars.java:4.5.2, must be "pom" but is "jar"
   \- org.apache.commons:commons-text:jar:1.10.0:compile
```

One wrinkle worth recording, because it made the fix look like it had not worked: Maven caches JitPack's POM, so the warning survives the change. `~/.m2/repository/com/github/jknack/handlebars.java/4.5.2/` has to be purged first. Its `_remote.repositories` lists both repositories, and the cached file is JitPack's 2 kB synthesized POM.

After purging, with `central` first:

```
Downloaded from central: .../handlebars.java-4.5.2.pom (16 kB)
   \- org.apache.commons:commons-text:jar:1.15.0:compile
```

Warning gone and `commons-text` at 1.15.0, as you predicted.

pirhoo (approved):
Thanks for the changes!

PR #2347, opened 2026-08-20
fix(redis): honour the rediss scheme when building the pool

### LINKED ISSUE #2346: bug: 21.x cannot connect to a TLS Redis, so every session lookup returns 500
## Summary

21.18.0 cannot connect to a Redis that requires TLS, so every session lookup returns 500.

`RedisPoolFactory` builds the pool with `JedisPool(GenericObjectPoolConfig, URI)`. In the bundled Jedis 2.9.0 the URI constructors pass `JedisFactory` a hardcoded `ssl=false` and never read the scheme, so a `rediss://` address connects in plaintext. `PING` never returns, `testOnBorrow` validation fails, and `borrowObject` throws.

20.13.1 called `new JedisPool(String)`, which does read the scheme, so it works against the same Redis with the same bundled Jedis.

Reading the `ssl` field Jedis stores, from the same URL:

| Constructor | `ssl` | `PING` |
| --- | --- | --- |
| `JedisPool(config, URI)` — 21.18.0 | `false` | fails |
| `JedisPool(String)` — 20.13.1 | `true` | `PONG` |

```
JedisException: Could not get a resource from the pool
	at org.icij.datashare.session.RedisSessionIdStore.getLogin(RedisSessionIdStore.java:46)
Caused by: java.util.NoSuchElementException: Unable to validate object
```

No option compensates: the only Redis options are `redisAddress` and `redisPoolSize`, so TLS can be expressed only through the scheme.

Affects `RedisSessionIdStore`, `UsersInRedis` and `UsersIdProviderRedisCache`. The queue and message bus use a different path, so workers keep running while the web server returns 500.

## Reproduce

Against a Redis with TLS required, using the JAR from the image:

```java
var pp = new PropertiesProvider(Map.of(
    "redisAddress", "rediss://<host>:6379", "sessionTtlSeconds", "43200"));
new RedisSessionIdStore(pp).getLogin("any-key");
```

Throws on 21.18.0. Replacing the pool construction with `new JedisPool("rediss://<host>:6379")` returns normally against the same server.

## Fix

#2347.

## Also

`createPool` never reads `redisPoolSize`, so a configured pool size has no effect.

neoReuters:
Fixes #2346.

`JedisPool`'s URI constructors pass `JedisFactory` a hardcoded `ssl=false` and never read the scheme, so a `rediss://` address connects in plaintext and every command fails. `JedisPool(String)` does read the scheme but takes no pool config, which is what this factory exists to set, so this passes host and port with `ssl` explicit.

Verified against a TLS-required Redis, driving an unmodified `RedisSessionIdStore` with only this class replaced:

```
get(absent)     -> null
put then get    -> probe-user
remove then get -> null
```

`ssl` resolves to `false` for `redis://`, so plaintext deployments are unaffected. A missing port falls back to `Protocol.DEFAULT_PORT`, and password and database index still come from `JedisURIHelper`.

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/session/RedisPoolFactory.java
+++ b/datashare-app/src/main/java/org/icij/datashare/session/RedisPoolFactory.java
@@ -3,6 +3,8 @@ package org.icij.datashare.session;
 import org.icij.datashare.PropertiesProvider;
 import redis.clients.jedis.JedisPool;
 import redis.clients.jedis.JedisPoolConfig;
+import redis.clients.jedis.Protocol;
+import redis.clients.util.JedisURIHelper;
 
 import java.net.URI;
 
@@ -11,6 +13,7 @@ import static org.icij.datashare.cli.DatashareCliOptions.REDIS_ADDRESS_OPT;
 
 final class RedisPoolFactory {
     private static final long EVICTION_RUN_INTERVAL_MILLIS = 30_000;
+    private static final String SSL_SCHEME = "rediss";
 
     private RedisPoolFactory() {}
 
@@ -25,6 +28,15 @@ final class RedisPoolFactory {
         poolConfig.setTestWhileIdle(true);
         poolConfig.setTimeBetweenEvictionRunsMillis(EVICTION_RUN_INTERVAL_MILLIS);
         String redisAddress = propertiesProvider.get(REDIS_ADDRESS_OPT).orElse(DEFAULT_REDIS_ADDRESS);
-        return new JedisPool(poolConfig, URI.create(redisAddress));
+        URI uri = URI.create(redisAddress);
+        // Take the host and port apart rather than passing the URI. JedisPool's URI constructors
+        // hand JedisFactory a hardcoded ssl=false and never look at the scheme, so a rediss://
+        // address would connect in plaintext and every command would time out against a TLS-only
+        // server. JedisPool(String) reads the scheme, but it accepts no pool config, and the
+        // settings above are the point of this factory.
+        boolean ssl = SSL_SCHEME.equalsIgnoreCase(uri.getScheme());
+        int port = uri.getPort() == -1 ? Protocol.DEFAULT_PORT : uri.getPort();
+        return new JedisPool(poolConfig, uri.getHost(), port, Protocol.DEFAULT_TIMEOUT,
+                JedisURIHelper.getPassword(uri), JedisURIHelper.getDBIndex(uri), null, ssl);
     }
 }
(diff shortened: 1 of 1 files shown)


datashare-app/src/main/java/org/icij/datashare/session/RedisPoolFactory.java
```diff
@@ -25,6 +27,10 @@ static JedisPool createPool(PropertiesProvider propertiesProvider) {
         poolConfig.setTestWhileIdle(true);
         poolConfig.setTimeBetweenEvictionRunsMillis(EVICTION_RUN_INTERVAL_MILLIS);
         String redisAddress = propertiesProvider.get(REDIS_ADDRESS_OPT).orElse(DEFAULT_REDIS_ADDRESS);
-        return new JedisPool(poolConfig, URI.create(redisAddress));
+        URI uri = URI.create(redisAddress);
+        boolean ssl = "rediss".equalsIgnoreCase(uri.getScheme());
```
pirhoo, line 35: `URI.create` here loses the validation the old path had: `JedisFactory(URI, ...)` ran `JedisURIHelper.isValid` and threw `InvalidURIException("Cannot open Redis connection due invalid URI. %s")`.

Without it, a `redisAddress` missing the `//` (`redis:6379`, `localhost:6379`) parses as an opaque URI, so `uri.getPath()` is `null` and `JedisURIHelper.getDBIndex` throws a bare `NullPointerException` with the address nowhere in the message. And `redisAddress=redis` gives `host=null`, builds the pool fine, then fails on first borrow with "Could not get a resource from the pool", the same symptom #2346 is about.

One guard covers both, since an opaque or relative URI always has a null host. Not `isValid`, which also rejects a missing port and would undo your `DEFAULT_PORT` fallback.

```suggestion
        URI uri = URI.create(redisAddress);
        if (uri.getHost() == null) {
            throw new InvalidURIException("Cannot open Redis connection due invalid URI. " + redisAddress);
        }
        boolean ssl = "rediss".equalsIgnoreCase(uri.getScheme());
```

Needs `import redis.clients.jedis.exceptions.InvalidURIException;`.

pirhoo (changes requested):
Thanks Thomas, we will likely add small regression tests after this is merged! In the meantime I just requested one small change.

datashare-app/src/main/java/org/icij/datashare/session/RedisPoolFactory.java
neoReuters: Good catch, applied as suggested.

Confirmed the behaviour against the bundled Jedis 2.9.0 before taking it, including why `getHost()` is the right guard rather than `isValid`:

| `redisAddress` | opaque | `getHost()` | `isValid` | `getDBIndex` |
| --- | --- | --- | --- | --- |
| `redis://localhost:6379` | false | `localhost` | true | 0 |
| `rediss://host:6379` | false | `host` | true | 0 |
| `redis:6379` | **true** | `null` | false | **NullPointerException** |
| `localhost:6379` | **true** | `null` | false | **NullPointerException** |
| `redis` | false | `null` | false | 0, then a pool on a null host |
| `redis://localhost` | false | `localhost` | **false** | 0 |

The last row is the one that decides it: `isValid` rejects a missing port, so it would have undone the `DEFAULT_PORT` fallback. The null-host check covers all three bad cases and leaves that one to the fallback, exactly as you described.

neoReuters:
Done — null-host guard applied in 86f5f389.

pirhoo (approved):
Perfect, thanks!

PR #2350, opened 2026-08-20
fix(search): route async search per index distribution

### LINKED ISSUE #2349: bug: async search ignores index.distribution, so search returns 500 on OpenSearch
## Summary

Datashare 21.x supports both index distributions — `src/enums/esDistributions.js` defines `ELASTICSEARCH` and `OPENSEARCH`, `RootResource` reports `index.distribution`, and there is a `neo4jAppUsesOpensearch` setting. But the async-search client consults neither, so on an OpenSearch backend the document list returns 500 and no results render.

`submitAsyncSearch` hardcodes the Elasticsearch route ([`src/api/elasticsearch.js#L258`](https://github.com/ICIJ/datashare-client/blob/main/src/api/elasticsearch.js#L258)), as do `getAsyncSearch` (L276) and `deleteAsyncSearch` (L292). `runAsyncSearch` ([`src/api/asyncSearch.js#L155-L173`](https://github.com/ICIJ/datashare-client/blob/main/src/api/asyncSearch.js#L155-L173)) reads `envelope.is_running` and `envelope.response`. The caller in `useSearchFilter` calls it unconditionally, and the wrapper rethrows, so there is no fallback.

```
ERROR Fluent - Unable to POST /api/index/search/<index>/_async_search
Caused by: ResponseException: ... status line [HTTP/1.1 400 Bad Request]
{"error":"no handler found for uri [/<index>/_async_search] and method [POST]"}
	at org.icij.datashare.web.IndexResource.esPost(IndexResource.java:134)
```

## This is the only gap

Same Datashare 21.18.0 build, identical requests, against Elasticsearch 8.15 and OpenSearch 2.11 — every endpoint `IndexResource` exposes:

| Request | ES 8.15 | OpenSearch 2.11 |
| --- | --- | --- |
| `GET /`, `PUT /{index}`, `HEAD /{index}` | 200 | 200 |
| `_search`, `_count`, `_search/scroll` | 200 | 200 |
| **`_async_search`** | **200** | **400** |
| `_close`, `_open`, `_snapshot` | 200 | 200 |
| `_nodes`, `_nodes/settings`, `_cluster/settings` | 200 | 200 |
| `DELETE /{index}` | 200 | 200 |

Also verified end to end through Datashare itself: `PUT /api/index/<name>` returns 201 with the full 21.x mappings on both, and `POST /api/index/search/<name>/_async_search` returns 200 on Elasticsearch and 500 on OpenSearch.

## Mapping

OpenSearch has the capability, as the `opensearch-asynchronous-search` plugin, under a different route and envelope:

| | Elasticsearch | OpenSearch |
| --- | --- | --- |
| submit | `POST /{indices}/_async_search` | `POST /_plugins/_asynchronous_search?index={indices}` |
| poll | `GET /_async_search/{id}` | `GET /_plugins/_asynchronous_search/{id}` |
| cancel | `DELETE /_async_search/{id}` | `DELETE /_plugins/_asynchronous_search/{id}` |
| running flag | `is_running` / `is_partial` | `state`: `RUNNING`, `CLOSED`, `STORE_RESIDENT` |
| retain after completion | implicit once an id is issued | requires `keep_on_completion=true` |

Comma-separated indices work in the `index` query parameter, so multi-project search is preserved (measured: 89 + 229 = 318 hits).

**The retention row is the trap.** Measured on both:

| | ES 8.15 | OpenSearch 2.11 |
| --- | --- | --- |
| submit `wait_for_completion_timeout=1ms`, no flag | `is_running: true` + id | `state: RUNNING` + id |
| poll 3s later | 200, full result | **404** |
| poll with `keep_on_completion=true` | n/a | 200, `STORE_RESIDENT` |
| completes within the wait | **no id** (matches `IndexResource.java:165`) | — |

A route rewrite alone would therefore turn a visible 500 into silent result loss for any search slower than `wait_for_completion_timeout`. Injecting `keep_on_completion=true` is not extra behaviour; it is what makes OpenSearch match the Elasticsearch semantics the client already assumes.

## Two ways to resolve it

**1. Adapt server-side.** Keep one wire contract and translate in `IndexResource`, which already holds async-specific logic (`isAsyncSearchSubmit`, `withInjectedKeepAlive`, `recordAsyncSearchOwnership`, `asyncSearchStatus`): rewrite the route, inject `keep_on_completion=true`, and normalise `state` to `is_running`/`is_partial`. The client, `IndexAccessVerifier` and `AsyncSearchStore` need no changes — `AsyncSearchStore` keys on `id`, which both envelopes carry.

This keeps distribution differences out of `IndexAccessVerifier`, which validates granted indices from `pathParts[0]`. Moving indices into a query parameter at the client would push that variance into the authorization path.

**2. Use synchronous search on OpenSearch.** Fewer moving parts, but it forfeits async search on that backend and returns long queries to HTTP timeouts.

Option 1 preserves the feature; option 2 is smaller. Happy to send a PR for whichever you prefer.

## Open questions

- Is `keep_on_completion=true` acceptable? It persists results to an OpenSearch system index. The existing `AsyncSearchStore` keep-alive and the DELETE path already provide cleanup.
- I observed `RUNNING`, `CLOSED` and `STORE_RESIDENT`. OpenSearch documents `FAILED` and `PERSIST_*`, which I did not reproduce, so I would suggest treating `RUNNING` as running and everything else as complete, with failures surfaced from the response body.

## Why CI did not catch it

[`tests/unit/specs/api/elasticsearch.asyncSearch.spec.js`](https://github.com/ICIJ/datashare-client/blob/main/tests/unit/specs/api/elasticsearch.asyncSearch.spec.js) mocks the transport and asserts the path string, with Elasticsearch envelopes in the fixtures:

```js
const envelope = { id: 'abc', is_running: true }
const spy = vi.spyOn(elasticsearch.transport, 'request').mockResolvedValue(envelope)
...
path: '/a,b/_async_search',
```

The tests pin the Elasticsearch contract, so they pass regardless of what the target server routes. No integration coverage exercises OpenSearch.

Reported from Datashare 21.18.0 on OpenSearch 2.11 (AWS OpenSearch Service), with `opensearch-asynchronous-search` installed.

neoReuters:
Fixes #2349. This is option 1 from that issue — keep async search working on both distributions. Happy to switch to option 2 (synchronous fallback on OpenSearch) if you would rather not carry the translation.

`submitAsyncSearch` targets the Elasticsearch API unconditionally, so on OpenSearch every search returns 500. OpenSearch has the same capability as the `opensearch-asynchronous-search` plugin, with the indices in an `index` parameter and progress in a `state` field.

`IndexResource` already owns async-search behaviour (`isAsyncSearchSubmit`, `withInjectedKeepAlive`, `recordAsyncSearchOwnership`, `asyncSearchStatus`), and `RootResource` already resolves `index.distribution`, so the translation goes there. The client, `IndexAccessVerifier` and `AsyncSearchStore` are unchanged — keeping distribution differences out of `IndexAccessVerifier` matters, because it validates granted indices from `pathParts[0]` and moving the indices into a query parameter would push that variance into the authorization path.

Three adaptations, all no-ops on Elasticsearch:

| | Elasticsearch | OpenSearch |
| --- | --- | --- |
| submit | `POST /{indices}/_async_search` | `POST /_plugins/_asynchronous_search?index={indices}` |
| poll, cancel | `/_async_search/{id}` | `/_plugins/_asynchronous_search/{id}` |
| progress | `is_running`, `is_partial` | `state` |
| retain on completion | implicit once an id is issued | needs `keep_on_completion=true` |

The last row is the one that is easy to miss. Elasticsearch keeps a result as soon as it hands out an id. OpenSearch discards it on completion unless asked. Without `keep_on_completion=true`, a search that outlives `wait_for_completion_timeout` returns an id and then 404s on the first poll — which would be worse than the current failure, because it loses results quietly.

`getVersion()` is a call to the cluster and the distribution cannot change while the process runs, so it is resolved once. If it cannot be read, the code assumes Elasticsearch and logs a warning, so existing deployments keep their current behaviour.

## Verification

Datashare 21.18.0 in Docker against Elasticsearch 8.15.0 and OpenSearch 2.11.1, testing through `/api/index/search/...` rather than against the engines directly. The patched class was placed in `dist/`, which precedes the distribution jar on the entrypoint classpath.

| | before | after |
| --- | --- | --- |
| OpenSearch: submit | 500 `Unable to apply route` | 200, `is_running:false`, `is_partial:false`, no `state` |
| OpenSearch: poll | — | 200, with hits |
| OpenSearch: cancel | — | 200 `{"acknowledged":true}` |
| Elasticsearch: submit | 200 | 200, unchanged |

The poll returning 200 is what shows `keep_on_completion` is being applied; without it OpenSearch answers 404.

Also measured, to confirm this is the only divergence — every endpoint `IndexResource` exposes, same build, both engines: `GET /`, `PUT /{index}`, `HEAD /{index}`, `_search`, `_count`, `_search/scroll`, `_close`, `_open`, `_snapshot`, `_nodes`, `_nodes/settings`, `_cluster/settings`, `DELETE /{index}` all return 200 on both. Only `_async_search` differed.

**What I have not done:** run the project's Maven test suite, and I have not added tests. If you want them, tell me where they belong — `IndexResourceTest` looked like the natural home, and I would rather match your conventions than guess. I observed `RUNNING`, `CLOSED` and `STORE_RESIDENT` from OpenSearch; the code treats only `RUNNING` as in-flight, so the documented `FAILED` and `PERSIST_*` states are handled as finished, with errors surfaced from the response body.

### DIFF
--- a/datashare-app/src/main/java/org/icij/datashare/web/IndexResource.java
+++ b/datashare-app/src/main/java/org/icij/datashare/web/IndexResource.java
@@ -1,6 +1,7 @@
 package org.icij.datashare.web;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
 import io.swagger.v3.oas.annotations.Operation;
@@ -40,6 +41,11 @@ import static net.codestory.http.payload.Payload.ok;
 @Prefix("/api/index")
 public class IndexResource {
     private static final Logger logger = LoggerFactory.getLogger(IndexResource.class);
+    private static final String OPENSEARCH_DISTRIBUTION = "opensearch";
+    private static final String ES_ASYNC_SEARCH_PATH = "_async_search";
+    private static final String OPENSEARCH_ASYNC_SEARCH_PATH = "_plugins/_asynchronous_search";
+    private static final String OPENSEARCH_RUNNING_STATE = "RUNNING";
+    private volatile Boolean openSearchBackend;
     private final Indexer indexer;
     private final AsyncSearchStore asyncSearchStore;
     private final ModeVerifier modeVerifier;
@@ -131,7 +137,14 @@ public class IndexResource {
     public Payload esPost(@Parameter(name = "index", description = "elasticsearch path", in = ParameterIn.PATH) final String path, Context context, final net.codestory.http.Request request) throws IOException {
         try {
             String esUrl = withInjectedKeepAlive(IndexAccessVerifier.checkPath(path, context), path, context);
+            boolean translateAsyncSearch = IndexAccessVerifier.isAsyncSearchSubmit(path) && isOpenSearch();
+            if (translateAsyncSearch) {
+                esUrl = toOpenSearchAsyncSubmitUrl(esUrl);
+            }
             String response = indexer.executeRaw("POST", esUrl, new String(request.contentAsBytes()));
+            if (translateAsyncSearch) {
+                response = toElasticsearchAsyncEnvelope(response);
+            }
             if (IndexAccessVerifier.isAsyncSearchSubmit(path)) {
                 try {
                     recordAsyncSearchOwnership(path, context, response);
@@ -206,7 +219,14 @@ public class IndexResource {
             return PayloadFormatter.error("async search not found", HttpStatus.NOT_FOUND);
         }
         try {
-            String response = indexer.executeRaw(method, IndexAccessVerifier.getUrlString(context, path), null);
+            String statusUrl = IndexAccessVerifier.getUrlString(context, path);
+            if (isOpenSearch()) {
+                statusUrl = toOpenSearchAsyncStatusUrl(statusUrl);
+            }
+            String response = indexer.executeRaw(method, statusUrl, null);
+            if (isOpenSearch() && !"DELETE".equalsIgnoreCase(method)) {
+                response = toElasticsearchAsyncEnvelope(response);
+            }
             if ("DELETE".equalsIgnoreCase(method)) {
                 // ES already cancelled the search; a store failure (e.g. Redis down) shouldn't surface
                 // as a 500 to the user. The TTL will reap the orphaned record on its own.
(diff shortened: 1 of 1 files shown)


pirhoo:
Closed in favor of https://github.com/ICIJ/datashare-client/pull/508

Issue #2206, opened 2026-06-16
feat: StructuredEntityExtractionTask and factory wiring


pirhoo:
Part of #2194.

Wire everything into a background task that runs a mapping from start to finish.

Named `StructuredEntityExtractionTask` (not `StructuredExtractionTask`) to avoid confusion with structured *content* extraction (un-flattening tables at the Tika boundary), which is separate and deferred. This task does structured *entity* extraction: rows to schema entities.

- [ ] Add `StructuredEntityExtractionTask`: read the mapping, read the file, run the executor, write the statements, rebuild the entities index.
- [ ] Register it on the task factory, adding a pipeline stage if needed.
- [ ] Match the factory method name and signature exactly (`createStructuredEntityExtractionTask`). It is resolved by reflection and fails at runtime, not at compile time, if it is wrong.
- [ ] Test a full run, and that running it again produces the same result.

Depends on: #2204, #2205, #2201

ClemDoum:
I would rename `StructuredExtractionTask` into `StructuredEntityExtractionTask` to avoid confusion with structured content extraction

pirhoo:
Sure! The distinction is worth making explicit: this is structured entity extraction (rows to schema entities), not structured content extraction (un-flattening tables and document structures), which is separate.

Note this also renames the factory method, since `TaskFactoryHelper` resolves it by reflection on the class simple name: `createStructuredEntityExtractionTask(...)`. I'll keep the two in sync.

pirhoo:
Two follow-ups land here out of #2204, both deliberate and both recorded when #2204 was designed on 2026-08-26.

**Delete-by-document retraction, and the `(prj_id, doc_id)` index it needs.** #2204's entity id is `SHA-384(model, type, sorted key values)`, so re-running the *same* mapping over the same file is already a no-op: every statement id comes out identical and `onConflictDoNothing` writes nothing. What is still open is *correcting* a mapping and re-running. Mappings are immutable, so a correction is a new mapping with a new id, and the corrected statements insert alongside the stale ones rather than replacing them. `entity()` then unions both, which in the worst case (a type correction, `Person` to `Company`) yields a two-typed entity with no common schema.

[...]

Issue #2204, opened 2026-06-16
feat: mapping executor, mapping to statements


pirhoo:
Part of #2194.

The core piece: turn a mapping plus a file's rows into statements. Confirmed 2026-08-19 that the FtM library provides none of this (see the corrected premise on #2200), so all of it is ours.

- [x] Build a stable entity id from the mapping's key columns.
- [x] Fill each property from a column, a fixed value, or a reference to another entity.
- [x] Support relationships between entities, for example an employment linking a person and a company.
- [x] Apply value formatting (such as dates) and keep the original raw value alongside it.
- [x] Look up types and properties through the model interface so this stays independent of FtM.
- [x] Test that a sample mapping and file produce the expected entities and at least one relationship.

Depends on: #2200, #2202, #2198

## Added 2026-08-25, when #2202 was designed

- **This issue now owns Liquibase changeset `048`.** #2202 deliberately ships `statement` without `original_value`. Nothing writes a statement until #2206, which is downstream of this issue, so there is no production data to retrofit and no deleted source to lose; adding the column in `047` would have meant one that could not be filled or tested. So "keep the original raw value alongside it" here means a migration **plus** a new field on `Statement` in `datashare-api`, not executor code alone. Budget accordingly.
- **Why the entity id must be deterministic.** `Statement.id` is a SHA-384 over nine components, one of which is `entityId`. A per-run UUID therefore silently breaks #2202's "writing the same data twice changes nothing except last-seen": every re-run mints new ids and duplicates the table instead of touching `last_seen`. Derive it from the mapping's `keys` columns as the first checkbox says, and #2202's idempotence test is what proves it.
- **Where the raw value has to come from.** `Row` is `(long number, Map<String,String> values)` and `WorkbookRowSource` formats at read time, so producing a value distinct from the raw one needs either the formatting moved into this issue or a second map on `Row`. That is this issue's call.

pirhoo:
## Added 2026-08-25, after #2202's review

[...]

PR #2353, opened 2026-08-21
feat(tabular): tabular source reader

### LINKED ISSUE #2198: feat: tabular source reader
Part of #2194.

Read rows straight out of tabular files while keeping the columns intact. This deliberately avoids the normal text-extraction path, which flattens tables into plain text and loses the structure.

- [ ] Define a small reader interface that streams each row as a column-name to value map, chosen by content type and overridable in the mapping.
- [ ] Read CSV and similar delimited files (tab, semicolon, pipe), with delimiter, quoting and encoding configurable. Default to the delimiter Tika already detected at index time.
- [ ] Read Excel files with POI: modern .xlsx, macro-enabled .xlsm and legacy .xls.
- [ ] Read JSON and line-delimited JSON, flattening nested values to dotted column names.
- [ ] Add one fallback reader over Tika's XHTML tables, which covers .ods, .xlsb, .docx/.doc, HTML and .numbers behind the same interface.
- [ ] Resolve the document named by the mapping through the index and stream its bytes, so a file embedded in an archive or an email works too.
- [ ] Test that each reader returns the same rows for an equivalent file.

Big-data formats (Parquet, ORC, Avro) are out of scope. PDF and RTF are unclaimed rather than out of scope: neither yields tables through Tika, so they need a real table extractor, which would arrive as one more reader behind the same interface.

pirhoo:
Reads rows straight out of tabular files, keeping the columns intact instead of going through the text-extraction path that flattens tables. Part of #2194, closes #2198. No production caller yet: #2204 and #2206 are the consumers.

- feat: add the `RowSource` contract with `Row` and `RowSourceOptions` in `datashare-api`
- feat: read delimited files (comma, tab, semicolon, pipe) with configurable delimiter, quote and charset
- feat: read xlsx, xlsm and legacy xls with POI
- feat: read json arrays and line-delimited json, flattening nested values to dotted column names
- feat: add a Tika XHTML table fallback covering ods, xlsb, docx/doc, html and numbers
- feat: resolve a document id to its rows through the index, so a file embedded in an archive or an email works
- fix: read a workbook cell's own value rather than its rendering, so long ids keep their digits and currency and percentage cells arrive as their underlying value
- fix: stop reading spans, nested tables and footers as data
- fix: refuse a nameless header row, and a json source that is empty or runs past its array
- fix: release the source on every failed read
- chore: drop the duplicate poi-ooxml-lite copy of the ooxml schemas
- test: 99 tests, one of them asserting every reader returns the same rows for an equivalent file

## Architecture

```mermaid
flowchart LR
  ID["project + documentId + rootId"] --> R["TabularRowReader"]
  R -->|"indexer.get, content excluded"| DOC[("Document: type, name, charset, metadata")]
  R -->|"SourceExtractor"| SRC[("source bytes, root or embedded")]
  DOC --> RES["resolve: content type, charset, delimiter"]
  RES --> SEL{"select by content type"}
  SEL -->|tier 1| D["DelimitedRowSource"] & W["WorkbookRowSource"] & J["JsonRowSource"]
  SEL -->|tier 2 fallback| T["TikaTableRowSource"]
  D & W & J & T --> ROWS(["Stream&lt;Row&gt;"])
```

**Routing.** The four `SUPPORTED` sets are disjoint, so selection order is inert today.

| Reader | Claims |
|---|---|
| `DelimitedRowSource` | `text/csv`, `text/tab-separated-values`, `text/tsv`, `text/plain` |
| `WorkbookRowSource` | xlsx, `application/vnd.ms-excel`, xlsm |
| `JsonRowSource` | `application/json`, `application/x-ndjson` |
| `TikaTableRowSource` (tier 2) | ods, xlsb, docx, doc, html, numbers + `.13` + `.18` |

**The two ladders**, worth reading twice:

- Content type: mapping override → stored type → *if generic* (`text/plain`, `application/octet-stream`, `unknown`), refine by extension.
- Delimiter: explicit option → Tika's `csv:delimiter` → extension (`tsv`, `psv`) → implied by a TSV content type → comma.

**Ownership.** `RowSource.rows` owns the source and releases it when the returned stream closes, or immediately if it throws before returning one. `TabularRowReader` also wraps with `onClose`, so the double close is expected and harmless.

**Deliberately absent**, so nobody spends a round on it:

- No production caller: #2204 and #2206 are the consumers.
- Authorization and the size guard belong to #2206, which must route through `DocumentSourceAccess.gated`.
- Tier 1 and tier 2 differ on purpose: tier 1 refuses surplus non-blank cells and forces ISO-8601 dates, tier 2 tolerates and passes dates through.

Reading order: `RowSource` → `Row` → `TabularRowReader` → the four readers.

### DIFF
--- a/datashare-api/src/main/java/org/icij/datashare/tabular/Row.java
+++ b/datashare-api/src/main/java/org/icij/datashare/tabular/Row.java
@@ -2,9 +2,11 @@ package org.icij.datashare.tabular;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 /**
  * One data row. {@code number} is the 1-based ordinal of the emitted row, not the row's position in
@@ -16,22 +18,36 @@ public record Row(long number, Map<String, String> values) {
     /**
      * Applies the one header rule every reader shares: a blank name means the column is dropped
      * (represented by a null entry the caller skips), a duplicate name is a mapping the reader
-     * cannot honour unambiguously and so fails.
+     * cannot honour unambiguously and so fails, and a header row with no name at all is not a header
+     * row: importing every data row as an empty map would report a total loss as a success.
      */
     public static List<String> headers(List<String> rawNames) {
         List<String> headers = new ArrayList<>();
+        Set<String> seen = new HashSet<>();
         for (String rawName : rawNames) {
-            String name = rawName == null ? "" : rawName.strip();
+            String name = rawName == null ? "" : strip(rawName);
             if (name.isEmpty()) {
                 headers.add(null);
                 continue;
             }
-            if (headers.contains(name)) {
+            if (!seen.add(name)) {
                 throw new IllegalArgumentException("duplicate column name in header: " + name);
             }
             headers.add(name);
         }
+        if (seen.isEmpty()) {
+            throw new IllegalArgumentException("no column name in the header row: every name is blank, "
+                    + "so the row above the data is a title or a spacer rather than a header");
+        }
         return headers;
+    }
+
+    // String.strip leaves a UTF-8 byte order mark in place, because U+FEFF is not whitespace. It can
+    // only ever precede the first name, and Excel writes one on every "CSV UTF-8" export, so without
+    // this the first column is named BOM + "id" and no mapping can address it.
+    private static String strip(String rawName) {
+        String name = rawName.strip();
+        return name.startsWith("\uFEFF") ? name.substring(1).strip() : name;
     }
 
     /**
--- a/datashare-api/src/main/java/org/icij/datashare/tabular/Row.java
+++ b/datashare-api/src/main/java/org/icij/datashare/tabular/Row.java
@@ -53,10 +53,13 @@ public record Row(long number, Map<String, String> values) {
     /**
      * The same mapping, refusing a row carrying more fields than the header declares: that is the
      * signature of a wrong delimiter or of a separator inside an unquoted value, and importing such a
-     * row would silently misalign every value in it.
+     * row would silently misalign every value in it. Only a surplus cell that is not blank counts,
+     * because both of those signatures put content in it: a trailing delimiter in a csv and a styled
+     * or previously emptied cell past the last filled column in a workbook produce an empty surplus
+     * that carries no data and cannot misalign anything.
      */
     public static Map<String, String> values(List<String> headers, List<String> cells, long number) {
-        if (cells.size() > headers.size()) {
+        if (cells.stream().skip(headers.size()).anyMatch(cell -> !cell.isBlank())) {
             throw new IllegalArgumentException("row " + number + " has " + cells.size()
                     + " fields but the header declares " + headers.size());
         }
--- a/datashare-api/src/main/java/org/icij/datashare/tabular/Row.java
+++ b/datashare-api/src/main/java/org/icij/datashare/tabular/Row.java
@@ -1,0 +1,30 @@
+package org.icij.datashare.tabular;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+/** One data row. {@code number} is 1-based and excludes the header row. */
+public record Row(long number, Map<String, String> values) {
+
+    /**
+     * Applies the one header rule every reader shares: a blank name means the column is dropped
+     * (represented by a null entry the caller skips), a duplicate name is a mapping the reader
+     * cannot honour unambiguously and so fails.
+     */
+    public static List<String> headers(List<String> rawNames) {
+        List<String> headers = new ArrayList<>();
+        for (String rawName : rawNames) {
+            String name = rawName == null ? "" : rawName.strip();
+            if (name.isEmpty()) {
+                headers.add(null);
+                continue;
+            }
+            if (headers.contains(name)) {
+                throw new IllegalArgumentException("duplicate column name in header: " + name);
+            }
+            headers.add(name);
+        }
+        return headers;
+    }
+}
--- a/datashare-api/src/main/java/org/icij/datashare/tabular/Row.java
+++ b/datashare-api/src/main/java/org/icij/datashare/tabular/Row.java
@@ -1,10 +1,16 @@
 package org.icij.datashare.tabular;
 
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
-/** One data row. {@code number} is 1-based and excludes the header row. */
+/**
+ * One data row. {@code number} is the 1-based ordinal of the emitted row, not the row's position in
+ * the file: the header row and the rows a reader skips, an entirely blank one for instance, consume
+ * no number.
+ */
 public record Row(long number, Map<String, String> values) {
 
     /**
@@ -26,5 +32,34 @@ public record Row(long number, Map<String, String> values) {
             headers.add(name);
         }
         return headers;
+    }
+
+    /**
+     * Maps a row's cells onto the header names, in header order. Every column the header declares is
+     * present, padded with an empty string when the row is short, so a consumer never has to tell a
+     * missing cell from an empty one. A column whose header name is blank is dropped, and cells past
+     * the last declared column are ignored.
+     */
+    public static Map<String, String> values(List<String> headers, List<String> cells) {
+        Map<String, String> values = new LinkedHashMap<>();
+        for (int column = 0; column < headers.size(); column++) {
+            if (headers.get(column) != null) {
+                values.put(headers.get(column), column < cells.size() ? cells.get(column) : "");
+            }
+        }
+        return Collections.unmodifiableMap(values);
+    }
+
+    /**
+     * The same mapping, refusing a row carrying more fields than the header declares: that is the
+     * signature of a wrong delimiter or of a separator inside an unquoted value, and importing such a
+     * row would silently misalign every value in it.
+     */
+    public static Map<String, String> values(List<String> headers, List<String> cells, long number) {
+        if (cells.size() > headers.size()) {
+            throw new IllegalArgumentException("row " + number + " has " + cells.size()
+                    + " fields but the header declares " + headers.size());
+        }
+        return values(headers, cells);
     }
 }
--- a/datashare-api/src/main/java/org/icij/datashare/tabular/Row.java
+++ b/datashare-api/src/main/java/org/icij/datashare/tabular/Row.java
@@ -25,7 +25,7 @@ public record Row(long number, Map<String, String> values) {
         List<String> headers = new ArrayList<>();
         Set<String> seen = new HashSet<>();
         for (String rawName : rawNames) {
-            String name = rawName == null ? "" : strip(rawName);
+            String name = rawName == null ? "" : rawName.strip();
             if (name.isEmpty()) {
                 headers.add(null);
                 continue;
@@ -40,14 +40,6 @@ public record Row(long number, Map<String, String> values) {
                     + "so the row above the data is a title or a spacer rather than a header");
         }
         return headers;
-    }
-
-    // String.strip leaves a UTF-8 byte order mark in place, because U+FEFF is not whitespace. It can
-    // only ever precede the first name, and Excel writes one on every "CSV UTF-8" export, so without
-    // this the first column is named BOM + "id" and no mapping can address it.
-    private static String strip(String rawName) {
-        String name = rawName.strip();
-        return name.startsWith("\uFEFF") ? name.substring(1).strip() : name;
     }
 
     /**
--- a/datashare-app/src/main/java/org/icij/datashare/tabular/DelimitedRowSource.java
+++ b/datashare-app/src/main/java/org/icij/datashare/tabular/DelimitedRowSource.java
@@ -9,10 +9,8 @@ import java.io.InputStream;
 import java.io.UncheckedIOException;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
-import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
 import java.util.Spliterator;
 import java.util.Spliterators;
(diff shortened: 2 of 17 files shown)


pirhoo:
## Architecture

```mermaid
flowchart LR
  ID["project + documentId + rootId"] --> R["TabularRowReader"]
  R -->|"indexer.get, content excluded"| DOC[("Document: type, name, charset, metadata")]
  R -->|"SourceExtractor"| SRC[("source bytes, root or embedded")]
  DOC --> RES["resolve: content type, charset, delimiter"]
  RES --> SEL{"select by content type"}
  SEL -->|tier 1| D["DelimitedRowSource"] & W["WorkbookRowSource"] & J["JsonRowSource"]
  SEL -->|tier 2 fallback| T["TikaTableRowSource"]
  D & W & J & T --> ROWS(["Stream&lt;Row&gt;"])
```

**Routing.** The four `SUPPORTED` sets are disjoint, so selection order is inert today.

| Reader | Claims |
|---|---|
| `DelimitedRowSource` | `text/csv`, `text/tab-separated-values`, `text/tsv`, `text/plain` |
| `WorkbookRowSource` | xlsx, `application/vnd.ms-excel`, xlsm |
| `JsonRowSource` | `application/json`, `application/x-ndjson` |
| `TikaTableRowSource` (tier 2) | ods, xlsb, docx, doc, html, numbers + `.13` + `.18` |

**The two ladders**, worth reading twice:

- Content type: mapping override → stored type → *if generic* (`text/plain`, `application/octet-stream`, `unknown`), refine by extension.
- Delimiter: explicit option → Tika's `csv:delimiter` → extension (`tsv`, `psv`) → implied by a TSV content type → comma.

**Ownership.** `RowSource.rows` owns the source and releases it when the returned stream closes, or immediately if it throws before returning one. `TabularRowReader` also wraps with `onClose`, so the double close is expected and harmless.

**Deliberately absent**, so nobody spends a round on it:

- No production caller: #2204 and #2206 are the consumers.
- Authorization and the size guard belong to #2206, which must route through `DocumentSourceAccess.gated`.
- Tier 1 and tier 2 differ on purpose: tier 1 refuses surplus non-blank cells and forces ISO-8601 dates, tier 2 tolerates and passes dates through.

Reading order: `RowSource` → `Row` → `TabularRowReader` → the four readers.

datashare-api/src/main/java/org/icij/datashare/tabular/Row.java
```diff
@@ -0,0 +1,84 @@
...
+        }
+        return headers;
+    }
+
+    // String.strip leaves a UTF-8 byte order mark in place, because U+FEFF is not whitespace. It can
+    // only ever precede the first name, and Excel writes one on every "CSV UTF-8" export, so without
+    // this the first column is named BOM + "id" and no mapping can address it.
+    private static String strip(String rawName) {
+        String name = rawName.strip();
+        return name.startsWith("\uFEFF") ? name.substring(1).strip() : name;
```
ClemDoum: I'm not sure to understand this one. Shouldn't excel related logic leave in the excel related code (maybe `WorkbookRowSource.java`)

datashare-app/src/main/java/org/icij/datashare/tabular/DelimitedRowSource.java
```diff
@@ -0,0 +1,182 @@
...
+    // parsed with a comma.
+    private static char delimiter(RowSourceOptions options) {
+        if (options.delimiter() != null) {
+            return options.delimiter();
+        }
+        return options.contentType() != null && TAB_SEPARATED_TYPES.contains(options.contentType())
+                ? '\t' : DEFAULT_DELIMITER;
+    }
+
+    private static Stream<Row> stream(Iterator<CSVRecord> records, List<String> headers) {
```
ClemDoum, line 138: Is it really necessary ?  `CSVRecord` provides `toMap` couldn't we just transform the iterator into a stream then filter it to remove empty rows ?

datashare-api/src/main/java/org/icij/datashare/tabular/Row.java
```diff
@@ -0,0 +1,84 @@
...
+     * Maps a row's cells onto the header names, in header order. Every column the header declares is
+     * present, padded with an empty string when the row is short, so a consumer never has to tell a
+     * missing cell from an empty one. A column whose header name is blank is dropped, and cells past
+     * the last declared column are ignored.
+     */
+    public static Map<String, String> values(List<String> headers, List<String> cells) {
+        Map<String, String> values = new LinkedHashMap<>();
+        for (int column = 0; column < headers.size(); column++) {
+            if (headers.get(column) != null) {
+                values.put(headers.get(column), column < cells.size() ? cells.get(column) : "");
```
ClemDoum, line 55: I'm not sure to understand how headers and cell could have varying lengths.

For my headers and cell should have matching length and missing values should be null strings, empty values should be empty strings.

I'm not sure what use case we're trying to cover here ?

ClemDoum: Also if length vary then missing values are accepted but only for cells which are at the end which to me doesn't make sense.

ClemDoum: If this does make sense for some file format then I think it should be addressed there and not in the generic implem

datashare-app/src/main/java/org/icij/datashare/tabular/DelimitedRowSource.java
```diff
@@ -0,0 +1,182 @@
...
+            return stream(records, headers).onClose(() -> close(parser));
+        } catch (RuntimeException failure) {
+            closeQuietly(parser);
+            throw failure;
+        }
+    }
+
+    // Releases the source itself when the parser cannot be built, since no stream reaches the caller
+    // to be closed on that path.
+    private static CSVParser open(InputStream source, RowSourceOptions options) throws IOException {
```
ClemDoum: What about `parser` or `parseCSV` ?

datashare-app/src/main/java/org/icij/datashare/tabular/JsonRowSource.java
```diff
@@ -0,0 +1,182 @@
...
+/**
+ * Reads flat records out of JSON. One reader covers both shapes a dump comes in: peeking the first
+ * token tells an array of objects apart from line-delimited or concatenated objects, and Jackson
+ * iterates both identically once positioned.
+ *
+ * A record carries its own column names, so the header rules the other readers share do not reach
+ * here: a blank key becomes a column named "", keys are not stripped, a duplicate key is
+ * last-write-wins, and {@code {"addr.city":"X","addr":{"city":"Y"}}} collides silently.
+ */
+public class JsonRowSource implements RowSource {
```
ClemDoum, line 32: Question is for list oriented json [{}, ...., {}] or jsonl ?

datashare-app/src/main/java/org/icij/datashare/tabular/JsonRowSource.java
```diff
@@ -0,0 +1,182 @@
...
+            }
+        });
+    }
+
+    /**
+     * An array of scalars joins on a fixed separator so a mapping can rely on splitting it. An array
+     * of objects is skipped: there is no column name that would describe it. If multi-valued
+     * properties turn out to matter, the fix is a multi-value seam in Row, not a cleverer separator.
+     */
+    private static void joinScalars(String column, JsonNode array, Map<String, String> values) {
```
ClemDoum: Do we really want to do this. I understand flattening for nested object, but jointed array values sounds weird to me.

How will it work if the value joined equals the delimiter, we'll then need to dump a proper csv in a table's column, which in the end is not very column in actual tables.

I think the underlying question is can we really represent an arbitrary JSON into a CSV. For me the answer is not, we can only support JSONs which are actual tabular data.

I think should inspire ourselves from https://pandas.pydata.org/docs/reference/api/pandas.read_json.html . I don't think we should necessarily support as many format as they do, but what I like about their API is that arbitrary json are not really supported, the lib expects JSON records to comply with some formatting

datashare-app/src/main/java/org/icij/datashare/tabular/DelimitedRowSource.java
```diff
@@ -0,0 +1,182 @@
...
+
+    private static void close(CSVParser parser) {
+        try {
+            parser.close();
+        } catch (IOException e) {
+            throw new UncheckedIOException("closing the delimited source failed", e);
+        }
+    }
+
+    private static void closeQuietly(Closeable closeable) {
```
ClemDoum, line 189: I'm not sure to understand the different implems of closeQuietly in this PR.
Why not using a standard try with resources syntax which feels simpler and more idiomatic.

datashare-app/src/main/java/org/icij/datashare/tabular/TabularRowReader.java
```diff
@@ -0,0 +1,196 @@
...
+     */
+    static Character delimiterFrom(Map<String, Object> metadata) {
+        if (metadata == null) {
+            return null;
+        }
+        Object name = metadata.get(DELIMITER_METADATA_KEY);
+        return name == null ? null : DELIMITER_BY_TIKA_NAME.get(name.toString().toLowerCase(Locale.ROOT));
+    }
+
+    private static void closeQuietly(InputStream source) {
```
ClemDoum, line 189: Why not reallying on try with resources instead ?

ClemDoum: Is it because the caller has opened the stream if so shouldn't the caller keep control on it and whether is should be closed ?

datashare-app/src/main/java/org/icij/datashare/tabular/DelimitedRowSource.java
ClemDoum: Is it because the caller has opened the stream if so shouldn't the caller keep control on it and whether is should be closed ?

datashare-app/src/main/java/org/icij/datashare/tabular/TikaTableRowSource.java
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    @Override
+    public Stream<Row> rows(InputStream source, RowSourceOptions options) throws IOException {
+        // Nothing here is lazy: the extractor has consumed the source to exhaustion before the first
+        // row is built, so the source is released on the way out rather than by the returned stream.
+        try {
+            return read(source, options).stream();
+        } finally {
+            close(source);
```
ClemDoum, line 79: Same question here, should we close the source if we're not the one opening it ?

ClemDoum (changes requested):

datashare-app/src/main/java/org/icij/datashare/tabular/JsonRowSource.java
pirhoo: Good call, the separator could collide with the data. Fixed in 6ff5103ab: an array value is now refused, naming the row and column, which also removes the silent skip of arrays of objects. Nested objects still flatten. That's the pandas stance you linked: json that isn't tabular is refused rather than approximated.

datashare-api/src/main/java/org/icij/datashare/tabular/Row.java
pirhoo: It's a UTF byte order mark rather than Excel logic (an xlsx can't carry one in a cell), Excel is just the most common producer via its "CSV UTF-8" export. But you have a point that only the delimited reader can ever see one, so I moved the strip there: a182870f5. Row no longer knows about it.

datashare-app/src/main/java/org/icij/datashare/tabular/DelimitedRowSource.java
pirhoo: Done, renamed to parse in 71f758071.

datashare-api/src/main/java/org/icij/datashare/tabular/Row.java
pirhoo: It's the formats themselves that only omit at the end: an xlsx row physically ends at its last defined cell, a CSV line can stop before the last column, and a missing middle value keeps its slot so the parser already hands it over as "". The padding just completes that encoding, and all three readers hit it, hence the shared place. On null vs "": the formats can't tell empty from missing (`a,,c`), so a null would promise a distinction the data doesn't carry. pandas does the same, short lines pad to NaN. The other direction is refused: more non-blank fields than headers is the wrong-delimiter signature.

datashare-app/src/main/java/org/icij/datashare/tabular/DelimitedRowSource.java
pirhoo: I'd have loved to, but toMap hands the header rules to commons-csv instead of the Row.headers rules the other readers share, silently drops surplus cells where we refuse them, and omits trailing keys where we pad with "". And the row numbering has to skip blank records and name the row on a malformed record mid-stream, which a filter can't do without the same mutable state. The iterator is that state written honestly.

datashare-app/src/main/java/org/icij/datashare/tabular/JsonRowSource.java
pirhoo: Both! The first token routes them: START_ARRAY means an array of objects, anything else is jsonl or concatenated objects. Both shapes are covered in JsonRowSourceTest.

datashare-app/src/main/java/org/icij/datashare/tabular/TabularRowReader.java
pirhoo: Answering the three close threads at once since it's one design. try-with-resources can't work here: when rows() returns, nothing has been read yet, so the parser must outlive the method. The t-w-r lives one level up, on the returned stream, which is what the RowSource javadoc asks callers to do. closeQuietly only runs on the failure paths, doing what t-w-r does when the body throws.

Same answer for ownership: the caller can't close correctly, it doesn't know when the lazy stream is exhausted, and on a failure there is no stream to close, so caller-owns would leak on every error. Hence the contract: rows() owns the source. The Tika reader is eager so it releases on the way out, same contract, which lets TabularRowReader drive all four readers identically.

Happy to swap the three private closeQuietly copies for commons-io's IOUtils.closeQuietly if you'd rather.

datashare-app/src/main/java/org/icij/datashare/tabular/DelimitedRowSource.java
pirhoo: Answered in the TabularRowReader thread: https://github.com/ICIJ/datashare/pull/2353#discussion_r3862874441

datashare-app/src/main/java/org/icij/datashare/tabular/TikaTableRowSource.java
pirhoo: Answered in the TabularRowReader thread: https://github.com/ICIJ/datashare/pull/2353#discussion_r3862874441

datashare-app/src/main/java/org/icij/datashare/tabular/DelimitedRowSource.java
ClemDoum: Clear !

datashare-api/src/main/java/org/icij/datashare/tabular/Row.java
ClemDoum: OK clear for the omission of the last values.
For the null vs empty string, in pandas NaN is for missing values which is not exactly the same as empty string value "" right ?

pirhoo: True, NaN is pandas's missing marker, but read_csv by default maps empty fields to NaN as well ('' is in the default na_values), so `a,,c` gives NaN in the middle: I pandas also collapses empty and missing into a single value, it just picks "missing" where we pick "". 

Since the formats can't encode the difference anyway, it's a representation choice, and "" keeps the map null-free so consumers never have to branch.

datashare-app/src/main/java/org/icij/datashare/tabular/JsonRowSource.java
ClemDoum: cool thanks !

datashare-api/src/main/java/org/icij/datashare/tabular/Row.java
ClemDoum: OK sounds good !

ClemDoum (approved):

pirhoo:
@ClemDoum merci !

Issue #2202, opened 2026-06-16
feat: statement store


pirhoo:
Part of #2194.

Store extracted data as individual facts ("statements": entity X has value V for property P, from run R). This makes re-running an extraction safe and lets the same entity show up across several files.

The table is the **model-neutral substrate**, not an FtM table: a statement is `entity · property · value · provenance`, where the property and the type are namespaced strings (`ftm:owner`, `schema:founder`, `wdt:P127`), so facts from different data schemas sit in it side by side (see #2197 for the agreed shape).

- [ ] Add a statement table with a deterministic id, **which data schema the fact belongs to**, **the property stored namespaced** rather than bare, the raw value as it was seen, the run it came from, and first-seen and last-seen timestamps. Add the migration for both SQLite and Postgres. Next free changelog number is `047` (checked 2026-08-19); share one changeset with #2201's `extraction_mapping` table so the jOOQ codegen cycle is paid once.
- [ ] Let an entity carry **more than one type**, rather than collapsing it to a single most-specific schema, since that is FtM's convention and not every model's.
- [ ] Add a repository that inserts-or-updates by id and can rebuild whole entities by grouping their statements.
- [ ] Keep the source/provenance field structured rather than a plain string. `ManifestRepository` in `datashare-api` (from the artifacts epic) is the local precedent for a backend-agnostic persistence contract whose implementation owns atomicity.
- [ ] Test that writing the same data twice changes nothing except last-seen.

> **Why the schema column and the namespaced property are not deferrable (2026-08-19).** They are the same cost-asymmetry argument that kept the statement shape itself in v1: adding a column now is a marginal delta, while retrofitting one later means re-running every extraction to recover which model each fact belonged to, and that is impossible for any source deleted since. The interfaces in #2197 and #2203 are the cheap half of the model-neutral promise and could be trimmed under pressure; this table is the expensive half. If anything gets cut, cut the interfaces, not the data shape.

Depends on: #2197

ClemDoum:
@pirhoo I'm not sure to understand this component and why it's needed.

From what you say above I ear two reason:
- deal with multiple runs or versions of extraction: I have the feeling that for all other extraction tasks we've made the choice to deal with this with admin roles + overwrite. I have the feeling that the issue is not really specific to this task, ideally, we should address is in an consistent way across task to make user experience more predictable
[...]

Issue #2201, opened 2026-06-16
feat: mapping persistence


pirhoo:
Part of #2194.

Store an extraction mapping so a run can be repeated exactly.

The mapping is the artifact that binds a tabular source to the data schema: it says which schema it targets, and how the table's columns fit that schema's entities and properties. This is FtM's own "mapping" (`ftm map`), so we reuse the term rather than inventing a separate "recipe" concept (see discussion in this issue).

- [x] Define the mapping object: which file to read, which entities to build, the columns that identify each entity, and how each property is filled (from a column, a fixed value, or a reference to another entity), plus value formats.
- [x] Give the mapping an explicit **target data schema** field (`model: ftm` for the MVP), resolved through the registry in #2197. Without it a mapping cannot say what it targets, the registry has nothing to select on, and the executor ends up hardcoding FtM.
- [x] Save it as JSON in a text column (table `extraction_mapping`), with a database migration for both SQLite and Postgres. Next free changelog number is `047` (checked 2026-08-19; `046-add-tasks-type.yml` exists). Share one changeset with #2202's table so the jOOQ codegen cycle is paid once.
- [x] Add a repository to read and write it, following the same JSON-in-text pattern already used for task arguments.
- [x] Test that a mapping saves and loads on both databases, and that an unknown target schema fails with a clear error rather than silently defaulting to FtM.

## Added 2026-08-25, when this was designed alongside #2202

- **Dependency corrected: this depends on #2197 *and* #2198.** The mapping record embeds `RowSourceOptions` (`datashare-api`, from #2198) verbatim rather than re-declaring its six reader knobs, so #2206 hands it straight to `RowSource.rows(stream, options)` with no translation layer to drift. This also removes #2201 from the 2026-08-19 state check's "slack, startable independently" list.
- **A mapping is immutable once saved.** `save` refuses an id that already exists; changing a mapping means a new row with a new id. Editing in place breaks the epic's reproducibility promise, since `run_id` would point at a task whose mapping no longer describes the statements it produced, and nothing would detect the divergence. Costs no schema change now, whereas versioning later needs a `version` column and a composite key, and SQLite cannot `ALTER COLUMN`.
[...]

PR #2354, opened 2026-08-25
feat: model-neutral core contracts and the FtM model in-JVM

### LINKED ISSUE #2197: feat: model-neutral core contracts
Part of #2194.

Set up the small set of interfaces so the rest of the work targets "a data model" in the abstract rather than FollowTheMoney directly. Only FtM is implemented now; this is what lets Schema.org or Wikidata be added later without reworking the core.

- [ ] Add a model interface that can look up a type and a property, validate an entity, and serialize it.
- [ ] Add a registry that selects a model by name. Only "ftm" is registered; an unknown name fails with a clear error.
- [ ] Agree on the shared shape of a stored fact ("statement"): which model it belongs to, a namespaced property, possibly several types per entity, and a structured source/provenance field rather than a plain string.
- [ ] Unit-test the registry.

Note (2026-08-19): design these contracts knowing that the FtM library on the classpath supplies nothing at runtime, no model data and no multi-valued entity (see the corrected premise on #2200). Since the result is an interface plus its only implementation, expect #2197 and #2200 to ship as one branch rather than two.

### LINKED ISSUE #2200: feat: FtM model in-JVM
Part of #2194.

Provide the FollowTheMoney data model inside the app as the first (and for the MVP, only) model implementation.

> **Premise corrected 2026-08-19.** The last checkbox below was an investigation. It is answered, and the answer changes this issue: `org.icij:ftm.java:0.3.1` (already on the classpath, `pom.xml:91`) is a **build-time code generator, not a runtime ontology**. The jar ships no `defaultModel.json` and no schema YAMLs; its loader helpers (`Utils.downloadYamlModels`, `getYamlContent`, `findParents`) are package-private and expect a YAML directory the artifact does not carry; `SourceGenerator` confirms the design. The generated types are single-valued (`Person.getName()` returns `String`, `getHeight()` returns `int`), so they cannot hold a multi-valued FtM entity. `0.3.1` (2024-11-04) is the latest release on Maven Central and the last tag, so there is no upgrade out of this. Only `org.icij.ftm.Model` is publicly reusable, as a schema wrapper over data we supply. The model is therefore a **build** item, not a reuse item.

- [ ] Bundle FtM's prebuilt `defaultModel.json` as a resource and write the loader for it, since the library on the classpath provides neither.
- [ ] Represent an entity with **multi-valued** properties. The generated `org.icij.ftm.*` interfaces are single-valued and cannot serve here.
- [ ] Implement the FtM version of the model interface (type and property lookup, validate, serialize), wrapping `org.icij.ftm.Model` if it earns its keep.
- [ ] Round-trip a sample Person entity to and from FtM JSON to prove it works.
- [x] ~~Investigate whether the library can also run a mapping and clean up values, or whether we have to build that ourselves. This answer decides how big the mapping executor task is.~~ Answered 2026-08-19: it does neither, and does not supply the model either. #2204 is entirely ours to build.

Because this is an interface plus its only implementation, expect it to ship in the same branch as #2197.

Depends on: #2197

pirhoo:
Ships the small set of interfaces the rest of #2194 targets ("a data model" rather than FollowTheMoney directly), the agreed shape of a stored fact, and FtM as the first and only implementation. New package `org.icij.datashare.model` in datashare-api, so #2201 and #2202 can see it from datashare-db. No new Maven dependency, nothing outside the new package touched, no production callers yet.

Closes #2197
Closes #2200

- feat: add `Statement`, one fact from one cell, with a deterministic SHA-384 id and cell-level provenance
- feat: add `ModelEntity.from(Collection<Statement>)`, rebuilding a multi-valued entity from its statements
- feat: add the `TargetModel` contract with structural validation shared as a default method, plus `EntityType` and `Property`
- feat: bundle FollowTheMoney 4.10.2 and load it into the contract's records
- feat: serialize and parse FtM's canonical single-entity JSON, collapsing a multi-type entity to its most specific schema
- feat: select a data model by name through `TargetModelRegistry`
- fix: resolve a property across every type of an entity, so validation no longer depends on JVM-randomised set iteration order
- fix: break the ancestor tie by name when two ancestors declare the same property, so the loaded ontology is reproducible
- fix: report an abstract type only when none of the entity's types is concrete, so `validate` agrees with `serialize`
- fix: reject a null document, a null value list and a null value in `parse`
- fix: normalise a null sheet to `""` and reject a NUL in any statement id component
- chore: bundle the upstream MIT licence and pin the vendored ontology to its source commit and sha256

**Why not [ftm.java](https://github.com/ICIJ/ftm.java)?** It is a build-time generator: the `0.3.1` jar ships no ontology (classes, `logback.xml` and its own pom only), its YAML loader is package-private, `Person.getName()` returns `String` where FtM properties are always arrays, and it is declared on `datashare-app` alone while `datashare-db` (#2201, #2202) sees only `datashare-api`. It stays the generator behind `FtmDocument` and `/api/ftm`, untouched here.

### DIFF
--- a/datashare-api/src/main/java/org/icij/datashare/model/Statement.java
+++ b/datashare-api/src/main/java/org/icij/datashare/model/Statement.java
@@ -9,18 +9,19 @@ public record Statement(String id, String model, String entityId, String entityT
 
     public Statement {
         Objects.requireNonNull(id, "id");
-        Objects.requireNonNull(model, "model");
-        Objects.requireNonNull(entityId, "entityId");
-        Objects.requireNonNull(entityType, "entityType");
-        Objects.requireNonNull(property, "property");
-        Objects.requireNonNull(value, "value");
+        model = component(model, "model");
+        entityId = component(entityId, "entityId");
+        entityType = component(entityType, "entityType");
+        property = component(property, "property");
+        value = component(value, "value");
         Objects.requireNonNull(provenance, "provenance");
     }
 
     public record Provenance(String documentId, String sheet, long rowNumber, String column) {
         public Provenance {
-            Objects.requireNonNull(documentId, "documentId");
-            Objects.requireNonNull(column, "column");
+            documentId = component(documentId, "documentId");
+            sheet = sheet == null ? "" : component(sheet, "sheet");
+            column = component(column, "column");
         }
     }
 
@@ -34,13 +35,22 @@ public record Statement(String id, String model, String entityId, String entityT
         return model + ":" + property;
     }
 
+    private static String component(String value, String field) {
+        Objects.requireNonNull(value, field);
+        if (value.indexOf('\u0000') >= 0) {
+            throw new IllegalArgumentException("'" + field + "' contains a NUL character");
+        }
+        return value;
+    }
+
     // NUL-separated rather than a printable delimiter: a cell value can hold any printable
-    // character, so only a value containing a literal NUL could forge a collision.
+    // character, and every joined component is rejected at construction if it holds a NUL, so no
+    // input can forge a collision.
     private static String id(String model, String entityId, String entityType,
                              String property, String value, Provenance provenance) {
         return Hasher.SHA_384.hash(String.join("\u0000", model, entityId, entityType, property, value,
                 provenance.documentId(),
-                provenance.sheet() == null ? "" : provenance.sheet(),
+                provenance.sheet(),
                 String.valueOf(provenance.rowNumber()),
                 provenance.column()));
     }
--- a/datashare-api/src/main/java/org/icij/datashare/model/Statement.java
+++ b/datashare-api/src/main/java/org/icij/datashare/model/Statement.java
@@ -2,10 +2,27 @@ package org.icij.datashare.model;
 
 import org.icij.datashare.text.Hasher;
 
+import java.util.Objects;
+
 public record Statement(String id, String model, String entityId, String entityType,
                         String property, String value, Provenance provenance) {
 
-    public record Provenance(String documentId, String sheet, long rowNumber, String column) { }
+    public Statement {
+        Objects.requireNonNull(id, "id");
+        Objects.requireNonNull(model, "model");
+        Objects.requireNonNull(entityId, "entityId");
+        Objects.requireNonNull(entityType, "entityType");
+        Objects.requireNonNull(property, "property");
+        Objects.requireNonNull(value, "value");
+        Objects.requireNonNull(provenance, "provenance");
+    }
+
+    public record Provenance(String documentId, String sheet, long rowNumber, String column) {
+        public Provenance {
+            Objects.requireNonNull(documentId, "documentId");
+            Objects.requireNonNull(column, "column");
+        }
+    }
 
     public static Statement of(String model, String entityId, String entityType,
                                String property, String value, Provenance provenance) {
--- a/datashare-api/src/main/java/org/icij/datashare/model/Statement.java
+++ b/datashare-api/src/main/java/org/icij/datashare/model/Statement.java
@@ -1,0 +1,30 @@
+package org.icij.datashare.model;
+
+import org.icij.datashare.text.Hasher;
+
+public record Statement(String id, String model, String entityId, String entityType,
+                        String property, String value, Provenance provenance) {
+
+    public record Provenance(String documentId, String sheet, long rowNumber, String column) { }
+
+    public static Statement of(String model, String entityId, String entityType,
+                               String property, String value, Provenance provenance) {
+        return new Statement(id(model, entityId, entityType, property, value, provenance),
+                model, entityId, entityType, property, value, provenance);
+    }
+
+    public String qualifiedProperty() {
+        return model + ":" + property;
+    }
+
+    // NUL-separated rather than a printable delimiter: a cell value can hold any printable
+    // character, so only a value containing a literal NUL could forge a collision.
+    private static String id(String model, String entityId, String entityType,
+                             String property, String value, Provenance provenance) {
+        return Hasher.SHA_384.hash(String.join("\u0000", model, entityId, entityType, property, value,
+                provenance.documentId(),
+                provenance.sheet() == null ? "" : provenance.sheet(),
+                String.valueOf(provenance.rowNumber()),
+                provenance.column()));
+    }
+}
--- a/datashare-api/src/main/java/org/icij/datashare/model/TargetModel.java
+++ b/datashare-api/src/main/java/org/icij/datashare/model/TargetModel.java
@@ -14,8 +14,17 @@ public interface TargetModel {
 
     Optional<EntityType> type(String name);
 
+    /**
+     * Writes the entity in this model's native JSON form, without validating it: an entity that
+     * {@link #validate} would reject (e.g. an abstract type) can still be serialized.
+     */
     String serialize(ModelEntity entity);
 
+    /**
+     * Reads an entity back from this model's native JSON form. The projection is lossy: a
+     * multi-type entity collapses to its most specific single type, so statements, not this JSON,
+     * remain the system of record.
+     */
     ModelEntity parse(String json);
 
     record Violation(String message) { }
@@ -24,6 +33,10 @@ public interface TargetModel {
         return type(type).map(found -> found.properties().get(name));
     }
 
+    /**
+     * Checks the entity's types and properties against this model's structure. Returns every
+     * violation found, or an empty list if the entity is structurally valid.
+     */
     default List<Violation> validate(ModelEntity entity) {
         List<Violation> violations = new ArrayList<>();
         if (entity.types().isEmpty()) {
--- a/datashare-api/src/main/java/org/icij/datashare/model/TargetModel.java
+++ b/datashare-api/src/main/java/org/icij/datashare/model/TargetModel.java
@@ -1,9 +1,11 @@
 package org.icij.datashare.model;
 
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
 import java.util.stream.Stream;
@@ -71,9 +73,12 @@ public interface TargetModel {
                 }
             }
         }
+        Set<String> reported = new HashSet<>();
         for (EntityType type : types) {
-            type.required().stream().filter(required -> isBlank(entity, required)).forEach(required ->
-                    violations.add(new Violation("type '" + type.name() + "' requires '" + required + "'")));
+            type.required().stream()
+                    .filter(required -> isBlank(entity, required) && reported.add(required))
+                    .forEach(required ->
+                            violations.add(new Violation("type '" + type.name() + "' requires '" + required + "'")));
             if (type.edge() != null) {
                 Stream.of(type.edge().source(), type.edge().target())
                         .filter(end -> !type.required().contains(end))
--- a/datashare-api/src/main/java/org/icij/datashare/model/TargetModel.java
+++ b/datashare-api/src/main/java/org/icij/datashare/model/TargetModel.java
@@ -1,10 +1,11 @@
 package org.icij.datashare.model;
 
 import java.util.ArrayList;
-import java.util.LinkedHashMap;
 import java.util.List;
-import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
+import java.util.SortedSet;
+import java.util.TreeSet;
 import java.util.stream.Stream;
 
 public interface TargetModel {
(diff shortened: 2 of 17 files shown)


datashare-api/src/main/java/org/icij/datashare/model/Statement.java
```diff
@@ -0,0 +1,57 @@
...
+        }
+        return value;
+    }
+
+    // NUL-separated rather than a printable delimiter: a cell value can hold any printable
+    // character, and every joined component is rejected at construction if it holds a NUL, so no
+    // input can forge a collision.
+    private static String id(String model, String entityId, String entityType,
+                             String property, String value, Provenance provenance) {
+        return Hasher.SHA_384.hash(String.join("\u0000", model, entityId, entityType, property, value,
```
caro3801, line 51: question: Is there any specific reason to choose sha 384?

datashare-api/src/main/resources/ftm/LICENSE
```diff
@@ -0,0 +1,22 @@
+MIT License
```
caro3801, line 1: question: why adding a licence here?

datashare-api/src/main/java/org/icij/datashare/model/TargetModel.java
```diff
@@ -0,0 +1,91 @@
...
+                    violations.add(new Violation("no property '" + property + "' on " + names));
+                } else if (declarations.stream().allMatch(Property::stub)) {
+                    violations.add(new Violation("property '" + property + "' is a stub: it is inferred from the '"
+                            + declarations.get(0).range() + "' relation rather than written"));
+                }
+            }
+        }
+        for (EntityType type : types) {
+            type.required().stream().filter(required -> isBlank(entity, required)).forEach(required ->
+                    violations.add(new Violation("type '" + type.name() + "' requires '" + required + "'")));
```
caro3801: suggestion: `validate()` loops for `(EntityType type : types)` and checks each type's own required() list independently. 

Example: if `name` happens to be required by both `LegalEntity` and `Company`, and the entity has no `name`, the loop emits two `Violations` - one per type - even though there's only one real problem ("this entity has no name"). 
Could we dedupe by the required property name so each missing fact is reported once? 

```suggestion
        Set<String> reportedRequired = new HashSet<>();
        for (EntityType type : types) {
            type.required().stream()
                    .filter(required -> isBlank(entity, required) && reportedRequired.add(required))
                    .forEach(required -> violations.add(new Violation("required property '" + required + "' is missing")));
```

(needs `import java.util.HashSet;` and `import java.util.Set;` added alongside the existing
`java.util.SortedSet`/`TreeSet` imports)

datashare-api/src/main/java/org/icij/datashare/model/ftm/FtmTargetModel.java
```diff
@@ -0,0 +1,174 @@
...
+        return Optional.ofNullable(types.get(name));
+    }
+
+    @Override
+    public String serialize(ModelEntity entity) {
+        String schema = mostSpecific(entity.types()).orElseThrow(() -> new IllegalArgumentException(
+                "types " + new TreeSet<>(entity.types()) + " have no common schema in the FtM model"));
+        try {
+            return JsonObjectMapper.getMapper()
+                    .writeValueAsString(new FtmEntity(entity.id(), schema, entity.properties()));
```
caro3801: suggestion: `serialize()` (lines 72-73) and `parse()` (line 83) call `JsonObjectMapper.getMapper()`
directly to reach `writeValueAsString`/`readValue`, but `JsonObjectMapper` already exposes
`writeValueAsString(Object)` and `readValue(String, Class)` as its intended delegate methods.

`getMapper()` exposes the shared, mutable ObjectMapper singleton (any caller holding that raw reference could reconfigure the shared mapper) - its own javadoc reserves it for initialization only, to avoid side effects, and says to prefer the delegates.
 
Could we call those delegates methods instead, so this new code doesn't bypass the one convention that already exists for it?

```suggestion
            return JsonObjectMapper.writeValueAsString(new FtmEntity(entity.id(), schema, entity.properties()));
```

(and similarly at line 83: `read = JsonObjectMapper.readValue(json, FtmEntity.class);`. The
`readTree(stream)` call at line 44 has no matching delegate, so that one is fine as-is.)

datashare-api/src/test/java/org/icij/datashare/model/ModelEntityTest.java
```diff
@@ -0,0 +1,110 @@
...
+        try {
+            ModelEntity.from(List.of(statement("name", "Jane Doe", "full_name"),
+                    Statement.of("wikidata", "person-1", "Q5", "name", "Jane Q",
+                            new Statement.Provenance("doc-1", "Sheet1", 12, "full_name"))));
+            fail("should have refused statements from two models");
+        } catch (IllegalArgumentException e) {
+            assertThat(e.getMessage()).contains("ftm");
+            assertThat(e.getMessage()).contains("wikidata");
+        }
+    }
```
caro3801, line 91: suggestion: `StatementTest.java` in factors the null-rejection idiom into a one-line `assertRejectsNull` helper, but these three tests (73-81, 83-91, 93-104) use the manual `try { ...; fail(...); } catch (X e) { assertThat(e.getMessage()).contains(...); }` shape. 

Could we reuse the same pattern here for consistency and readability?

```java
private static void assertRejectsNull(String field, Runnable construction) {
    assertThat(assertThrows(NullPointerException.class, construction::run).getMessage()).contains(field);
}

@Test
public void test_a_null_id_is_rejected() {
    assertRejectsNull("id", () -> new ModelEntity(null, Set.of("Person"), Map.of("name", List.of("Jane Doe"))));
}

@Test
public void test_null_types_are_rejected() {
    assertRejectsNull("types", () -> new ModelEntity("p-1", null, Map.of("name", List.of("Jane Doe"))));
}
```

(the third block, `test_refuses_statements_belonging_to_two_models`, checks two message fragments on an `IllegalArgumentException` and could use an analogous small helper)

datashare-api/src/test/java/org/icij/datashare/model/TargetModelRegistryTest.java
```diff
@@ -0,0 +1,49 @@
...
+        assertThat(TargetModelRegistry.get("ftm").name()).isEqualTo("ftm");
+    }
+
+    @Test
+    public void test_parses_the_model_once() {
+        assertThat(TargetModelRegistry.get("ftm")).isSameAs(TargetModelRegistry.get("ftm"));
+    }
+
+    @Test
+    public void test_an_unknown_model_names_the_known_ones() {
```
caro3801: suggestion: `test_an_unknown_model_names_the_known_ones` and `test_an_unknown_model_keeps_the_requested_name` both call `TargetModelRegistry.get("wikidata")` in a duplicated try/fail/catch just to check two aspects of the same thrown `UnknownTargetModel`. Could we merge them into one test with `assertThrows`?

```suggestion
    @Test
    public void test_an_unknown_model_names_the_known_ones_and_keeps_the_requested_name() {
        UnknownTargetModel e = assertThrows(UnknownTargetModel.class, () -> TargetModelRegistry.get("wikidata"));

        assertThat(e.getMessage()).contains("wikidata");
        assertThat(e.getMessage()).contains("ftm");
        assertThat(e.name).isEqualTo("wikidata");
    }
```

(needs `import static org.junit.Assert.assertThrows;` added; `fail` stays imported since
`test_a_null_name_is_rejected` still uses it)

datashare-api/src/test/java/org/icij/datashare/model/ftm/FtmTargetModelSerializationTest.java
```diff
@@ -0,0 +1,142 @@
...
+                Map.of("name", List.of("Jane Doe")));
+
+        List<TargetModel.Violation> violations = model.validate(confused);
+
+        assertThat(violations).hasSize(1);
+        assertThat(violations.get(0).message()).contains("no common schema");
+    }
+
+    @Test
+    public void test_serializing_types_with_no_common_schema_fails() {
```
caro3801, line 55: suggestion(minor): Seven near-identical `try { ...; fail("..."); } catch (IllegalArgumentException e) {
assertThat(e.getMessage()).contains(...); }` blocks in a row (55-64, 68-74, 78-85, 89-96, 100-106,
110-117, 121-128), six of them also repeating `assertThat(e.getMessage()).contains("FtM")`. Could we
extract a small helper, the same way `StatementTest.assertRejectsNull` does, so each test shows only
the input and the expected message fragments?

```java
private static void assertThrowsContaining(Runnable action, String... messageParts) {
    String message = assertThrows(IllegalArgumentException.class, action::run).getMessage();
    for (String part : messageParts) {
        assertThat(message).contains(part);
    }
}

@Test
public void test_unreadable_json_fails_with_a_clear_error() {
    assertThrowsContaining(() -> model.parse("{\"id\": "), "FtM");
}

@Test
public void test_a_missing_id_fails_with_a_clear_error() {
    assertThrowsContaining(() -> model.parse("{\"schema\":\"Person\"}"), "id", "FtM");
}
```

(this also drops the now-unused `org.junit.Assert.fail` import, replaced with
`org.junit.Assert.assertThrows`)

caro3801 (approved):
Strong foundation for the model contract here. The contracts read cleanly, the error handling on the read boundary
(`FtmTargetModel`'s constructor, `serialize`/`parse`) is solid: specific catches, real context in every message, no swallowed exceptions or silent fallbacks anywhere in the diff. The two production findings are both small and localized (a validation-dedup fix and reusing an existing delegate method), and the test findings are all the same one idea - lean on the `assertRejectsNull`-style helper this PR already introduced in `StatementTest`, everywhere else too.

None of these block the PR; happy to see this merged once you've had a look.

datashare-api/src/main/java/org/icij/datashare/model/Statement.java
pirhoo: Good question! No crypto reason, just consistency: `Entity.DEFAULT_DIGESTER` is already SHA-384, so document and NER entity ids all have that shape. Statement ids sit next to them, so reusing the same digest avoids introducing a second convention. SHA-256 would be perfectly fine here, it would just be the odd one out.

datashare-api/src/main/resources/ftm/LICENSE
pirhoo: Because the ontology file next to it is vendored third-party code: `defaultModel-4.10.2.json` is a verbatim copy from opensanctions/followthemoney, which is MIT. We have to bundle it since `ftm.java` is only a build-time generator and ships no model data. MIT asks that the notice travels with any copy, and Datashare is AGPL, so our own license does not cover it. Keeping it as a sibling resource means it stays with the file it applies to inside the jar. Provenance (commit sha, date, sha256) is in the `FtmTargetModel` javadoc.

datashare-api/src/main/java/org/icij/datashare/model/TargetModel.java
pirhoo: Good catch, this one bites in the common case: an entity typed {Person, LegalEntity, Thing} with no name got three violations for the same missing fact. Deduped in 6db84e4a2, with a test that was red first. I kept the existing message wording (type X requires Y) rather than the generic one, since a few tests assert the type name and it is still useful to know which schema imposed it. The type named is now the alphabetically first of the ones requiring it, which is deterministic because the types are collected from a sorted set. Left the edge-end check per-type, as edge types are concrete siblings and cannot overlap in one entity.

datashare-api/src/main/java/org/icij/datashare/model/ftm/FtmTargetModel.java
pirhoo: Fair point, no reason to bypass the convention. Done in 7dbe9bc4e. One knock-on worth noting: the readValue delegate declares IOException rather than JsonProcessingException, so parse now catches the wider one. Same message, same thrown type. And agreed on readTree, left as is.

datashare-api/src/test/java/org/icij/datashare/model/ModelEntityTest.java
pirhoo: Done in 12fc14de0. I also folded the two null tests into a single one, to match what StatementTest ended up doing, and the two-models test now uses assertThrows on the message.

datashare-api/src/test/java/org/icij/datashare/model/TargetModelRegistryTest.java
pirhoo: Merged in 12fc14de0, thanks. The null-name test uses assertThrows too now, so the fail import is gone.

datashare-api/src/test/java/org/icij/datashare/model/ftm/FtmTargetModelSerializationTest.java
pirhoo: Done in 12fc14de0: assertThrowsContaining with a varargs list of message fragments, so each test is one line now.

Issue #2358, opened 2026-08-25
fix: deleting a project leaves its statements and mappings behind


pirhoo:
Part of #2194.

Deleting a project does not remove its rows from `statement` or `extraction_mapping`. Both tables carry `prj_id` (changeset `047`, from #2202 and #2201) and nothing deletes on it, so a later team granted the same project name inherits the previous team's extracted facts and the mappings that produced them.

This is the database half of a gap #2205 already tracks for Elasticsearch: `ProjectResource.deleteProject` and `ProjectAdminServiceImpl` delete on the base name only, so `<id>.entities` survives the same way. Split out rather than folded into #2205 because #2205 is already noted as under-costed, and this half lives in `datashare-db`.

- [x] Delete a project's `statement` rows on project deletion.
- [x] Delete a project's `extraction_mapping` rows on project deletion.
- [x] Test that a project deleted and re-created under the same name sees none of the previous team's statements or mappings.

Depends on: #2201, #2202

pirhoo:
Done, merged with #2359. Deleting a project now also deletes its statements and its extraction mappings, with a test covering it. The wider sweep for records that are not keyed by project, such as tasks and batch searches, is tracked in #2372.

PR #2359, opened 2026-08-26
feat(db): add the statement store and persist the extraction mapping

### LINKED ISSUE #2202: feat: statement store

pirhoo:
Adds the `statement` and `extraction_mapping` tables with their jOOQ repositories, so extracted facts get an idempotent system of record and the mapping that produced them is persisted. Part of #2194.

- feat: add changeset 047 with the `statement` and `extraction_mapping` tables
- feat: add `StatementRepository`, upserting by statement id so a re-run is a no-op
- feat: rebuild whole entities by grouping a project's statements, streamed rather than buffered
- feat: add `ExtractionMapping`, an immutable mapping persisted as JSON and validated on save
- fix: scope both tables to their project with a composite `(id, prj_id)` primary key
- fix: purge both tables on project deletion
- test: cover both repositories on SQLite and Postgres

Requires Postgres 13 or later: `entities` streams via Incremental Sort over `statement_prj_entity`, and the pre-13 planner falls back to a full external merge sort.

Reviewers: run `make reset-db` before building. Changesets 81 and 82 were edited in place during review, so Liquibase reports a checksum mismatch against an earlier build of this branch. 047 has never shipped, so a reset is the right remedy rather than a `validCheckSum`.

### DIFF
--- a/datashare-api/src/main/java/org/icij/datashare/model/ModelEntity.java
+++ b/datashare-api/src/main/java/org/icij/datashare/model/ModelEntity.java
@@ -1,20 +1,19 @@
 package org.icij.datashare.model;
 
-import java.util.ArrayList;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.SortedSet;
+import java.util.TreeMap;
 import java.util.TreeSet;
 
-public record ModelEntity(String id, Set<String> types, Map<String, List<String>> properties) {
+public record ModelEntity(String model, String id, Set<String> types, Map<String, List<String>> properties) {
 
     public ModelEntity {
+        Objects.requireNonNull(model, "model");
         Objects.requireNonNull(id, "id");
         Objects.requireNonNull(types, "types");
         Objects.requireNonNull(properties, "properties");
@@ -24,19 +23,23 @@ public record ModelEntity(String id, Set<String> types, Map<String, List<String>
         properties = Collections.unmodifiableMap(copy);
     }
 
-    public static ModelEntity from(Collection<Statement> statements) {
-        if (statements.isEmpty()) {
-            throw new IllegalArgumentException("cannot rebuild an entity from no statement");
-        }
+    /** Folds statements into the entity they describe, one at a time: only its distinct values are
+     *  held, so a group larger than memory is not built to be collapsed. Properties and values come
+     *  out in natural order, not in the order the statements arrived, so a database collation never
+     *  decides what an entity looks like. */
+    public static ModelEntity from(Iterable<Statement> statements) {
         SortedSet<String> ids = new TreeSet<>();
         SortedSet<String> models = new TreeSet<>();
-        Set<String> types = new LinkedHashSet<>();
-        Map<String, Set<String>> values = new LinkedHashMap<>();
+        Set<String> types = new TreeSet<>();
+        Map<String, SortedSet<String>> values = new TreeMap<>();
         for (Statement statement : statements) {
             ids.add(statement.entityId());
             models.add(statement.model());
             types.add(statement.entityType());
-            values.computeIfAbsent(statement.property(), property -> new LinkedHashSet<>()).add(statement.value());
+            values.computeIfAbsent(statement.property(), property -> new TreeSet<>()).add(statement.value());
+        }
+        if (ids.isEmpty()) {
+            throw new IllegalArgumentException("cannot rebuild an entity from no statement");
         }
         if (ids.size() > 1) {
             throw new IllegalArgumentException("statements belong to " + ids.size() + " entities: " + ids);
@@ -45,7 +48,7 @@ public record ModelEntity(String id, Set<String> types, Map<String, List<String>
             throw new IllegalArgumentException("statements belong to " + models.size() + " models: " + models);
         }
         Map<String, List<String>> properties = new LinkedHashMap<>();
-        values.forEach((property, distinct) -> properties.put(property, new ArrayList<>(distinct)));
-        return new ModelEntity(ids.first(), types, properties);
+        values.forEach((property, distinct) -> properties.put(property, List.copyOf(distinct)));
+        return new ModelEntity(models.first(), ids.first(), types, properties);
     }
 }
--- a/datashare-api/src/main/java/org/icij/datashare/model/Property.java
+++ b/datashare-api/src/main/java/org/icij/datashare/model/Property.java
@@ -1,3 +1,3 @@
 package org.icij.datashare.model;
 
-public record Property(String name, String qname, String type, String range, boolean stub) { }
+public record Property(String qname, String range, boolean stub) { }
--- a/datashare-api/src/main/java/org/icij/datashare/model/Statement.java
+++ b/datashare-api/src/main/java/org/icij/datashare/model/Statement.java
@@ -34,10 +34,12 @@ public record Statement(String id, String model, String entityId, String entityT
     }
 
     /** Builds a statement for the write path, rejecting a model no {@link TargetModelRegistry} knows
-     *  before any of it reaches the store. The canonical constructor stays lenient, so a row written
-     *  under a model since retired still reads back. */
+     *  before any of it reaches the store. Leniency is a read-only promise: the canonical constructor
+     *  takes any model name, so a row written under a model since retired still reads back, but
+     *  writing that row again fails, since the store stamps every row with the model's version. */
     public static Statement of(String model, String entityId, String entityType,
                                String property, String value, Provenance provenance) {
+        Objects.requireNonNull(model, "model");
         TargetModelRegistry.get(model);
         return new Statement(id(model, entityId, entityType, property, value, provenance),
                 model, entityId, entityType, property, value, provenance);
--- a/datashare-api/src/main/java/org/icij/datashare/model/Statement.java
+++ b/datashare-api/src/main/java/org/icij/datashare/model/Statement.java
@@ -6,11 +6,19 @@ import java.util.Objects;
 
 public record Statement(String id, String model, String entityId, String entityType,
                         String property, String value, Provenance provenance) {
+    /** An entity id ends up in the (prj_id, entity_id) index, whose entries Postgres caps at 2704
+     *  bytes and SQLite does not cap at all. Bounding it here is what keeps the two dialects from
+     *  disagreeing on which mapping key a project accepts. */
+    public static final int MAX_ENTITY_ID_LENGTH = 512;
 
     public Statement {
         Objects.requireNonNull(id, "id");
         model = component(model, "model");
         entityId = component(entityId, "entityId");
+        if (entityId.length() > MAX_ENTITY_ID_LENGTH) {
+            throw new IllegalArgumentException("'entityId' is longer than " + MAX_ENTITY_ID_LENGTH
+                    + " characters: " + entityId.length());
+        }
         entityType = component(entityType, "entityType");
         property = component(property, "property");
         value = component(value, "value");
@@ -25,8 +33,12 @@ public record Statement(String id, String model, String entityId, String entityT
         }
     }
 
+    /** Builds a statement for the write path, rejecting a model no {@link TargetModelRegistry} knows
+     *  before any of it reaches the store. The canonical constructor stays lenient, so a row written
+     *  under a model since retired still reads back. */
     public static Statement of(String model, String entityId, String entityType,
                                String property, String value, Provenance provenance) {
+        TargetModelRegistry.get(model);
         return new Statement(id(model, entityId, entityType, property, value, provenance),
                 model, entityId, entityType, property, value, provenance);
     }
--- a/datashare-api/src/main/java/org/icij/datashare/model/StatementRepository.java
+++ b/datashare-api/src/main/java/org/icij/datashare/model/StatementRepository.java
@@ -7,7 +7,7 @@ import java.util.stream.Stream;
 /** Persists statements. Implementations own atomicity, as {@code ManifestRepository} does. */
 public interface StatementRepository {
     /** Upserts by statement id. An existing row keeps its first_seen and takes the new last_seen.
-     *  Returns the number of rows written. */
+     *  Returns the number of statements written. */
     int save(String projectId, String runId, Collection<Statement> statements);
 
     /** Every entity of a project, rebuilt by grouping its statements. Lazily read: the caller closes it. */
--- a/datashare-api/src/main/java/org/icij/datashare/model/StatementRepository.java
+++ b/datashare-api/src/main/java/org/icij/datashare/model/StatementRepository.java
@@ -2,16 +2,27 @@ package org.icij.datashare.model;
 
 import java.util.Collection;
 import java.util.Optional;
+import java.util.function.Function;
 import java.util.stream.Stream;
 
-/** Persists statements. Implementations own atomicity, as {@code ManifestRepository} does. */
+/** Persists statements. A save is chunked into independent transactions: a partial failure leaves a
+ *  partial, re-runnable result rather than an all-or-nothing outcome, which the upsert's idempotence
+ *  makes safe. */
 public interface StatementRepository {
-    /** Upserts by statement id. An existing row keeps its first_seen and takes the new last_seen.
-     *  Returns the number of statements written. */
+    /** Upserts by statement id. An existing row is left untouched: the write is a no-op on a re-save.
+     *  Returns the number of statements actually inserted. */
     int save(String projectId, String runId, Collection<Statement> statements);
 
-    /** Every entity of a project, rebuilt by grouping its statements. Lazily read: the caller closes it. */
+    /** Every entity of a project, rebuilt by grouping its statements. Lazily read, and backed by a
+     *  pooled connection that is released only when the returned stream is closed. The caller MUST
+     *  close it, including on every early exit (e.g. {@code findFirst}, {@code limit},
+     *  {@code anyMatch}) or exception: abandoning it without closing leaks a pooled connection.
+     *  Prefer {@link #entities(String, Function)}, which closes it for you. */
     Stream<ModelEntity> entities(String projectId);
+
+    /** Same as {@link #entities(String)}, but closes the stream for the caller once {@code consumer}
+     *  returns or throws. */
+    <R> R entities(String projectId, Function<Stream<ModelEntity>, R> consumer);
 
     Optional<ModelEntity> entity(String projectId, String entityId);
 }
--- a/datashare-api/src/main/java/org/icij/datashare/model/StatementRepository.java
+++ b/datashare-api/src/main/java/org/icij/datashare/model/StatementRepository.java
@@ -1,0 +1,17 @@
+package org.icij.datashare.model;
+
+import java.util.Collection;
+import java.util.Optional;
+import java.util.stream.Stream;
+
+/** Persists statements. Implementations own atomicity, as {@code ManifestRepository} does. */
+public interface StatementRepository {
+    /** Upserts by statement id. An existing row keeps its first_seen and takes the new last_seen.
+     *  Returns the number of rows written. */
+    int save(String projectId, String runId, Collection<Statement> statements);
+
+    /** Every entity of a project, rebuilt by grouping its statements. Lazily read: the caller closes it. */
+    Stream<ModelEntity> entities(String projectId);
+
+    Optional<ModelEntity> entity(String projectId, String entityId);
+}
--- a/datashare-api/src/main/java/org/icij/datashare/model/StatementRepository.java
+++ b/datashare-api/src/main/java/org/icij/datashare/model/StatementRepository.java
@@ -1,6 +1,5 @@
 package org.icij.datashare.model;
 
-import java.util.Collection;
 import java.util.Optional;
 import java.util.function.Function;
 import java.util.stream.Stream;
@@ -9,17 +8,20 @@ import java.util.stream.Stream;
  *  partial, re-runnable result rather than an all-or-nothing outcome, which the upsert's idempotence
  *  makes safe. */
 public interface StatementRepository {
-    /** Upserts by statement id. An existing row is left untouched: the write is a no-op on a re-save.
-     *  Returns the number of statements inserted, or fewer when the JDBC driver rewrites the batch
-     *  and reports no per-row count. */
-    int save(String projectId, String runId, Collection<Statement> statements);
+    /** Upserts by statement id, refreshing the run and the model version of a statement this run has
+     *  seen again, so a row's age tells a re-observed statement from a stale one. The statements are
+     *  consumed lazily, one chunk at a time, so a whole extraction never has to fit in memory; the
+     *  stream is closed on return. Returns the number of statements written, or fewer when the JDBC
+     *  driver rewrites the batch and reports no per-row count. */
+    int save(String projectId, String runId, Stream<Statement> statements);
 
     /** Every entity of a project, rebuilt by grouping its statements. The stream is read lazily from
      *  a pooled connection and closed once {@code consumer} returns or throws, so {@code consumer}
      *  has to consume it: returning the stream itself hands back a closed one. */
     <R> R entities(String projectId, Function<Stream<ModelEntity>, R> consumer);
 
-    /** The entity a project holds under this id. An id shared by two models yields the first by model
-     *  order, since an entity belongs to one model. */
+    /** The entity a project holds under this id. An id shared by two models yields the first model in
+     *  natural order, since an entity belongs to one model, and the entity names the model it came
+     *  from. */
     Optional<ModelEntity> entity(String projectId, String entityId);
 }
--- a/datashare-api/src/main/java/org/icij/datashare/model/StatementRepository.java
+++ b/datashare-api/src/main/java/org/icij/datashare/model/StatementRepository.java
@@ -10,19 +10,16 @@ import java.util.stream.Stream;
  *  makes safe. */
 public interface StatementRepository {
     /** Upserts by statement id. An existing row is left untouched: the write is a no-op on a re-save.
-     *  Returns the number of statements actually inserted. */
+     *  Returns the number of statements inserted, or fewer when the JDBC driver rewrites the batch
+     *  and reports no per-row count. */
     int save(String projectId, String runId, Collection<Statement> statements);
 
-    /** Every entity of a project, rebuilt by grouping its statements. Lazily read, and backed by a
-     *  pooled connection that is released only when the returned stream is closed. The caller MUST
-     *  close it, including on every early exit (e.g. {@code findFirst}, {@code limit},
-     *  {@code anyMatch}) or exception: abandoning it without closing leaks a pooled connection.
-     *  Prefer {@link #entities(String, Function)}, which closes it for you. */
-    Stream<ModelEntity> entities(String projectId);
-
-    /** Same as {@link #entities(String)}, but closes the stream for the caller once {@code consumer}
-     *  returns or throws. */
+    /** Every entity of a project, rebuilt by grouping its statements. The stream is read lazily from
+     *  a pooled connection and closed once {@code consumer} returns or throws, so {@code consumer}
+     *  has to consume it: returning the stream itself hands back a closed one. */
     <R> R entities(String projectId, Function<Stream<ModelEntity>, R> consumer);
 
+    /** The entity a project holds under this id. An id shared by two models yields the first by model
+     *  order, since an entity belongs to one model. */
     Optional<ModelEntity> entity(String projectId, String entityId);
 }
--- a/datashare-api/src/main/java/org/icij/datashare/model/TargetModelRegistry.java
+++ b/datashare-api/src/main/java/org/icij/datashare/model/TargetModelRegistry.java
@@ -4,16 +4,25 @@ import org.icij.datashare.model.ftm.FtmTargetModel;
 
 import java.util.Map;
 import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Supplier;
 
 public class TargetModelRegistry {
-    private static final Map<String, TargetModel> MODELS = Map.of("ftm", new FtmTargetModel());
+    private static final Map<String, Supplier<TargetModel>> FACTORIES = Map.of("ftm", FtmTargetModel::new);
+    private static final Map<String, TargetModel> PARSED = new ConcurrentHashMap<>();
 
+    private TargetModelRegistry() {
+    }
+
+    /** Reads a model's ontology on first use rather than in a static initializer, so a bundle that
+     *  cannot be read reaches the caller as {@link UnreadableModelResource} on every call, instead of
+     *  an ExceptionInInitializerError once and a causeless NoClassDefFoundError from then on. */
     public static TargetModel get(String name) {
         Objects.requireNonNull(name, "name");
-        TargetModel model = MODELS.get(name);
-        if (model == null) {
-            throw new UnknownTargetModel(name, MODELS.keySet());
+        Supplier<TargetModel> factory = FACTORIES.get(name);
+        if (factory == null) {
+            throw new UnknownTargetModel(name, FACTORIES.keySet());
         }
-        return model;
+        return PARSED.computeIfAbsent(name, ignored -> factory.get());
     }
 }
--- a/datashare-api/src/main/java/org/icij/datashare/model/ftm/FtmTargetModel.java
+++ b/datashare-api/src/main/java/org/icij/datashare/model/ftm/FtmTargetModel.java
@@ -42,8 +42,8 @@ public class FtmTargetModel implements TargetModel {
                 throw new UnreadableModelResource(RESOURCE);
             }
             JsonNode root = JsonObjectMapper.getMapper().readTree(stream);
-            this.version = root.get("version").asText();
-            this.types = types(root.get("schemata"));
+            this.version = present(root, "version").asText();
+            this.types = types(present(root, "schemata"));
         } catch (IOException e) {
             throw new UnreadableModelResource(RESOURCE, e);
         }
@@ -101,7 +101,7 @@ public class FtmTargetModel implements TargetModel {
                 throw new IllegalArgumentException("property '" + property + "' has a null value in FtM JSON");
             }
         });
-        return new ModelEntity(read.id(), Set.of(read.schema()), properties);
+        return new ModelEntity(name(), read.id(), Set.of(read.schema()), properties);
     }
 
     @Override
@@ -123,39 +123,51 @@ public class FtmTargetModel implements TargetModel {
     }
 
     private static Map<String, EntityType> types(JsonNode schemata) {
-        Map<String, Map<String, Property>> declared = new HashMap<>();
-        schemata.fields().forEachRemaining(schema ->
-                declared.put(schema.getKey(), properties(schema.getValue().get("properties"))));
+        Map<String, Map<String, Property>> properties = new HashMap<>();
+        Map<String, Set<String>> required = new HashMap<>();
+        schemata.properties().forEach(schema -> {
+            properties.put(schema.getKey(), properties(schema.getValue().path("properties")));
+            required.put(schema.getKey(), strings(schema.getValue().path("required")));
+        });
         Map<String, EntityType> types = new HashMap<>();
-        schemata.fields().forEachRemaining(schema ->
-                types.put(schema.getKey(), type(schema.getKey(), schema.getValue(), declared)));
+        schemata.properties().forEach(schema ->
+                types.put(schema.getKey(), type(schema.getKey(), schema.getValue(), properties, required)));
         return Map.copyOf(types);
     }
 
-    private static EntityType type(String name, JsonNode schema, Map<String, Map<String, Property>> declared) {
-        Set<String> ancestors = strings(schema.get("schemata"));
+    // The prebuilt ontology resolves neither properties nor required transitively, so both are merged
+    // over the ancestor closure here: 28 of the 69 concrete schemata require a property they inherit
+    // without redeclaring it. The type's own requirements come first so a violation list still reads
+    // in the order the schema states them.
+    private static EntityType type(String name, JsonNode schema, Map<String, Map<String, Property>> declared,
+                                   Map<String, Set<String>> declaredRequired) {
+        Set<String> ancestors = strings(present(schema, "schemata"));
         Map<String, Property> properties = new HashMap<>();
+        Set<String> required = new LinkedHashSet<>(declaredRequired.getOrDefault(name, Set.of()));
         // Two ancestors can declare the same property with different qnames: last one by name wins.
-        new TreeSet<>(ancestors).forEach(ancestor -> properties.putAll(declared.getOrDefault(ancestor, Map.of())));
+        new TreeSet<>(ancestors).forEach(ancestor -> {
+            properties.putAll(declared.getOrDefault(ancestor, Map.of()));
+            required.addAll(declaredRequired.getOrDefault(ancestor, Set.of()));
+        });
         return new EntityType(name, schema.path("abstract").asBoolean(false), ancestors,
-                Map.copyOf(properties), strings(schema.path("required")), edge(schema.path("edge")));
+                Map.copyOf(properties), Collections.unmodifiableSet(required), edge(schema.path("edge")));
     }
 
-    private static Map<String, Property> properties(JsonNode properties) {
+    private static Map<String, Property> properties(JsonNode node) {
         Map<String, Property> declared = new HashMap<>();
-        properties.fields().forEachRemaining(property ->
-                declared.put(property.getKey(), property(property.getValue())));
+        node.properties().forEach(property -> declared.put(property.getKey(), property(property.getValue())));
         return declared;
     }
 
     private static Property property(JsonNode property) {
-        return new Property(property.get("qname").asText(), text(property, "range"),
+        return new Property(present(property, "qname").asText(), property.path("range").asText(null),
                 property.path("stub").asBoolean(false));
     }
 
     private static EntityType.Edge edge(JsonNode edge) {
-        return edge.isMissingNode() ? null : new EntityType.Edge(edge.get("source").asText(),
-                edge.get("target").asText(), edge.get("directed").asBoolean(false));
+        return edge.isMissingNode() || edge.isNull() ? null
+                : new EntityType.Edge(present(edge, "source").asText(), present(edge, "target").asText(),
+                        edge.path("directed").asBoolean(false));
     }
 
     private static Set<String> strings(JsonNode array) {
@@ -164,8 +176,14 @@ public class FtmTargetModel implements TargetModel {
         return Collections.unmodifiableSet(values);
     }
 
-    private static String text(JsonNode node, String field) {
-        return node.has(field) ? node.get(field).asText() : null;
+    // Every field this parser needs is read through here rather than with get(), so a bundle whose
+    // shape moved fails as UnreadableModelResource naming the field instead of as a bare NPE.
+    private static JsonNode present(JsonNode node, String field) {
+        JsonNode value = node.path(field);
+        if (value.isMissingNode() || value.isNull()) {
+            throw new UnreadableModelResource(RESOURCE, "missing '" + field + "'");
+        }
+        return value;
     }
 
     record FtmEntity(String id, String schema, Map<String, List<String>> properties) { }
--- a/datashare-api/src/main/java/org/icij/datashare/model/ftm/FtmTargetModel.java
+++ b/datashare-api/src/main/java/org/icij/datashare/model/ftm/FtmTargetModel.java
@@ -149,8 +149,7 @@ public class FtmTargetModel implements TargetModel {
     }
 
     private static Property property(JsonNode property) {
-        return new Property(property.get("name").asText(), property.get("qname").asText(),
-                property.get("type").asText(), text(property, "range"),
+        return new Property(property.get("qname").asText(), text(property, "range"),
                 property.path("stub").asBoolean(false));
     }
 
--- a/datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
+++ b/datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
@@ -1,0 +1,75 @@
+package org.icij.datashare.tabular;
+
+import org.icij.datashare.model.TargetModel;
+import org.icij.datashare.model.TargetModelRegistry;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+public record ExtractionMapping(String id, String projectId, String userId, String name, String model,
+                                String documentId, RowSourceOptions options,
+                                Map<String, EntityMapping> entities) {
+
+    public ExtractionMapping {
+        Objects.requireNonNull(id, "id");
+        Objects.requireNonNull(projectId, "projectId");
+        Objects.requireNonNull(name, "name");
+        Objects.requireNonNull(documentId, "documentId");
+        options = options == null ? RowSourceOptions.defaults() : options;
+        Objects.requireNonNull(entities, "entities");
+        if (entities.isEmpty()) {
+            throw new IllegalArgumentException("a mapping builds no entity");
+        }
+        entities = Map.copyOf(entities);
+        TargetModelRegistry.get(Objects.requireNonNull(model, "model"));
+    }
+
+    public record EntityMapping(String type, List<String> keys, Map<String, PropertyMapping> properties) {
+        public EntityMapping {
+            Objects.requireNonNull(type, "type");
+            keys = List.copyOf(Objects.requireNonNull(keys, "keys"));
+            if (keys.isEmpty()) {
+                throw new IllegalArgumentException("entity type '" + type + "' has no key column");
+            }
+            properties = Map.copyOf(Objects.requireNonNull(properties, "properties"));
+        }
+    }
+
+    public record PropertyMapping(List<String> columns, String join, String literal, String entity, String format) {
+        public PropertyMapping {
+            columns = List.copyOf(columns == null ? List.of() : columns);
+            long sources = (columns.isEmpty() ? 0 : 1) + (literal == null ? 0 : 1) + (entity == null ? 0 : 1);
+            if (sources != 1) {
+                throw new IllegalArgumentException(
+                        "a property is filled from exactly one of columns, literal or entity, got " + sources);
+            }
+            if (join != null && columns.size() < 2) {
+                throw new IllegalArgumentException("'join' needs more than one column");
+            }
+            if (format != null && columns.isEmpty()) {
+                throw new IllegalArgumentException("'format' applies to columns only");
+            }
+        }
+    }
+
+    /** Checks every entity type and property against the target model. Called on save, not on read,
+     *  so a mapping stored before the ontology moved underneath it still loads. */
+    public List<TargetModel.Violation> validate() {
+        TargetModel target = TargetModelRegistry.get(model);
+        List<TargetModel.Violation> violations = new ArrayList<>();
+        entities.forEach((alias, entity) -> {
+            if (target.type(entity.type()).isEmpty()) {
+                violations.add(new TargetModel.Violation(
+                        "unknown type '" + entity.type() + "' for entity '" + alias + "' in model '" + model + "'"));
+                return;
+            }
+            entity.properties().keySet().stream()
+                    .filter(property -> target.property(entity.type(), property).isEmpty())
+                    .forEach(property -> violations.add(new TargetModel.Violation(
+                            "no property '" + property + "' on '" + entity.type() + "'")));
+        });
+        return violations;
+    }
+}
--- a/datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
+++ b/datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
@@ -1,5 +1,7 @@
 package org.icij.datashare.tabular;
 
+import org.icij.datashare.model.EntityType;
+import org.icij.datashare.model.Property;
 import org.icij.datashare.model.TargetModel;
 import org.icij.datashare.model.TargetModelRegistry;
 
@@ -7,6 +9,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 
 public record ExtractionMapping(String id, String projectId, String userId, String name, String model,
                                 String documentId, RowSourceOptions options,
@@ -60,15 +63,30 @@ public record ExtractionMapping(String id, String projectId, String userId, Stri
         TargetModel target = TargetModelRegistry.get(model);
         List<TargetModel.Violation> violations = new ArrayList<>();
         entities.forEach((alias, entity) -> {
-            if (target.type(entity.type()).isEmpty()) {
+            Optional<EntityType> type = target.type(entity.type());
+            if (type.isEmpty()) {
                 violations.add(new TargetModel.Violation(
                         "unknown type '" + entity.type() + "' for entity '" + alias + "' in model '" + model + "'"));
                 return;
             }
-            entity.properties().keySet().stream()
-                    .filter(property -> target.property(entity.type(), property).isEmpty())
-                    .forEach(property -> violations.add(new TargetModel.Violation(
-                            "no property '" + property + "' on '" + entity.type() + "'")));
+            if (type.get().isAbstract()) {
+                violations.add(new TargetModel.Violation(
+                        "type '" + entity.type() + "' for entity '" + alias + "' is abstract and cannot be instantiated"));
+            }
+            entity.properties().forEach((name, mapping) -> {
+                Optional<Property> property = target.property(entity.type(), name);
+                if (property.isEmpty()) {
+                    violations.add(new TargetModel.Violation("no property '" + name + "' on '" + entity.type() + "'"));
+                } else if (property.get().stub()) {
+                    violations.add(new TargetModel.Violation(
+                            "property '" + name + "' on '" + entity.type() + "' is a stub: it is inferred from the '"
+                                    + property.get().range() + "' relation rather than written"));
+                }
+                if (mapping.entity() != null && !entities.containsKey(mapping.entity())) {
+                    violations.add(new TargetModel.Violation(
+                            "property '" + name + "' on entity '" + alias + "' references unknown entity '" + mapping.entity() + "'"));
+                }
+            });
         });
         return violations;
     }
--- a/datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
+++ b/datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
@@ -20,10 +20,10 @@ public record ExtractionMapping(String id, String projectId, String userId, Stri
         Objects.requireNonNull(projectId, "projectId");
         Objects.requireNonNull(name, "name");
         Objects.requireNonNull(documentId, "documentId");
-        options = options == null ? RowSourceOptions.defaults() : options;
+        Objects.requireNonNull(options, "options");
         Objects.requireNonNull(entities, "entities");
         if (entities.isEmpty()) {
-            throw new IllegalArgumentException("a mapping builds no entity");
+            throw new EmptyExtractionMapping(id);
         }
         entities = Map.copyOf(entities);
         TargetModelRegistry.get(Objects.requireNonNull(model, "model"));
@@ -34,7 +34,7 @@ public record ExtractionMapping(String id, String projectId, String userId, Stri
             Objects.requireNonNull(type, "type");
             keys = List.copyOf(Objects.requireNonNull(keys, "keys"));
             if (keys.isEmpty()) {
-                throw new IllegalArgumentException("entity type '" + type + "' has no key column");
+                throw new InvalidEntityMapping(type);
             }
             properties = Map.copyOf(Objects.requireNonNull(properties, "properties"));
         }
@@ -45,14 +45,14 @@ public record ExtractionMapping(String id, String projectId, String userId, Stri
             columns = List.copyOf(columns == null ? List.of() : columns);
             long sources = (columns.isEmpty() ? 0 : 1) + (literal == null ? 0 : 1) + (entity == null ? 0 : 1);
             if (sources != 1) {
-                throw new IllegalArgumentException(
+                throw new InvalidPropertyMapping("sources",
                         "a property is filled from exactly one of columns, literal or entity, got " + sources);
             }
             if (join != null && columns.size() < 2) {
-                throw new IllegalArgumentException("'join' needs more than one column");
+                throw new InvalidPropertyMapping("join", "'join' needs more than one column");
             }
             if (format != null && columns.isEmpty()) {
-                throw new IllegalArgumentException("'format' applies to columns only");
+                throw new InvalidPropertyMapping("format", "'format' applies to columns only");
             }
         }
     }
--- a/datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
+++ b/datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
@@ -60,10 +60,11 @@ public record ExtractionMapping(String id, String projectId, String userId, Stri
         }
     }
 
-    /** Checks every entity type and property against the target model. Called on save, not on read,
-     *  so a mapping stored before the ontology moved underneath it still loads. Aliases and property
-     *  names are walked in sorted order, so the same mapping always reports the same violations in
-     *  the same order. */
+    /** Checks every entity type and property against the target model. The compact constructor only
+     *  checks the model is known, so this is the only complete structural check, and a writer that
+     *  skips it stores a mapping the model rejects. Called on save, not on read, so a mapping stored
+     *  before the ontology moved underneath it still loads. Aliases and property names are walked in
+     *  sorted order, so the same mapping always reports the same violations in the same order. */
     public List<TargetModel.Violation> validate() {
         TargetModel target = TargetModelRegistry.get(model);
         List<TargetModel.Violation> violations = new ArrayList<>();
@@ -95,7 +96,11 @@ public record ExtractionMapping(String id, String projectId, String userId, Stri
         if (target.type(entity.type()).isEmpty()) {
             return Optional.empty();
         }
-        String range = target.property(entity.type(), property).map(Property::range).orElse(null);
+        Optional<Property> declared = target.property(entity.type(), property);
+        if (declared.isEmpty()) {
+            return Optional.empty();
+        }
+        String range = declared.get().range();
         if (range == null) {
             return Optional.of(new TargetModel.Violation(on + "holds a value, not a reference to an entity"));
         }
--- a/datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
+++ b/datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
@@ -1,15 +1,16 @@
 package org.icij.datashare.tabular;
 
-import org.icij.datashare.model.EntityType;
-import org.icij.datashare.model.Property;
+import org.icij.datashare.model.ModelEntity;
 import org.icij.datashare.model.TargetModel;
 import org.icij.datashare.model.TargetModelRegistry;
 
 import java.util.ArrayList;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-import java.util.Optional;
+import java.util.Set;
+import java.util.TreeSet;
 
 public record ExtractionMapping(String id, String projectId, String userId, String name, String model,
                                 String documentId, RowSourceOptions options,
@@ -20,8 +21,8 @@ public record ExtractionMapping(String id, String projectId, String userId, Stri
         Objects.requireNonNull(projectId, "projectId");
         Objects.requireNonNull(name, "name");
         Objects.requireNonNull(documentId, "documentId");
-        Objects.requireNonNull(options, "options");
         Objects.requireNonNull(entities, "entities");
+        options = options == null ? RowSourceOptions.defaults() : options;
         if (entities.isEmpty()) {
             throw new EmptyExtractionMapping(id);
         }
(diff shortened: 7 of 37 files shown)


datashare-api/src/main/java/org/icij/datashare/model/Statement.java
```diff
@@ -25,8 +33,12 @@ public record Provenance(String documentId, String sheet, long rowNumber, String
         }
     }
 
+    /** Builds a statement for the write path, rejecting a model no {@link TargetModelRegistry} knows
+     *  before any of it reaches the store. The canonical constructor stays lenient, so a row written
+     *  under a model since retired still reads back. */
     public static Statement of(String model, String entityId, String entityType,
                                String property, String value, Provenance provenance) {
+        TargetModelRegistry.get(model);
```
caro3801, line 43: suggestion: Looks like a null `model` here surfaces as `NullPointerException("name")` from
`TargetModelRegistry.get`, instead of the `"model"` name every other `Statement.of` caller would see.
Could we null-check `model` before handing it to the registry, so the field name in the exception stays consistent?

```suggestion
        Objects.requireNonNull(model, "model");
        TargetModelRegistry.get(model);
```

datashare-api/src/main/java/org/icij/datashare/model/TargetModelRegistry.java
```diff
@@ -4,16 +4,25 @@
...
         Objects.requireNonNull(name, "name");
-        TargetModel model = MODELS.get(name);
-        if (model == null) {
-            throw new UnknownTargetModel(name, MODELS.keySet());
+        Supplier<TargetModel> factory = FACTORIES.get(name);
+        if (factory == null) {
+            throw new UnknownTargetModel(name, FACTORIES.keySet());
         }
-        return model;
+        return PARSED.computeIfAbsent(name, ignored -> factory.get());
```
caro3801, line 26: important: Looks like `computeIfAbsent` won't record a mapping when `factory.get()` throws, so a broken or missing ontology bundle gets re-parsed and re-thrown on every single call instead of once.

`Statement.of` calls this per statement and `JooqStatementRepository.row()` calls it again per row inside `save()`, so a bulk import against a broken bundle turns what used to be a one-time failure into hundreds of thousands of repeated parses and exceptions. 

Would it be interesting to memoize the failure too, e.g. by caching a sentinel/`Supplier` that rethrows the cached exception on subsequent calls?

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import java.util.TreeSet;
+
+public record ExtractionMapping(String id, String projectId, String userId, String name, String model,
```
caro3801, line 17: hint: `EmptyExtractionMapping`, `InvalidEntityMapping` and `InvalidPropertyMapping` are three
near-identical one-purpose `IllegalArgumentException` subclasses, and nothing in production code
catches them by type (only tests use `assertThrows(SpecificType.class, ...)`). 

Could these collapse into the single `InvalidPropertyMapping`-style exception, constructed with a formatted message at the three call sites in `ExtractionMapping.java:29,40,51,55,58`? or do you think it'll be valuable to keep them separate in the future?

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
```diff
@@ -0,0 +1,117 @@
...
+        Objects.requireNonNull(projectId, "projectId");
+        Objects.requireNonNull(name, "name");
+        Objects.requireNonNull(documentId, "documentId");
+        Objects.requireNonNull(entities, "entities");
+        options = options == null ? RowSourceOptions.defaults() : options;
+        if (entities.isEmpty()) {
+            throw new EmptyExtractionMapping(id);
+        }
+        entities = Map.copyOf(entities);
+        TargetModelRegistry.get(Objects.requireNonNull(model, "model"));
```
caro3801, line 32: hint: "Is this model known" gets checked independently at three altitudes: `Statement.of`, this
compact constructor, and again inside `ExtractionMapping.validate()`. Only `JooqExtractionMappingRepository.save()` remembers to call `validate()` today, so a future writer of a
`Statement`/`ExtractionMapping` outside that one repository would get the cheap "unknown model" guard for free but silently skip the full structural check. 

Worth a comment noting `validate()` is the only place that does the complete job, or folding the two behind one entry point?

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
```diff
@@ -0,0 +1,117 @@
...
+        if (referenced == null) {
+            return Optional.of(new TargetModel.Violation(on + "references unknown entity '" + reference + "'"));
+        }
+        if (target.type(entity.type()).isEmpty()) {
+            return Optional.empty();
+        }
+        String range = target.property(entity.type(), property).map(Property::range).orElse(null);
+        if (range == null) {
+            return Optional.of(new TargetModel.Violation(on + "holds a value, not a reference to an entity"));
+        }
```
caro3801: important: Looks like `reference()` reports "holds a value, not a reference to an entity" both 
- when a property holds a scalar 
- and when the property isn't declared on the type at all. 
=> `target.property(...)`  returns empty for both cases, so an unknown property mapped through
`entity` gets this factually wrong diagnostic on top of the correct "no property X on [type]"
violation already raised earlier in `validate()`. 

Could we short-circuit when the property isn't declared at all, since that case is already covered?

```suggestion
        Optional<Property> declared = target.property(entity.type(), property);
        if (declared.isEmpty()) {
            return Optional.empty();
        }
        String range = declared.map(Property::range).orElse(null);
        if (range == null) {
            return Optional.of(new TargetModel.Violation(on + "holds a value, not a reference to an entity"));
        }
```

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
```diff
@@ -0,0 +1,117 @@
...
+                .filter(type -> !type.name().equals(range) && !type.ancestors().contains(range))
+                .map(type -> new TargetModel.Violation(on + "needs a '" + range + "', but entity '" + reference
+                        + "' is a '" + type.name() + "'"));
+    }
+
+    // A mapping declares which properties it fills, not what they will hold, so the model's own
+    // checker runs against a probe carrying a placeholder value per mapped property. That makes the
+    // model's "required property is blank" test read as "required property is not mapped", and picks
+    // up the required-property and edge-endpoint rules a mapping-only check would miss.
+    private ModelEntity probe(String alias, EntityMapping entity) {
```
caro3801, line 117: hint: `probe()` builds a synthetic `ModelEntity` with a placeholder `"?"` per mapped property to
piggyback on `TargetModel.validate()`'s blank-check semantics, reverse-engineered two layers up. 

If `TargetModel.validate()` ever grows a format- or type-aware blank check, the placeholder could start
producing false positives/negatives silently, and the two files would need to stay in sync by
convention rather than through a shared entry point. 

Might be worth a `TargetModel` method that answers "does this shape satisfy the model" without needing real values, so this workaround isn't needed.

datashare-app/src/main/java/org/icij/datashare/tabular/JsonRowSource.java
```diff
@@ -36,6 +37,10 @@ public class JsonRowSource implements RowSource {
 
     public static final Set<String> SUPPORTED = Set.of("application/json", NDJSON_CONTENT_TYPE);
 
+    // A private mapper rather than the shared one: JsonObjectMapper's StreamReadConstraints are cut
+    // for HTTP request bodies, and on a data dump they are wrong in both directions. They raise the
+    // single-string cap from 20 MB to 1 GB, which is the guard that matters for a user-supplied file,
+    // and they lower the nesting cap from 1000 to 20, which flatten() is built to handle.
     private final ObjectMapper mapper = new ObjectMapper();
```
caro3801: crucial: The comment above this field says the private mapper raises Jackson's single-string cap from
20 MB to 1 GB, but nothing here actually calls `setStreamReadConstraints`, so it keeps Jackson's
default 20 MB cap. A data dump with one field between 20 MB and 1 GB (a long text blob, a base64
payload) would fail to import even though the comment documents it as supported. 

Using `StreamReadConstraints` Jackson could actually help apply the constraint promised in the comment

```suggestion
    private final ObjectMapper mapper = new ObjectMapper(JsonFactory.builder()            .streamReadConstraints(StreamReadConstraints.builder().maxStringLength(1_000_000_000).build())
            .build());
```

(needs `import com.fasterxml.jackson.core.JsonFactory;` and `import com.fasterxml.jackson.core.StreamReadConstraints;`)

datashare-db/src/main/java/org/icij/datashare/db/JooqExtractionMappingRepository.java
```diff
@@ -0,0 +1,114 @@
...
+    // order falls to id. That tie is broken here rather than in the ORDER BY, because a database
+    // collation would otherwise decide it, and SQLite and Postgres do not sort text alike.
+    @Override
+    public List<ExtractionMapping> list(String projectId) {
+        return create().select(EXTRACTION_MAPPING.CREATED_AT, EXTRACTION_MAPPING.ID, EXTRACTION_MAPPING.DEFINITION)
+                .from(EXTRACTION_MAPPING)
+                .where(EXTRACTION_MAPPING.PRJ_ID.eq(projectId))
+                .fetch().stream()
+                .sorted(Comparator.<Record3<LocalDateTime, String, String>, LocalDateTime>comparing(Record3::value1)
+                        .thenComparing(Record3::value2, Comparator.naturalOrder()))
```
caro3801, line 74: suggestion: The comment here only justifies keeping the `ID` tie-break out of SQL (collation mismatch
between SQLite and Postgres) - `CREATED_AT` itself is a plain timestamp with no such ambiguity, but
every row still gets fetched unordered and fully sorted in the JVM by both keys. 

Could `CREATED_AT` be pushed to `ORDER BY EXTRACTION_MAPPING.CREATED_AT` in the query, so the JVM comparator only needs to resolve the documented same-millisecond ties? 

This one grows with row count as written, so flagging it rather than inlining a suggestion since the java-side comparator would need a matching pass.

datashare-db/src/main/java/org/icij/datashare/db/JooqStatementRepository.java
```diff
@@ -0,0 +1,288 @@
...
+        return inserted(batch.execute());
+    }
+
+    private static StatementRecord row(Write write, Statement statement) {
+        StatementRecord row = new StatementRecord();
+        row.setId(statement.id());
+        row.setPrjId(write.projectId());
+        row.setRunId(write.runId());
+        row.setModel(statement.model());
+        row.setModelVersion(TargetModelRegistry.get(statement.model()).version());
```
caro3801, line 104: question/decision: `Statement.of`'s javadoc indicates the canonical constructor "stays lenient, so a row
written under a model since retired still reads back," but `row()` (used by `save()` for every
statement) unconditionally calls `TargetModelRegistry.get(statement.model()).version()`. 

If a future re-run/refresh workflow ever reconstructs statements read from the DB (via the lenient canonical constructor) and feeds them back into `save()` to refresh `RUN_ID`/`LAST_SEEN`, a retired model would throw `UnknownTargetModel` mid-chunk, aborting just that chunk while the rest of the batch succeeds.

Could we decide up front whether `save()` should honor that leniency (e.g. skip the version lookup
when a cached version is already known) or whether the javadoc promise should be narrowed to reads
only?

datashare-db/src/main/java/org/icij/datashare/db/JooqExtractionMappingRepository.java
```diff
@@ -0,0 +1,114 @@
...
+    // no @type marker to force: the plain mapper round-trips it, and it also spares Charset (e.g.
+    // sun.nio.cs.UTF_8) from the typed mapper's polymorphic deserialization, which has no
+    // string-argument constructor to satisfy.
+    private static String write(ExtractionMapping mapping) {
+        try {
+            return JsonObjectMapper.writeValueAsString(mapping);
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+    }
```
caro3801, line 105: question: `write()` and `read()` both catch `IOException` and rethrow as `UncheckedIOException` /
`UnreadableExtractionMapping`, which fits "throw early, catch late" well. 

One thing to double check: is `UncheckedIOException` on the write side (line 103) meant to reach a catch-all at the top of the call stack, or would a named exception (mirroring `UnreadableExtractionMapping` on the read side) make the write failure easier to recognize in a log without reading the stack trace? 

Not blocking, just flagging the asymmetry between the two directions.

caro3801 (changes requested):
This is a solid, carefully-documented foundational PR, most of the trickier logic (streaming cursor lifecycle, chunked upsert batching, the grouping iterator's pending/primed state machine, FTM required-property inheritance) held up well under review. 
The one thing worth fixing before merge is the default Jackson cap in [JsonRowSource.java:44](https://github.com/ICIJ/datashare/pull/2359/changes#r3882103887) , since it silently contradicts its own documented guarantee for real user data; [TargetModelRegistry.java:26](https://github.com/ICIJ/datashare/pull/2359/changes#r3882002338) and [ExtractionMapping.java:98](https://github.com/ICIJ/datashare/pull/2359/changes#r3882079879) are worth a look too since they'd surprise someone debugging a broken deployment or a rejected mapping. 
The rest are smaller cleanups that would tighten things up but don't block. 

Impressive work!

datashare-api/src/main/java/org/icij/datashare/model/Statement.java
pirhoo: Fixed in 7df268b6b: `Objects.requireNonNull(model, "model")` runs before the registry lookup, so `Statement.of(null, ...)` now names `model` like every other component. Pinned in `test_a_null_component_is_rejected_and_names_the_field`, which failed on the previous code with `<'name'> should contain the String:<'model'>`.

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
pirhoo: Fixed in 7df268b6b. `reference()` short-circuits when the property is not declared on the type at all, so the probe's `no property 'X' on [Type]` is the only violation left. New test `test_validate_reports_a_reference_on_an_undeclared_property_once`, which failed on the previous code with both violations present.

datashare-app/src/main/java/org/icij/datashare/tabular/JsonRowSource.java
pirhoo: Good catch, the comment described a cap nothing set. Fixed in 7df268b6b: the mapper builds its factory with `StreamReadConstraints.maxStringLength(JsonObjectMapper.MAX_STRING_LENGTH)`, so the 1 GB figure now comes from the constant the comment refers to, and Jackson's default nesting of 1000 stands. Comment rewritten to match the code.

New test `test_reads_a_string_larger_than_jacksons_default_cap` failed on the previous code with `String value length (20000001) exceeds the maximum allowed (20000000, from StreamReadConstraints.getMaxStringLength())`.

datashare-api/src/main/java/org/icij/datashare/model/TargetModelRegistry.java
pirhoo: Had a look and the amplification stays bounded: neither `save()` nor `saveChunk` catches per row, so the first `row()` throw kills the whole call. A broken bundle costs one failed parse per `save()`, not one per statement.

Leaving it, since caching the exception would pin every later stack trace to the first call site. Happy to add it if you would rather have the guarantee.

datashare-db/src/main/java/org/icij/datashare/db/JooqStatementRepository.java
pirhoo: Good question, and the honest answer is that the javadoc was overpromising. `save()` stamps every row with the model's current version, so a retired model has nothing to stamp and the write should fail. Narrowed the doc in 7df268b6b to say leniency is a read-only promise.

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
pirhoo: Documented it in 7df268b6b: `validate()`'s javadoc now says the constructor only checks the model is known, so `validate()` is the only complete structural check.

Kept them separate, since one entry point would mean validating on every construction, including reads, which is what the "on save, not on read" rule is there to avoid.

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
pirhoo: Leaning towards keeping them: the type name is what carries the diagnostic, and folding them into `InvalidPropertyMapping` would put that name on an entity-level error (no key column) and a mapping-level one (builds no entity). If they ever collapse it would be under a new name. Shout if you disagree!

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
pirhoo: Agreed on the coupling, but a shape-only `TargetModel` method would be an interface addition with one caller, guarding a change to `validate()` that has not happened yet. Keeping the probe with its comment for now, which is what makes it findable if that day comes.

datashare-db/src/main/java/org/icij/datashare/db/JooqExtractionMappingRepository.java
pirhoo: Fair, though `extraction_mapping` holds one row per authored mapping per project, so we are in the tens. Pushing `created_at` into SQL would put ordering in two places while the comparator still runs both keys. Keeping it in the JVM until the table gives us a reason.

datashare-db/src/main/java/org/icij/datashare/db/JooqExtractionMappingRepository.java
pirhoo: Deliberate asymmetry: `read()` parses a string that came out of the database and can genuinely fail on an older shape, hence the named type. `write()` serializes a record whose fields are fixed at compile time, so an `IOException` there is a bug rather than a runtime condition worth naming.

datashare-db/src/main/java/org/icij/datashare/db/JooqStatementRepository.java
```diff
@@ -0,0 +1,288 @@
...
+
+    // A driver that rewrites a batch into one multi-row insert (pgjdbc's reWriteBatchedInserts)
+    // reports SUCCESS_NO_INFO rather than a per-row count, so a negative entry has to contribute
+    // nothing instead of subtracting from the total.
+    private static int inserted(int[] results) {
+        return IntStream.of(results).filter(result -> result > 0).sum();
+    }
+
+    @Override
+    public Optional<ModelEntity> entity(String projectId, String entityId) {
```
caro3801, line 126: `entity()` vs `entities()` disagree on id collisions across models
Same underlying rows, two different answers: `entity()` picks the alphabetically-first model and silently drops the other model's data; `entities()` deliberately keeps both as two separate entities. A caller that lists then looks up by id gets a structurally different result for identical data, with no error in either path. 

question:  Is this divergence intentional (e.g. `entity()` is meant to resolve a single canonical model for a point lookup, while `entities()` is meant to preserve every model during a bulk scan), or should both methods converge on the same policy?

datashare-db/src/main/java/org/icij/datashare/db/JooqExtractionMappingRepository.java
```diff
@@ -0,0 +1,114 @@
...
+                .set(EXTRACTION_MAPPING.USER_ID, mapping.userId())
+                .set(EXTRACTION_MAPPING.NAME, mapping.name())
+                .set(EXTRACTION_MAPPING.DEFINITION, write(mapping))
+                .set(EXTRACTION_MAPPING.CREATED_AT,
+                        new Timestamp(DatashareTime.getInstance().currentTimeMillis()).toLocalDateTime())
+                .onConflict(EXTRACTION_MAPPING.ID, EXTRACTION_MAPPING.PRJ_ID).doNothing().execute() > 0;
+    }
+
+    @Override
+    public Optional<ExtractionMapping> get(String projectId, String id) {
```
caro3801, line 57: suggestion: `list()` vs `get()` disagree on corrupt rows
`list()` catches `UnreadableExtractionMapping`, logs, and skips - the row is invisible in every listing. `get()` has no such handling and throws uncaught for the identical row. A "list then open" flow hides the mapping, then crashes the moment it's opened directly.
Could we align `get()` on the same "hide, don't crash" policy already chosen for `list()`?

   ```suggestion
       @Override
       public Optional<ExtractionMapping> get(String projectId, String id) {
           return create().select(EXTRACTION_MAPPING.DEFINITION).from(EXTRACTION_MAPPING)
                   .where(EXTRACTION_MAPPING.ID.eq(id)).and(EXTRACTION_MAPPING.PRJ_ID.eq(projectId))
                   .fetchOptional(EXTRACTION_MAPPING.DEFINITION)
                   .flatMap(definition -> {
                       try {
                           return Optional.of(read(id, definition));
                       } catch (UnreadableExtractionMapping e) {
                           logger.warn("skipping unreadable extraction mapping '{}'", id, e);
                           return Optional.empty();
                       }
                   });
       }
   ```

datashare-db/src/main/java/org/icij/datashare/db/JooqStatementRepository.java
pirhoo: Intentional, and both sides are pinned: `test_entities_splits_an_entity_id_shared_by_two_models_into_two_entities` and `test_entity_serves_the_first_model_of_an_entity_id_shared_by_two`. `entities()` is a lossless scan, so a shared id must stay two entities (`ModelEntity.from` throws on a multi-model group). `entity()` returns a single `Optional<ModelEntity>`, so a collision has to resolve to one, and `min(model)` makes the pick deterministic rather than collation-dependent.

Nothing calls `entity()` yet, so my bet is that when a real caller appears the point lookup grows a `model` parameter rather than the two methods converging. Shout if you would rather have the divergence spelled out in the comment above `entity()` too.

datashare-db/src/main/java/org/icij/datashare/db/JooqExtractionMappingRepository.java
pirhoo: Deliberate asymmetry, pinned by `test_get_on_a_poisoned_row_names_the_failure_rather_than_disguising_it_as_io`. Aligning `get()` on empty would make a corrupt row indistinguishable from an absent one, leaving a warn line as the only trace of data corruption. `list()` degrades because one bad row should not blank the whole listing, but the point lookup is where the failure is actionable: the caller holds the id, the exception names it, and `delete()` still works to remediate.

The list-then-open flow cannot hit the crash either: the row `list()` hides never gets opened from the listing, so a throwing `get()` only fires on a direct lookup, where hiding it would be worse. Happy to align them if you still prefer one policy.

datashare-api/src/main/java/org/icij/datashare/model/ftm/FtmTargetModel.java
```diff
@@ -42,8 +42,8 @@ public FtmTargetModel() {
                 throw new UnreadableModelResource(RESOURCE);
             }
```
ClemDoum, line 43: Not for this PR but I'm discovering this class as I missed the review of https://github.com/ICIJ/datashare/pull/2354/changes#diff-a86d72db773f8d18eb7ee29ccf3233aafed32b5140a8e5d5018af50c6a86ba37

From what I understand our FtM biding was not enough to be able to implement the feature, why not using the [official java binding] instead of reimplementing the parsing/validation?

datashare-db/src/main/java/org/icij/datashare/db/JooqExtractionMappingRepository.java
caro3801: Ok thanks for pointing out to the test

datashare-db/src/main/java/org/icij/datashare/db/JooqStatementRepository.java
caro3801: It's good for me thanks for clarifying

caro3801 (approved):

datashare-api/src/main/java/org/icij/datashare/model/ftm/FtmTargetModel.java
ClemDoum: Additional comment still linked to the previous PR, I don't really understand the `target` part of the `TargetModel` naming.
The fact that it's a target is because we parsing into this schema.

This object is standalone and we'll use it to display entities and so on. I think we should get rid of the `target` prefix.

Also discussing the vocabulary together with @Soliine we though that we should use `Schema` or `Ontology` as this is the dedicate term for the domain, we'll probably `entity schema` or `data schema`.

Once again this can be left for another PR but I think it's important to get the naming right.

ClemDoum: Some other linked comments `EntityType` has very FTM specific kind of attributes with `isAbstract`, `ancestors` and `edges`.

FtM is very powerful but is also not as canonical as other semantic web ontologies / entity schema. I don't know if we'll really manage to make other schemas provides this info.  Ftm uses properties instead of proper entities + relationships (for instance for person's employer), I think we should be careful to use the most canonical data model and not the one which matches FtM the best. In the domain it seems more canonical to have entities which are linked together by relationships than related entities as properties. Said differently we should make FtM fit in the canonical entity/relationship representation and not the contrary. I think that FtM can be extracted into RDF so I think it should easily fit into the canonical entities with properties linked together by relationships with property data model.

Maybe they'll just be empty (which is fine), the question is will we use these properties, why do we need them in the model in the context of structured entity extraction ? 
Here we're just doing structured entity extraction, FtM has been conceived to do a lot more and that might be why some of these properties are inside the entity schema (if something is a Person, then it's a LegalEntity and a Thing), this might be useful for reconciliation, but I don't think it is for extraction. I think we should restrict entity type to hold only information we need in the context of entity extraction.

ClemDoum: The quantity of new object/code/logic added for this feature is huge and it can be hard to keep track of everything, I would restrict everything introduced to the minimum we actually need.

ClemDoum: Another domain naming comment, we decided a while ago (and recently spoke reassessed the validity of the domain naming recently with @Soliine) that we should use `Entity` for type of entities and add the `records` for instantiation of an entity. The `records` qualificative will be added only in context where theres ambiguity. In most contexts is obvious if we're speaking about entity types or entity record and we'll just use entity.

In the code to avoid ambiguity I think we should use `Entity` (for types) and `EntityRecord` (for records).

We can re-discuss the naming but from our reassessment with Soline this to be a valuable solution for the frontend. From a DDD perspective it should probably percolate in lower layers.

datashare-api/src/main/java/org/icij/datashare/model/ModelEntity.java
```diff
@@ -1,20 +1,19 @@
...
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.SortedSet;
+import java.util.TreeMap;
 import java.util.TreeSet;
 
-public record ModelEntity(String id, Set<String> types, Map<String, List<String>> properties) {
+public record ModelEntity(String model, String id, Set<String> types, Map<String, List<String>> properties) {
```
ClemDoum, line 13: Still because I missed the review of https://github.com/ICIJ/datashare/pull/2354/changes#diff-a86d72db773f8d18eb7ee29ccf3233aafed32b5140a8e5d5018af50c6a86ba37 I don't understand how entities could have multiple types is it a FtM corner case ?

datashare-api/src/main/java/org/icij/datashare/model/Property.java
```diff
@@ -1,3 +1,3 @@
 package org.icij.datashare.model;
 
-public record Property(String name, String qname, String type, String range, boolean stub) { }
+public record Property(String qname, String range, boolean stub) { }
```
ClemDoum, line 3: I'm not sure to understand the `qname` vs `name`, why was `name` removed and `name` kept ?

datashare-api/src/main/java/org/icij/datashare/model/StatementRepository.java
```diff
@@ -0,0 +1,27 @@
...
+     *  seen again, so a row's age tells a re-observed statement from a stale one. The statements are
+     *  consumed lazily, one chunk at a time, so a whole extraction never has to fit in memory; the
+     *  stream is closed on return. Returns the number of statements written, or fewer when the JDBC
+     *  driver rewrites the batch and reports no per-row count. */
+    int save(String projectId, String runId, Stream<Statement> statements);
+
+    /** Every entity of a project, rebuilt by grouping its statements. The stream is read lazily from
+     *  a pooled connection and closed once {@code consumer} returns or throws, so {@code consumer}
+     *  has to consume it: returning the stream itself hands back a closed one. */
+    <R> R entities(String projectId, Function<Stream<ModelEntity>, R> consumer);
```
ClemDoum, line 21: I don't understand why passing the consumer here isn't more flexible to return a stream of entity and let the caller consume it as necessary ?
Is it to handle things like transactions and so on (this way the implementer can consume the stream read for DB and the everything can be closed when reading is done ?

datashare-api/src/main/java/org/icij/datashare/model/StatementRepository.java
```diff
@@ -0,0 +1,27 @@
...
+
+    /** The entity a project holds under this id. An id shared by two models yields the first model in
+     *  natural order, since an entity belongs to one model, and the entity names the model it came
+     *  from. */
+    Optional<ModelEntity> entity(String projectId, String entityId);
```
ClemDoum, line 26: A while ago I think we chose to use exception rather returning null values for unknown db records, that's why we have `UnknownTask` in TaskRepository.getTask for instance.

That forces users to deal with unknown values. Optional is another way to deal with it probably because we need to unwrap the option. I think I like the optional implem too (because you have to explicitely unwrap it and dill with null values) but maybe it would be nice to harmonize implems

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import java.util.TreeSet;
+
+public record ExtractionMapping(String id, String projectId, String userId, String name, String model,
+                                String documentId, RowSourceOptions options,
```
ClemDoum, line 18: Do we really want to store the row source options in the mapping ?
Shouldn't we keep the parsing of the rows and the mapping between rows and properties separate ?
The `RowSourceOptions` are the options used to parse the tabular data into a collection of rows right ?

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
```diff
@@ -0,0 +1,122 @@
...
+        Objects.requireNonNull(entities, "entities");
+        options = options == null ? RowSourceOptions.defaults() : options;
+        if (entities.isEmpty()) {
+            throw new EmptyExtractionMapping(id);
+        }
+        entities = Map.copyOf(entities);
+        TargetModelRegistry.get(Objects.requireNonNull(model, "model"));
+    }
+
+    public record EntityMapping(String type, List<String> keys, Map<String, PropertyMapping> properties) {
```
ClemDoum, line 35: Are `keys` the set of columns used to deduplicate records ?
From a naming perspective I find the `ExtractionMapping` naming a bit decorrelated from the tabular data source it maps.

Maybe `primaryKeyColumns` would be more explicit (if that's what I think it is).

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
```diff
@@ -0,0 +1,122 @@
...
+        public EntityMapping {
+            Objects.requireNonNull(type, "type");
+            keys = List.copyOf(Objects.requireNonNull(keys, "keys"));
+            if (keys.isEmpty()) {
```
ClemDoum, line 39: For my empty primary key column is a valid scenario, it just means each row will is a different record.

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
```diff
@@ -0,0 +1,122 @@
...
+            }
+            properties = Map.copyOf(Objects.requireNonNull(properties, "properties"));
+        }
+    }
+
+    public record PropertyMapping(List<String> columns, String join, String literal, String entity, String format) {
+        public PropertyMapping {
+            columns = List.copyOf(columns == null ? List.of() : columns);
+            long sources = (columns.isEmpty() ? 0 : 1) + (literal == null ? 0 : 1) + (entity == null ? 0 : 1);
+            if (sources != 1) {
```
ClemDoum, line 50: Shouldn't we create different constructors I think this will be more explicit `fromColums`, `fromLiteral`, `fromEntity`

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
```diff
@@ -0,0 +1,122 @@
...
+            Objects.requireNonNull(type, "type");
+            keys = List.copyOf(Objects.requireNonNull(keys, "keys"));
+            if (keys.isEmpty()) {
+                throw new InvalidEntityMapping(type);
+            }
+            properties = Map.copyOf(Objects.requireNonNull(properties, "properties"));
+        }
+    }
+
+    public record PropertyMapping(List<String> columns, String join, String literal, String entity, String format) {
```
ClemDoum, line 46: I don't understand what is the `entity` String: columns are the tabular data row colums, literal is for a constant property value, but I don't get what entity is ?

Maybe this is an entity relationship but as stated earlier I don't think relationships between entities should be stored inside properties (we should extract entities on one side and relationships on the other). To me relationships stored as entity props is a FtM specificity and is not very canonical.

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
```diff
@@ -0,0 +1,122 @@
...
+        public PropertyMapping {
+            columns = List.copyOf(columns == null ? List.of() : columns);
+            long sources = (columns.isEmpty() ? 0 : 1) + (literal == null ? 0 : 1) + (entity == null ? 0 : 1);
+            if (sources != 1) {
+                throw new InvalidPropertyMapping(
+                        "a property is filled from exactly one of columns, literal or entity, got " + sources);
+            }
+            if (join != null && columns.size() < 2) {
```
ClemDoum, line 54: I feel like we're doing many things at once here, probably `PropertyMapping` should be a interface and then we should have several implems for literal, columns and other ways to map properties.

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
ClemDoum, line 46: If `literal` is what I think I don't think this is correct from a DDD perspective. Please check with @Soliine the name we want to use for user provided value as property. I can't fully recall what we planned to use.

ClemDoum: Maybe it was `fixed property`

ClemDoum: From a technical standpoint `literal` sounds better, but maybe we should align with what user will use / see

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
```diff
@@ -0,0 +1,122 @@
...
+                        + "' is a '" + type.name() + "'"));
+    }
+
+    // A mapping declares which properties it fills, not what they will hold, so the model's own
+    // checker runs against a probe carrying a placeholder value per mapped property. That makes the
+    // model's "required property is blank" test read as "required property is not mapped", and picks
+    // up the required-property and edge-endpoint rules a mapping-only check would miss.
+    private ModelEntity probe(String alias, EntityMapping entity) {
+        Map<String, List<String>> properties = new LinkedHashMap<>();
+        new TreeSet<>(entity.properties().keySet()).forEach(property -> properties.put(property, List.of("?")));
```
ClemDoum, line 119: Isn't there a risk if displaying wrong validation error if the validation code is evolves to display values ?

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
```diff
@@ -0,0 +1,122 @@
...
+                reference(target, alias, entity, property).ifPresent(violations::add);
+            }
+        }
+        return violations;
+    }
+
+    /** A cross-reference has to point at an entity the mapping declares, through a property that
+     *  takes entities, whose declared range the referenced type satisfies: an edge pointing at the
+     *  wrong kind of entity is invalid in the target model, not just in this mapping. */
+    private Optional<TargetModel.Violation> reference(TargetModel target, String alias, EntityMapping entity,
```
ClemDoum, line 85: If I understand correctly this if to validate properties of entities which are actual entities.
I don't think this is the way to go for non FtM entity schema.

I feel like in other schema / ontologies we have entities on one side and relationships on the other. To me references replace relationships which is not a good thing in the gereric/canonical case.

Let's say a table has 1 person per row with their employer  listed in columns, the user will define 2 entities (Person, Company), then they will add the `Employer` relationship.

This is the generic case, then in the FtM subcase, we can use relationships to populate the property (and not the other way around).

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMappingRepository.java
```diff
@@ -0,0 +1,23 @@
...
+import java.util.List;
+import java.util.Optional;
+
+/** Persists extraction mappings. Implementations own atomicity, as {@code ManifestRepository} does. */
+public interface ExtractionMappingRepository {
+    /** Saves a new mapping. Returns false when the project already holds that id: a mapping is
+     *  immutable, so a change is a new mapping with a new id. Ids are project-scoped, so two projects
+     *  can each hold one under the same id.
+     *  @throws InvalidExtractionMapping when the mapping does not validate against its target model. */
+    boolean save(ExtractionMapping mapping);
```
ClemDoum, line 12: What happens if the user provides a different mapping with the same mapping. 
If mapping are immutable maybe this should throw a `MappingAlreadyExists` exception

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMappingRepository.java
```diff
@@ -0,0 +1,23 @@
...
+
+    /** @throws UnreadableExtractionMapping when the stored row exists but its definition no longer
+     *  parses, e.g. against a model the registry no longer knows. A type or property the model has
+     *  since dropped still parses: {@link ExtractionMapping#validate()} runs on save, not on read. */
+    Optional<ExtractionMapping> get(String projectId, String id);
```
ClemDoum, line 17: Same comment about harmonization between throwing vs optional on unknown ID

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMappingRepository.java
```diff
@@ -0,0 +1,23 @@
...
+
+    /** Skips a row it cannot read rather than failing the whole list. */
+    List<ExtractionMapping> list(String projectId);
```
ClemDoum, line 20: Why not streaming ?

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMappingRepository.java
```diff
@@ -0,0 +1,23 @@
...
+     *  can each hold one under the same id.
+     *  @throws InvalidExtractionMapping when the mapping does not validate against its target model. */
+    boolean save(ExtractionMapping mapping);
+
+    /** @throws UnreadableExtractionMapping when the stored row exists but its definition no longer
+     *  parses, e.g. against a model the registry no longer knows. A type or property the model has
+     *  since dropped still parses: {@link ExtractionMapping#validate()} runs on save, not on read. */
+    Optional<ExtractionMapping> get(String projectId, String id);
+
+    /** Skips a row it cannot read rather than failing the whole list. */
```
ClemDoum, line 19: I don't understand his comment ?

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMappingRepository.java
```diff
@@ -0,0 +1,23 @@
...
+    List<ExtractionMapping> list(String projectId);
+
+    boolean delete(String projectId, String id);
```
ClemDoum, line 22: for the task repository we have `delete(String taskId) throws IOException, UnknownTask`

datashare-db/src/main/java/org/icij/datashare/db/JooqRepository.java
```diff
@@ -411,8 +413,12 @@ public boolean deleteAll(String projectId) {
             int deleteUserRecommendationResult = inner.deleteFrom(DOCUMENT_USER_RECOMMENDATION).where(DOCUMENT_USER_RECOMMENDATION.PRJ_ID.eq(projectId)).execute();
             List<Integer> deletedUserHistoryProjectIds = inner.deleteFrom(USER_HISTORY_PROJECT).where(USER_HISTORY_PROJECT.PRJ_ID.eq(projectId)).returning().fetch().getValues(USER_HISTORY_PROJECT.USER_HISTORY_ID);
             int deleteUserHistoryResult = inner.deleteFrom(USER_HISTORY).where(USER_HISTORY.ID.in(deletedUserHistoryProjectIds)).execute();
+            int deleteStatementResult = inner.deleteFrom(STATEMENT).where(STATEMENT.PRJ_ID.eq(projectId)).execute();
```
ClemDoum, line 416: Wow I didn't know about this one I think tasks are other objects are not deleted correctly

datashare-db/src/main/java/org/icij/datashare/db/JooqStatementRepository.java
```diff
@@ -0,0 +1,288 @@
...
+    // driver buffers the whole result before the first row is emitted. So Postgres holds one
+    // connection, outside the pool's default autocommit, for the stream's lifetime. SQLite ignores
+    // fetchSize and has no server-side cursor, and a read it keeps open holds a shared lock on the
+    // whole file until the consumer returns, which makes every concurrent writer fail "database is
+    // locked": the embedded database buffers instead.
+    private Stream<ModelEntity> entities(String projectId) {
+        if (dialect != SQLDialect.POSTGRES) {
+            return group(read(create(), projectId).fetch(JooqStatementRepository::toStatement).stream());
+        }
+        Connection connection = openStreamingConnection();
```
ClemDoum, line 155: we stream from the DB in other implems, if this is needed here, this is probably needed elsewhere too, we should harmonize, the question is do we really need it (I would have expected crashes for other repository implems ) ?

datashare-db/src/main/java/org/icij/datashare/db/JooqStatementRepository.java
```diff
@@ -0,0 +1,288 @@
...
+                new Statement.Provenance(row.get(STATEMENT.DOC_ID), row.get(STATEMENT.SHEET),
+                        row.get(STATEMENT.ROW_NUMBER), row.get(STATEMENT.COLUMN_NAME)));
+    }
+
+    // The query is ordered by (entity_id, model), so an entity's statements are consecutive and each
+    // one is folded into its entity as it is read: one CSV can produce 100k+ entities, and a mapping
+    // keyed on a low-cardinality column puts every one of its rows under a single entity, so neither
+    // the project nor one group is ever collected first. Grouping on entity_id alone would let a
+    // same-id entity spanning two models reach ModelEntity.from, which throws on a multi-model group.
+    private static Stream<ModelEntity> group(Stream<Statement> statements) {
```
ClemDoum, line 225: This expects statements are correctly ordered first and will yield weird results otherwise ?
Shouldn't we make sure either that everything is sorted either by sorting at the same time as grouping or by raising if we seen discontinuity in the stream

datashare-api/src/main/java/org/icij/datashare/model/ftm/FtmTargetModel.java
pirhoo: > From what I understand our FtM biding was not enough to be able to implement the feature, why not using the official java binding instead of reimplementing the parsing/validation?

ftm.java is a build-time code generator: one Java class per schema, and no model data at runtime. Mapping validation needs the opposite, runtime lookup by string name ("does Person have employer, what's its range") plus a violations list, which generated static classes can't answer. That's why the prebuilt ontology JSON is vendored and pinned by sha (see the class javadoc). If ftm.java grows a runtime model, it can slot in behind `TargetModel`.

pirhoo: > I don't really understand the `target` part of the `TargetModel` naming. [...] we though that we should use `Schema` or `Ontology` as this is the dedicate term for the domain

No attachment to `Target`, happy to rename. One flag: `Schema` is already loaded here (DB schema, ES schema), while `Ontology` collides with nothing. Tracked in #2370 as one rename PR, best done before more code stacks on these names. Which term do you and @Soliine pick?

pirhoo: > `EntityType` has very FTM specific kind of attributes with `isAbstract`, `ancestors` and `edges`. [...] I think we should restrict entity type to hold only information we need in the context of entity extraction.

These three are exactly the subset the validation consumes, the rest of FtM's schema (labels, matchable, featured...) was left out. `isAbstract` rejects mapping a bare abstract type (TargetModel.java:58), `ancestors` is what lets a `Company` satisfy a property whose range is `LegalEntity` (ExtractionMapping.java:108), and `edge` rejects a relationship mapped without both endpoints (TargetModel.java:82). A model without these concepts returns false/empty/null and the checks no-op, so they don't force FtM's shape onto other ontologies.

pirhoo: > The quantity of new object/code/logic added for this feature is huge and it can be hard to keep track of everything, I would restrict everything introduced to the minimum we actually need.

Fair, and it's why the review waves already deleted the seams nothing used (7768f40e3). My rule for the stack: anything without a consumer by the end of #2194 gets deleted rather than kept warm. Call out anything you already see as dead and it goes now.

pirhoo: > we decided a while ago [...] that we should use `Entity` for type of entities and add the `records` for instantiation of an entity. [...] In the code to avoid ambiguity I think we should use `Entity` (for types) and `EntityRecord` (for records).

Works for me, `ModelEntity` -> `EntityRecord` is mechanical. One heads-up: `EntityType` -> `Entity` collides with the existing `org.icij.datashare.Entity` interface that `Document` and `NamedEntity` implement, so either that one moves too or the type keeps a qualifier. Added to #2370 with the Ontology rename.

datashare-api/src/main/java/org/icij/datashare/model/ModelEntity.java
pirhoo: It comes from statements being the system of record rather than from FtM: each statement names its own entity type, and two mappings (or one revised between runs) can type the same id differently. `from()` keeps the union so the rebuild is lossless; the FtM model then flags a set with no common schema at validation and collapses to the most specific type at serialization. With a single mapping it's a singleton set in practice.

datashare-api/src/main/java/org/icij/datashare/model/Property.java
pirhoo: `name` restated the map key (properties live in a map keyed by name on `EntityType`), so it went in the dead-seam sweep (7768f40e3). `qname` is FtM's qualified form ("Person:name"), kept for the FtM statement wire format, but honestly nothing consumes it yet either: tracked in #2371 to drop it or document what it waits for.

datashare-api/src/main/java/org/icij/datashare/model/StatementRepository.java
pirhoo: Exactly what you guessed: the stream reads lazily from a pooled connection, and the consumer scope is what lets the repository close the cursor and return the connection deterministically when the consumer returns or throws. Returning the Stream would either hand back an already-closed one or make every caller responsible for close-or-leak.

datashare-api/src/main/java/org/icij/datashare/model/StatementRepository.java
pirhoo: I read `UnknownTask` as "this id came from the system, absence is a bug", while here absence is a normal answer to a user-supplied id, which is what Optional models. Corrupt rows still throw a named `UnreadableExtractionMapping`, so absent and broken stay distinct. Up for one convention across repositories, but as its own pass.

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
pirhoo: They stay separate types (the readers never see mappings), they're just stored together: a mapping is the reproducible recipe for "this document into these entities", and the delimiter or sheet decides what a row even is. Splitting the storage would mean versioning two objects in lockstep to re-run one extraction.

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
pirhoo: Exactly that: the key columns build the entity id, so rows sharing their values collapse into one entity. `primaryKeyColumns` says it better, added to the naming PR (#2370).

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
pirhoo: Agreed it's a legitimate mapping. The guard exists because re-runs are upserts by entity id: with no key columns the id must come from something stable, and a row index isn't (insert one row at the top of the CSV and every entity below changes identity). #2204 owns entity id determinism, so I'd lift the guard there with a real row-identity scheme; noted on the issue.

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
pirhoo: Nothing builds mappings from Java yet, they arrive as JSON, so the factories would have no caller today. The day something does, `fromColumns`/`fromLiteral`/`fromEntity` are five lines and I'm happy to add them, or now if you want the reading-side clarity already.

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
pirhoo: It's a cross-reference: the property is filled with the id of another entity declared in the same mapping, named by its alias. That's how an edge names its endpoints, e.g. an `Employment` whose `employer`/`employee` reference the `Person` and `Company` aliases. More on the relationships point in the thread on `reference()`.

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
pirhoo: The record is the stored JSON shape, and the constructor enforces exactly-one-source in four lines, so a sealed interface with three implementations would mostly buy a Jackson type-discriminator in the wire format for behavior we already have. If a Java-side builder shows up I'd revisit.

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
pirhoo: `literal` was just the technical pick, no attachment. Whatever term you settled on with @Soliine ("fixed property"?) should land before a UI stores mappings, since the field name is the wire format. It's on #2370 waiting for the confirmed term.

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
pirhoo: Same coupling @caro3801 flagged above: acknowledged, and the comment on `probe()` keeps the two files findable together. A shape-only `TargetModel` entry point is one interface addition away, I'd add it the day `validate()` grows a value-aware check.

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
pirhoo: That's actually how it already works: the person-with-employer-column table maps as three entities, `Person`, `Company`, and an `Employment` edge whose endpoint properties reference the two aliases. Relationships are first-class (the edge types), `entity` is how an edge names its endpoints, and `reference()` validates them against the declared range. An ontology with native relationships fits the same shape: a relationship instance references what it links.

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMappingRepository.java
pirhoo: Today it returns false, documented as "a mapping is immutable, a change is a new mapping with a new id", and the boolean matches `Repository.save(Project)`. An exception carries more for a 409 though, so I'd decide it in the PR that adds the REST endpoint, where the caller exists. Tracked in #2371.

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMappingRepository.java
pirhoo: Same answer as on `StatementRepository.entity()`: absence of a user-supplied id is a normal outcome, which Optional models, and `UnreadableExtractionMapping` keeps corruption distinct from absence. One convention across repositories works for me as its own pass.

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMappingRepository.java
pirhoo: One row per authored mapping per project, so tens at most, a screenful. Streaming would need the consumer-scoped connection plumbing `entities()` has, with no memory to save. Statements stream because they come in millions.

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMappingRepository.java
pirhoo: It means one poisoned row must not blank the whole listing: `list()` logs and skips a row whose stored definition no longer parses, where `get()` on the same row throws. Rewording the javadoc to say exactly that is on #2371.

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMappingRepository.java
pirhoo: Same family as the Optional question: `delete` returning false makes the call idempotent (a retry after a timeout doesn't blow up), where `UnknownTask` treats a missing id as a bug. I'd fold it into the same harmonization pass so every repository picks one convention, rather than align this one on tasks now.

datashare-db/src/main/java/org/icij/datashare/db/JooqRepository.java
pirhoo: This PR added `statement` and `extraction_mapping` to that cascade (that was #2358). Tasks are a pre-existing gap the cascade can't fix by key: the `task` table has no `prj_id`, the project only lives inside the `args` json. Opened #2372 to audit what project deletion leaves behind (tasks, batch searches, anything else project-shaped but not project-keyed).

datashare-db/src/main/java/org/icij/datashare/db/JooqStatementRepository.java
pirhoo: Statements are the one genuinely unbounded read: rows = entities x properties x values, one CSV can mint millions, and `entities()` scans a whole project. The other repositories fetch bounded sets (a document's tags, a page of history), so they never crash, but the subtle part is that their `.stream()` calls buffer anyway: pgjdbc only streams with fetchSize > 0 AND autoCommit off, which is exactly the dance this class does. So nothing to harmonize until another unbounded read appears, and then this is the recipe to copy.

datashare-db/src/main/java/org/icij/datashare/db/JooqStatementRepository.java
pirhoo: The order is established in this same file: `read()` puts `ORDER BY entity_id, model` on the only query that feeds `group()` (line 172), so the precondition is structural rather than trusted input. A Java-side monotonicity guard would compare with `String.compareTo` while the database sorted with its own collation, and the two disagree (SQLite and Postgres don't sort text alike, which is why `entity()` avoids ORDER BY for its model pick), so it would raise false alarms on correct streams. And if the invariant ever broke, adjacent grouping degrades to emitting the same id twice; it never merges two different entities.

datashare-api/src/main/java/org/icij/datashare/model/ftm/FtmTargetModel.java
ClemDoum: @pirhoo for FtM I'm not speaking about Bruno's binding but about the official one which is meant to be used to expose model and perform validation (from what I understood). See here https://javadoc.io/doc/tech.followthemoney/followthemoney/latest/tech/followthemoney/model/package-summary.html and https://followthemoney.tech/libraries/

ClemDoum: For the naming I guess `EntitySchema` would be a nice middle ground to avoid collision and keep the naming aligned from the back to the front.

ClemDoum: @pirhoo about `isAbstract` and `ancestors`, I think the fact FtM supports inheritance in an exception rather than the norm for ontology/data schema.

The questions is why do we need to check inheritance related stuff when parsing entities ?
Inheritance is useful for later processing (or inside the FtM validation if it's really needed) but I don't think we need to introduce the concept on the whole codebase. For me it's added complexity that we don't use I think using the official tech.followthemoney/followthemoney will provide the validation we need about not trying to instanciate an abstract entity etc...

ClemDoum: Thanks for the ModelEntity -> EntityRecord renaming you're right about `EntityType` I don't think we have a choice.

datashare-api/src/main/java/org/icij/datashare/model/ModelEntity.java
ClemDoum: I think there's something I don't get. If entity schemas don't allow the same entity to have different entity types, why would we allow 2 mapping to type differently the same entity ID. We don't we create a new entity, it it has. a different type then it can be the same entity (or let's override the old type).

I mean maybe for parsing we could allow to have multiple type but then I see that this percolated to `StructuredEntity` in #2361 were entity that we index also have multiple types (and I think no model allows it)

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
ClemDoum: That works for me, thanks for the explanation

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
ClemDoum: You're right we can add them later

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
ClemDoum: Honestly I'd rather do it now, it's hard to keep track of this kind of choices for other implementers/devs or implementers/devs which are not as careful/aware when updating a component.

ClemDoum: I mean there's some documentation in the comments on this file but it's hard to keep track of context comments in other files. I prefer code to behave as expected by function/property/class naming.

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
ClemDoum: Why not creating a dedicate `Relationship` object/store etc. I find it a lot more explicit to model rather than having entity only + references. 

In the user flow, user will:
- define entities
- then potentially define relationships linking them

I think I really prefer to model relationships as a standalone object as you just have to retrieve the relationship to get the 2 links objects. Otherwise you need to keep 2 references.

Say company C is the employer of employee E, then for now we stored employee(id: E, employer: C). If we want to know all employees of C then we have to go through all entities and check if they reference E.  I feel like this is made easier with relationships.

But maybe that's something we won't need to do in the index, and we'll have to create a proper graph (neo4j) to do this kind of things.

For now I feel like explicitly extracting relationships models closer what the user is doing + it might be better suite for downstream usages. The second point is conditional, I don't know for sure I think we should discuss usages and see what's the best option.

That's a very structural point, we shouldn't the pick a FtM-like choice because FtM is the first entity schema we implement. Some data models/format like RDF which are supposed to represent arbitrary ontologies/Schema are made different choices. I just think we should think about it twice when it comes to not explicitly reprenseting relationships

datashare-api/src/main/java/org/icij/datashare/model/ftm/FtmTargetModel.java
pirhoo: > for FtM I'm not speaking about Bruno's binding but about the official one which is meant to be used to expose model and perform validation

My earlier answer was about the wrong library, I read "official binding" as Bruno's `org.icij:ftm.java`. You're right: `tech.followthemoney:followthemoney` is a real runtime model, I pulled 4.10.2 from Maven Central to check. `Model.loadDefault()` bundles the ontology, and `Schema.isA`/`getEdge`/`getRequiredProperties` cover what `FtmTargetModel` hand-parses today, so it can replace the vendored JSON and the parsing. Two caveats from the jar: no validation API (no violations, not even the `abstract` flag), so `validate()` stays ours, and `Model` drops the JSON's `version` field, which we stamp on every statement row, so that stays a one-line read on our side. It even helps multi-version support: `Model.fromJson` is public and each instance is self-contained, though the jar only bundles its own version, so older ontologies stay vendored JSON. Opened #2374 for the swap.

pirhoo: > For the naming I guess `EntitySchema` would be a nice middle ground to avoid collision and keep the naming aligned from the back to the front.

`EntitySchema` works for me: it avoids both collisions and it's FtM's own vocabulary (they call `Person` a schema). Noted on #2370. That leaves the container currently called `TargetModel`: `Ontology`, or `Model` to align with the official lib?

pirhoo: > The questions is why do we need to check inheritance related stuff when parsing entities ? [...] For me it's added complexity that we don't use

We never check inheritance while parsing entities, it's consumed in exactly two places, both at validation time: range satisfaction, where a mapping filling `Employment.employer` (range `LegalEntity`) with a `Company` must pass and would be wrongly rejected without `isA`, and picking the most specific type at serialization. FtM's own ontology relies on inheritance for its ranges, so mappings can't validate correctly without it. And a heads-up from checking the official lib's jar: it keeps inheritance first-class too (`Schema.isA`, `getExtends`) and ships no validation, no abstract-instantiation check included, so adopting it outsources the parsing, not the concept.

datashare-api/src/main/java/org/icij/datashare/model/ModelEntity.java
pirhoo: I agree the end state is yours: a differently-typed entity should just be a different entity. The union at read time isn't an endorsement of multi-type, it's the rebuild refusing to silently drop or overwrite statements another mapping wrote. The structural fix belongs to id minting: #2204 derives the entity id deterministically, and folding the type into it means two types can never share an id. Then `types` is a singleton by construction and #2361's set collapses to a single field. Noted on #2204 next to the row-identity point.

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
pirhoo: Doing it now then: `TargetModel` grows a shape-level entry point ("are these mapped properties a valid shape for this type") and `probe()` with its placeholder trick goes away. Added to #2371.

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
pirhoo: One correction on the example: with the mapping as it is, employer(C)/employee(E) live on a third mapped entity, an `Employment` edge, not on the employee, so "all employees of C" is one indexed query on `Employment.employer = C`, no scan. That edge entity is already a standalone relationship object: own id, own statements, own index document. So the open question is narrower than it looks: whether the domain model names it `Relationship` instead of "an entity whose schema is an edge", which matters for non-FtM ontologies and matches the flow you describe (define entities, then relationships). Agreed we shouldn't default into a structural choice because FtM came first: let's put it on the #2194 agenda with @Soliine, I can write up the two options with the query patterns we expect.

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMappingRepository.java
ClemDoum: that works for me perfect thanks

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMappingRepository.java
ClemDoum: Sounds reasonable thanks !

datashare-db/src/main/java/org/icij/datashare/db/JooqRepository.java
ClemDoum: Great thank you !

datashare-db/src/main/java/org/icij/datashare/db/JooqStatementRepository.java
ClemDoum: I think we have the same issue with tasks for instance `JooqTaskRepository`. There can be a lot of them in the DB and they can be very large objects.

datashare-db/src/main/java/org/icij/datashare/db/JooqStatementRepository.java
ClemDoum: I wasn't thinking about a java side order but rather a sql side. I think separating the logic creates an implicit rule that implement should call the sort before grouping.

This is something to keep in mind when using the group function. I find it error prone for non AI implementers which can't keep as much context in mind and are also not as good as spotting corner cases.

Why not calling sort in the group function ? Instead of taking a stream of statement the group function could take a sql statement returning a group of statement an systematically call the sort on the sql side on the top of the sql statement. It's just an idea, but I'm sure we can find ways to never call group on unsorted stream while keeping everything sql side.

datashare-api/src/main/java/org/icij/datashare/model/ftm/FtmTargetModel.java
ClemDoum: Sorry I was speaking about validation and not parsing.

I understand your point about the jar not providing checks for abstract things but I still think the check should be implemented as an extra validation on ftm validation on the top of what the jar does and shouldn't propagate the whole data modeling for all ontologies. I'd keep it there for now and add these attribute to the data model if we see other ontologies support this feature. 

I think the real question is will we really need inheritance for downstream applications inside DS ?

From an entity resolution/reconciliation I'm tempted to say no. It's likely that reconcilation logics will be ontology / schema specific so we'll benefit from each schema tools to get parents/ancestors (that will likely not happen in Java btw).

Maybe the exception I see is if we want to blindly export a neo4j graph from entities coming from any schema. Having ancestors in the index will help to add multiple labels on node `:Person:LegalEntity:Thing(name: "clement")` without having to implement schema specific logic when exporting to neo4j.

So I guess my final call is let's keep them !

datashare-api/src/main/java/org/icij/datashare/model/ModelEntity.java
ClemDoum: Perfect thank you !

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
ClemDoum: Soline's answer is `fixed property`, maybe we could use `fixedValue` or just `fixed`

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
ClemDoum: cool thank you !

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
ClemDoum: OK understood, I missed the point you explained. Thanks for adding the discussion to the issues !

Issue #2207, opened 2026-06-16
feat: REST and CLI trigger


pirhoo:
Part of #2194.

Let users start an extraction.

- [ ] Add a REST endpoint that saves a mapping and launches the task, reporting status like other tasks.
- [ ] Add a command-line command to run a mapping file without the UI. Note (2026-08-19) that the CLI is dual-surface: the JOpt options in `DatashareCliOptions` **and** the picocli command tree in `datashare-cli/.../cli/command/`. A new trigger has to be wired on both, plus a `Stage` enum entry and its line in `StageRunCommand`'s help text.
- [ ] Test both entry points.

Depends on: #2206, #2201

## Added 2026-08-25, when #2201 was designed

- **This issue owns the mapping authorization decision.** `extraction_mapping` carries a nullable `user_id` (the author; null means a CLI run as `nullUser()`), but the repository is deliberately user-blind. #2201 and #2202 ship no HTTP surface, and authorization designed with no endpoint to protect is how it gets designed wrong. The endpoint lands here, so this is where owner-only vs owner-or-published vs project-wide gets decided. `JooqBatchSearchRepository` is the local precedent for owner-or-published (`USER_ID.eq(user.id).or(PUBLISHED.greaterThan(0))`, `:208`), and `published` would be a new column if that is the choice.
- **Mappings are immutable.** `save` refuses an existing id, so the endpoint reports the conflict rather than overwriting, and a change is a new mapping with a new id. The CLI command needs the same behaviour.

PR #2361, opened 2026-08-26
feat: <project>.entities index and ExtractedEntity type

### LINKED ISSUE #2205: feat: <project>.entities index and ExtractedEntity type
Part of #2194.

Make extracted entities searchable in their own per-project index, separate from the documents index. They get their own index because the same entity can come from many documents, which a parent/child link to a single document cannot represent.

- [x] Add an entity type the indexer can store.
- [x] Add a mappings file for the entities index that allows multiple, namespaced properties.
- [x] Carry the entity's **data schema and its type(s)** as fields on the indexed document, so entities from different schemas coexist in one index and can be filtered apart. The namespaced-property convention is what keeps the field mapping from exploding when a second schema is added (see the mapping-isolation rationale in the memo's §6).
- [x] Create the index through the existing Elasticsearch setup code, which today needs a new seam (see below).
- [x] Rebuild the index from the statement store. The index is a disposable copy; the database is the source of truth.
- [ ] Make backups, snapshots and the client aware of this extra index.

## Added 2026-08-19, after checking the existing code

Two facts about what is already there:

- `ElasticsearchConfiguration.createIndex` (`:118`) does **not** accept a mappings file. It always calls `createIndexBody()` (`:135`), which reads the single `datashare_index_mappings.json`. So "create the index through the existing setup code" includes building a mappings-per-index seam.
- That call must stay on its raw-REST path (`new Request("PUT", "/" + indexName)` + `setJsonEntity`) rather than move to the typed `client.indices().create(...)` builder. The typed client is what sends `match_mapping_type` as an array, which broke index creation against production Elasticsearch in July, and this index needs exactly the dynamic templates that trigger it.

Four items deferred out of #2199 / #2311, none of them fixed there:

- [x] Deleting a project never removes `<id>.entities`. `ProjectResource.deleteProject` and `ProjectAdminServiceImpl` both delete on the base name only, so a later team granted the same project name inherits the previous team's entity data.
- [ ] No migration or startup check for a project row already named with a dot or a leading underscore. #2311's guards only run on creation.
- [ ] The `.entities` suffix rule deliberately bypasses `ForbiddenException.requireGranted`, the exact-match gate used by the project-scoped call sites, so a `.entities` hit can be searched but not opened.
- [x] Nothing in the repo produces a `.entities` index yet, so until this issue lands the authorization rule can only match an index Datashare did not create.

Depends on: #2202, #2199

## Added 2026-08-25, when #2202 was designed

- **Carry `model_version` on the indexed entity**, alongside the data schema and its type(s). `statement` stores the ontology version (e.g. `4.10.2`) per fact, so the projection can carry it and entities extracted under different ontology versions stay distinguishable within one index.
- **The project-deletion gap now has a database half**, tracked in #2358: `statement` and `extraction_mapping` both carry `prj_id` and nothing deletes on it, exactly as nothing removes `<id>.entities` today.
- **Properties are stored namespaced** in `statement` (`ftm:birthDate`, from `Statement.qualifiedProperty()`), so the projection reads the wire form directly instead of re-qualifying it.
- **An entity can carry several types.** `ModelEntity.from` unions `entityType` across a group (`ModelEntity.java:38`), so the indexed document's type field is a list, not a single most-specific schema.

pirhoo:
Projects a project's extracted entities from the statement store into a searchable `<project>.entities` index, created next to the documents index and deleted with it. Part of #2194, closes #2205.

**Nothing triggers indexing in this PR.** `EntitiesIndexRebuilder` is inert plumbing with no task, no endpoint and no Guice wiring: #2206 wraps it in a task, #2207 exposes it via REST and CLI. Until those land, the entities index is created but stays empty. Stacked on #2359.

- feat: carry the model, its versions and the source documents on `ModelEntity`
- feat: rebuild an entity with its `model_version` and source documents in `JooqStatementRepository`
- feat: add `datashare_entities_index_mappings.json` and its settings, plus a mappings-per-index seam on `ElasticsearchConfiguration.createIndex`, still on the raw REST path
- feat: add `ExtractedEntity`, the indexable projection, namespacing the property keys back to their wire form
- feat: add `EntitiesIndexRebuilder`, purging then refilling the index from the statement stream in chunks
- feat: create the entities index with the project, and lazily on rebuild for projects that predate it
- fix: delete a project's entities index with the project, contained so a failure there cannot mask the documents result
- fix: fail a rebuild whose bulk write is rejected, rather than reporting a count it did not write
- fix: create an entities index with the entity mappings over `PUT /api/index/:index`
- fix: turn off date detection, so a date-shaped value cannot mint a `date` mapping that wedges every later rebuild
- refactor: name the entities index in one place, `Project.entitiesIndex`

Requires `make reset-db` if your build database carries #2204's `original_value` column: jOOQ codegen reads the live schema and this branch's changeset 047 does not define it.

### DIFF
--- a/datashare-api/src/main/java/org/icij/datashare/text/ExtractedEntity.java
+++ b/datashare-api/src/main/java/org/icij/datashare/text/ExtractedEntity.java
@@ -1,0 +1,27 @@
+package org.icij.datashare.text;
+
+import org.icij.datashare.Entity;
+import org.icij.datashare.model.ModelEntity;
+import org.icij.datashare.text.indexing.IndexId;
+import org.icij.datashare.text.indexing.IndexType;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/** An entity rebuilt from the statement store, in the shape the "&lt;project&gt;.entities" index holds:
+ *  the property keys are the namespaced wire form ("ftm:birthDate") the statements were stored under. */
+@IndexType("ExtractedEntity")
+public record ExtractedEntity(@IndexId String id, String model, Set<String> modelVersions, Set<String> types,
+                              Set<String> documentIds, Map<String, List<String>> properties) implements Entity {
+
+    public static ExtractedEntity from(ModelEntity entity) {
+        return new ExtractedEntity(entity.id(), entity.model(), entity.modelVersions(), entity.types(),
+                entity.documentIds(), entity.properties());
+    }
+
+    @Override
+    public String getId() {
+        return id;
+    }
+}
--- a/datashare-api/src/main/java/org/icij/datashare/text/ExtractedEntity.java
+++ b/datashare-api/src/main/java/org/icij/datashare/text/ExtractedEntity.java
@@ -1,5 +1,6 @@
 package org.icij.datashare.text;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import org.icij.datashare.Entity;
 import org.icij.datashare.model.ModelEntity;
 import org.icij.datashare.text.indexing.IndexId;
@@ -33,8 +34,10 @@ public record ExtractedEntity(@IndexId String entityId, String model, Set<String
     }
 
     /** The document id, which the entity id alone cannot be: the statement store emits one entity per
-     *  (entity id, model) pair, so two models describing the same entity would overwrite each other. */
+     *  (entity id, model) pair, so two models describing the same entity would overwrite each other.
+     *  Hidden from the source like {@link NamedEntity#getId()}: it already lives in the _id. */
     @Override
+    @JsonIgnore
     public String getId() {
         return model + NAMESPACE_SEPARATOR + entityId;
     }
--- a/datashare-api/src/main/java/org/icij/datashare/text/ExtractedEntity.java
+++ b/datashare-api/src/main/java/org/icij/datashare/text/ExtractedEntity.java
@@ -11,10 +11,14 @@ import java.util.Map;
 import java.util.Set;
 
 /** An entity rebuilt from the statement store, in the shape the "&lt;project&gt;.entities" index holds:
- *  the property keys are the namespaced wire form ("ftm:birthDate") the statements were stored under. */
+ *  the property keys are the namespaced wire form ("ftm_birthDate") the statements were stored under. */
 @IndexType("ExtractedEntity")
-public record ExtractedEntity(@IndexId String id, String model, Set<String> types, Set<String> modelVersions,
+public record ExtractedEntity(@IndexId String entityId, String model, Set<String> types, Set<String> modelVersions,
                               Set<String> documentIds, Map<String, List<String>> properties) implements Entity {
+    /** Not the colon {@link org.icij.datashare.model.Statement#qualifiedProperty()} stores: a colon is
+     *  the field/value delimiter of elasticsearch's query_string, so "properties.ftm:name:Jane" is a
+     *  parse error and every client would have to escape the separator to reach a single property. */
+    private static final String NAMESPACE_SEPARATOR = "_";
 
     // ModelEntity's keys are bare ("birthDate"): JooqStatementRepository.toRow strips the model
     // prefix a statement's property is stored under. The namespace goes back on here, at the index
@@ -22,13 +26,16 @@ public record ExtractedEntity(@IndexId String id, String model, Set<String> type
     // consume on the bare form.
     public static ExtractedEntity from(ModelEntity entity) {
         Map<String, List<String>> namespaced = new LinkedHashMap<>();
-        entity.properties().forEach((property, values) -> namespaced.put(entity.model() + ":" + property, values));
+        entity.properties().forEach((property, values) ->
+                namespaced.put(entity.model() + NAMESPACE_SEPARATOR + property, values));
         return new ExtractedEntity(entity.id(), entity.model(), entity.types(), entity.modelVersions(),
                 entity.documentIds(), namespaced);
     }
 
+    /** The document id, which the entity id alone cannot be: the statement store emits one entity per
+     *  (entity id, model) pair, so two models describing the same entity would overwrite each other. */
     @Override
     public String getId() {
-        return id;
+        return model + NAMESPACE_SEPARATOR + entityId;
     }
 }
--- a/datashare-api/src/main/java/org/icij/datashare/text/ExtractedEntity.java
+++ b/datashare-api/src/main/java/org/icij/datashare/text/ExtractedEntity.java
@@ -13,7 +13,7 @@ import java.util.Set;
 /** An entity rebuilt from the statement store, in the shape the "&lt;project&gt;.entities" index holds:
  *  the property keys are the namespaced wire form ("ftm:birthDate") the statements were stored under. */
 @IndexType("ExtractedEntity")
-public record ExtractedEntity(@IndexId String id, String model, Set<String> modelVersions, Set<String> types,
+public record ExtractedEntity(@IndexId String id, String model, Set<String> types, Set<String> modelVersions,
                               Set<String> documentIds, Map<String, List<String>> properties) implements Entity {
 
     // ModelEntity's keys are bare ("birthDate"): JooqStatementRepository.toRow strips the model
@@ -23,7 +23,7 @@ public record ExtractedEntity(@IndexId String id, String model, Set<String> mode
     public static ExtractedEntity from(ModelEntity entity) {
         Map<String, List<String>> namespaced = new LinkedHashMap<>();
         entity.properties().forEach((property, values) -> namespaced.put(entity.model() + ":" + property, values));
-        return new ExtractedEntity(entity.id(), entity.model(), entity.modelVersions(), entity.types(),
+        return new ExtractedEntity(entity.id(), entity.model(), entity.types(), entity.modelVersions(),
                 entity.documentIds(), namespaced);
     }
 
--- a/datashare-api/src/main/java/org/icij/datashare/text/ExtractedEntity.java
+++ b/datashare-api/src/main/java/org/icij/datashare/text/ExtractedEntity.java
@@ -5,6 +5,7 @@ import org.icij.datashare.model.ModelEntity;
 import org.icij.datashare.text.indexing.IndexId;
 import org.icij.datashare.text.indexing.IndexType;
 
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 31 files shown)


datashare-api/src/main/java/org/icij/datashare/text/ExtractedEntity.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/** An entity rebuilt from the statement store, in the shape the "&lt;project&gt;.entities" index holds:
+ *  the property keys are the namespaced wire form ("ftm_birthDate") the statements were stored under. */
+@IndexType("ExtractedEntity")
+public record ExtractedEntity(@IndexId String entityId, String model, Set<String> types, Set<String> modelVersions,
```
ClemDoum: What about `StructuredEntity` which is the high level name of the feature ?

datashare-app/src/main/java/org/icij/datashare/model/EntitiesIndexRebuilder.java
```diff
@@ -0,0 +1,63 @@
...
+        this.indexer = indexer;
+        this.statements = statements;
+    }
+
+    /** Drops the index and refills it, so an entity that left the store leaves the index too, and
+     *  returns how many entities were indexed. Dropping rather than emptying repairs an index created
+     *  before the entity mappings existed, which holds the document ones and rejects every entity, and
+     *  re-creating it is what gives a project older than the entities index one at all. */
+    public int rebuild(String projectId) throws IOException {
+        if (!Project.NAME_PATTERN.matcher(projectId).matches()) {
```
ClemDoum: Do we really need this check, I would expect project ID to be validated when creating the project, then the only check we need is to validate that the project actually exists.

ClemDoum: I feel like doing a check here could lead to errors if pattern evolved or validation for project name evolves.

ClemDoum (changes requested):

datashare-api/src/main/java/org/icij/datashare/text/ExtractedEntity.java
caro3801: question: Looks like `entityId` (the `@IndexId` field) is not `@JsonIgnore`d, and `getId()` returns a
different, namespaced value (`model + "_" + entityId`) that is also not `@JsonIgnore`d. 
Both `NamedEntity` and `Document` keep the `@IndexId` field and `getId()` in lockstep and hide both from JSON with `@JsonIgnore`, so every `ExtractedEntity` document ends up with two id-shaped keys in
`_source`: `"entityId": "person-1"` and `"id": "ftm_person-1"`. 

Is the duplicate `"id"` in the stored document intentional (e.g. useful for search/debugging), or should this follow the same `@JsonIgnore` convention as the other two entities?

datashare-app/src/main/java/org/icij/datashare/model/EntitiesIndexRebuilder.java
caro3801: I'd find more interesting to add a null-guard to avoid raising a `NullPointerException`

datashare-app/src/main/java/org/icij/datashare/web/ProjectResource.java
```diff
@@ -199,6 +199,7 @@ public Payload projectDelete(String id, Context context) throws IOException {
             Logger logger = LoggerFactory.getLogger(getClass());
             logger.info("Deleted {}'s record: {}", id, repository.deleteAll(id));
             logger.info("Deleted {}'s index: {}", id, indexer.deleteAll(id));
```
caro3801, line 201: hint: The comment just above `deleteEntitiesIndex` explains that a cascade step "must not abort the
delete and leave the queues, report map and artifacts behind," and that's exactly why the entities
index delete right below it is exception-guarded. 
The main index delete one line above it (`indexer.deleteAll(id)`) is pre-existing and out of this diff's scope, but it is not guarded the same way, so the same failure mode this PR just fixed for the entities index still exists for its sibling.
Worth a follow-up if you want the cascade to be resilient end to end, not blocking here.

datashare-app/src/main/java/org/icij/datashare/web/ProjectResource.java
```diff
@@ -307,11 +319,26 @@ ReportMap getReportMap(Project project) {
...
+        /** Saves the row and creates the indices, undoing the row if index creation fails: a row left
+         *  behind makes {@code projectExists} true, so every retry would answer 409 instead. */
+        boolean saveAndCreateIndex(Project project) {
+            if (!repository.save(project)) {
+                return false;
+            }
+            if (createIndexOnce(project.getId())) {
+                return true;
+            }
+            repository.deleteAll(project.getId());
```
caro3801: hint: Looks like the compensating `repository.deleteAll(project.getId())` return value is
dropped here. The docstring on `saveAndCreateIndex` promises that a failed index creation gets undone so a retry doesn't hit "project already exists" - but if this rollback itself fails (e.g. a transient
DB error), the row survives silently and the client is still told creation failed, with no signal
that a stale row is now blocking every retry. 

Could we log when the rollback itself doesn't succeed so we keep a trace of it?

```suggestion
            if (!repository.deleteAll(project.getId())) {
                LoggerFactory.getLogger(getClass()).error(
                        "could not roll back project row for {} after index creation failed", project.getId());
            }
            return false;
```

datashare-app/src/main/java/org/icij/datashare/web/IndexResource.java
```diff
@@ -81,6 +81,8 @@ public Payload getClusterInfo() throws IOException {
     public Payload createIndex(@Parameter(name = "index", description = "index to create", in = ParameterIn.PATH) final String index, Context context) throws IOException {
         modeVerifier.checkAllowedMode(Mode.LOCAL, Mode.EMBEDDED);
         try{
+            // a ".entities" name is granted through its base project, so it reaches here; createIndex
+            // picks the entity mappings for it rather than the document ones
             return indexer.createIndex(checkGrantedIndices(index, context)) ? created() : ok();
         } catch (IllegalArgumentException e){
```
caro3801: suggestion: `indexer.createIndex` can now throw a `RuntimeException` (`ConfigurationException`) from
the new entities-index creation step, but this catch block still only handles `IllegalArgumentException`. `ProjectResource.createIndexOnce` was widened to `RuntimeException | IOException`
for the same new failure mode.
Could we mirror that here so a transient ES error on the entities side returns a structured `BAD_REQUEST` instead of an unhandled 500?

```suggestion
        } catch (RuntimeException e){
```

datashare-index/src/main/java/org/icij/datashare/text/indexing/elasticsearch/ElasticsearchIndexer.java
```diff
@@ -534,7 +534,43 @@ public Searcher search(final List<String> indexesNames, Class<? extends Entity>
...
+                try {
+                    deleteIndex(indexName);
+                } catch (IOException rollback) {
+                    e.addSuppressed(rollback);
+                }
+            }
+            throw e;
+        }
+        return created;
+    }
```
caro3801, line 557: important: Before this change, `createIndex(indexName)` was a side-effect-free idempotent check: call
it on an index that already exists and nothing happens. 
Now it unconditionally also calls `createEntitiesIndex(indexName)`, even when `created` is `false` because the primary index was already there - so every "already exists, nothing to do" call now does an extra network round trip that can throw and abort the caller. 

`ElasticsearchSpewer.createIndexIfNotExists()` calls this uncaught from the `IndexTask` constructor, so a transient ES hiccup on the entities-index check alone would now abort an indexing run that needed zero work on the primary index. 
Could this be scoped to only run when `created` is true (leaving `EntitiesIndexRebuilder.rebuild()` as the path that retrofits an entities index onto an older project), something like:

```java
boolean created = ElasticsearchConfiguration.createIndex(client, indexName);
if (!created) {
    return false;
}
try {
    createEntitiesIndex(indexName);
} catch (RuntimeException e) {
    ...
```

caro3801 (changes requested):
Thanks! 

Very nice and easy-to-follow with the comments and the tests! The entities-index feature is well isolated and most of the new failure paths (rollback on partial create, per-step containment in the delete cascade) are clearly reasoned about in comments. 
The two things that will need to be resolved before merge are [the missed catch clause in `IndexResource`](https://github.com/ICIJ/datashare/pull/2361/changes#r3896607872) and [createIndex's new unconditional entities-index check](https://github.com/ICIJ/datashare/pull/2361/changes#r3896620405) turning a routine idempotent call into one that can fail on transient ES errors. 

There are 2 genuine open questions rather than bugs - happy to be told they're deliberate. Other comments are small, low-risk cleanups you can take or leave. 

I agree with Clément's comments too.

datashare-api/src/main/java/org/icij/datashare/text/ExtractedEntity.java
pirhoo: Done in ff0d40660

datashare-app/src/main/java/org/icij/datashare/model/EntitiesIndexRebuilder.java
pirhoo: Added the null check in 50668fa77

On removing the pattern check: it uses `Project.NAME_PATTERN`, the same constant project creation validates with, so the two cannot evolve separately. I would keep it because the id becomes part of the deletion request sent to Elasticsearch, and a value like `*` or `_all` there deletes every index in the cluster. Checking that the project exists instead would need a project repository the rebuilder does not have yet; happy to switch to that in #2206 once it gets wired if you prefer.

datashare-api/src/main/java/org/icij/datashare/text/ExtractedEntity.java
pirhoo: Good catch, the duplication was not intentional. `getId()` now carries `@JsonIgnore` like on `NamedEntity` and `Document`, so the namespaced id only exists as the Elasticsearch document id; `entityId` stays in the stored document because the entity is rebuilt from it when read back. Removed the `id` field from the mappings too. 50668fa77

datashare-app/src/main/java/org/icij/datashare/web/ProjectResource.java
pirhoo: Agreed, opened #2373 so the main index deletion gets the same protection.

datashare-app/src/main/java/org/icij/datashare/web/ProjectResource.java
pirhoo: Applied in 50668fa77 with a comment explaining why it matters: the leftover row makes every retry answer "project already exists".

datashare-app/src/main/java/org/icij/datashare/web/IndexResource.java
pirhoo: Done in 50668fa77 with one adjustment: `checkGrantedIndices` can throw `ForbiddenException` inside the same block, and it is also a `RuntimeException`, so catching everything would have turned that 403 into a 400. The catch lets `HttpException` keep its own status and returns a structured 400 for the rest.

datashare-index/src/main/java/org/icij/datashare/text/indexing/elasticsearch/ElasticsearchIndexer.java
pirhoo: Done in 50668fa77 : when the main index already exists, `createIndex` now returns false before doing anything about the entities index, so the call made from `IndexTask` behaves exactly as before this PR, and the rollback only applies to an index this call created. Added a test that pins this down. Giving an entities index to a project created before this feature stays `EntitiesIndexRebuilder`'s job, as you suggested.

caro3801 (approved):
Thanks! 🥳

datashare-app/src/main/java/org/icij/datashare/model/EntitiesIndexRebuilder.java
ClemDoum: Thanks for the precisions I'm fine with both solutions

ClemDoum (approved):

Issue #2203, opened 2026-06-16
feat: FtM-JSON exporter


pirhoo:
Part of #2194.

Export the extracted entities of a run as FollowTheMoney JSON.

- [ ] Add a small exporter interface so more output formats can be added later, one class each.
- [ ] Implement the FtM JSON exporter, reading a run from the statement store.
- [ ] Test that the output can be read back by the FtM library.

Depends on: #2202, #2200

## Added 2026-08-25, when #2202 was designed

- **This issue adds the `run_id` index on `statement`.** `047` ships only `statement_prj_entity` on `(prj_id, entity_id)`. Reading a run is this issue's access pattern and nothing else's, and an index is a cheap additive migration unlike a column, so it was deliberately deferred here rather than shipped unused.
- **Statements carry `model_version`** (the ontology version, e.g. `4.10.2`) alongside `model`, outside the id hash. Worth deciding whether the exported FtM JSON records it, since FtM's own entity JSON has no field for it.

pirhoo:
One item lands here out of #2204, recorded when #2204 was designed on 2026-08-26.

**The conditional `model_version` update was left undone.** #2202 shipped `save` as `onConflictDoNothing`, so a statement written under ontology 4.10.2 keeps `model_version = 4.10.2` forever even after the ontology moves; only new rows carry the new version, and the column becomes a mix rather than an answer to "which ontology validated this fact". The cheap remedy is a conditional update that writes only when the version actually differs, so a plain re-run stays churn-free:

```java
.onConflict(STATEMENT.PRJ_ID, STATEMENT.ID).doUpdate()
    .set(STATEMENT.MODEL_VERSION, version)
    .where(STATEMENT.MODEL_VERSION.ne(version))
```

It was not done in #2204 because nothing reads the column until this issue's exporter does, and it needs no migration, so it stays additive whenever you want it. Both dialects support `DO UPDATE ... WHERE` (SQLite since 3.24; confirm against the bundled driver before relying on it).

Also still open here, from the 2026-08-25 note above: the `run_id` index, and whether the exported FtM JSON records `model_version` at all, given FtM's own entity JSON has no field for it.

pirhoo:
## Inherited from #2204: the conditional `model_version` update

#2204 did not touch `model_version`, so the gap described in the [2026-08-26 note there](https://github.com/ICIJ/datashare/issues/2204) is still open and lands here, since this issue is the first thing that reads the column.

[...]

PR #2362, opened 2026-08-26
feat(tabular): turn a mapping and a row into statements

### LINKED ISSUE #2204: feat: mapping executor, mapping to statements

pirhoo:
Implements #2204. A `StatementBuilder` reads a persisted `ExtractionMapping` plus a tabular file's rows and produces `Statement`s with a deterministic entity id, so the same file mapped twice converges on the same entities.

Stacked on #2379. Review only this PR's commits; the three below carry the model and cleaning changes it builds on.

- feat(tabular): a row becomes the statements a mapping asks for, under a keyed or a keyless entity id
- feat(tabular): a property is filled from a column, a literal, several columns joined, a date read through its pattern, or the id of another entity the row builds
- feat(tabular): `validate()` refuses at save the mapping flaws no row could honour, and the executor checks again on run, since the ontology can move between the two
- feat(tabular): the executor is given the section the reader resolved instead of reading the mapping's options
- fix(tabular): an edge drops an endpoint that stored nothing, and a run counts a missing column once, not once per row

## Decisions

- Keyed entity id: model, type, optional key literal, sorted key *values*. No alias and no column name, so two files naming one identifier differently converge.
- Keyless entity id: alias, document, section, row number, so two keyless entities of one type stay two.
- Both are pinned by a literal-hash test. Changing either orphans stored statements.
- `new StatementBuilder(mapping, section)` replaces the one-argument form. Options hold what was asked for, an id needs what was read: `1` and `Sheet1` are one sheet and must hash alike, two tables of one document must not, and before this they produced identical ids.
- The section is a workbook's resolved sheet name, the Tika fallback's table index, or empty. #2206 passes whichever its reader resolved.
- No whole-entity validation here: a row contributes to an entity, it does not constitute one. That check moves to #2203.
- Left for #2206: no production caller, no web error translation for `InvalidExtractionMapping`, no `RepositoryFactory` method, no Guice binding.

Worth knowing: a JSON key is the one column name a mapping cannot always reach. Mapping columns get the cleaning headers get, `JsonRowSource` deliberately leaves record keys untouched, so a key holding an invisible character is named by no mapping and the read fails on its first row.


datashare-api/src/main/java/org/icij/datashare/tabular/DateFormats.java
```diff
@@ -0,0 +1,148 @@
+package org.icij.datashare.tabular;
```
ClemDoum, line 1: I think I don't understand the special treatment of dates here.
If that for literal only or for all dates coming from a tabular data and stored as a string ?

If it's for any tabular data with dates store as string will this tabular format also provide the date format they use ?

How will this work for CSVs or JSON, who will provide the date format ? Will we expect the user to provide it ?

datashare-api/src/main/java/org/icij/datashare/tabular/MappingExecutor.java
```diff
@@ -0,0 +1,264 @@
...
+
+    // Whatever the pattern cannot read is left exactly as it was rather than dropped or rewritten,
+    // so one 'n/a' in a date column does not cost a run, and is counted so a whole column that never
+    // converts cannot pass for a clean import. The log names the column, never the cell: an
+    // unconvertible cell is document content, and DEBUG logs are not bound by project access rules.
+    private String value(String cell, String format, Statement.Provenance provenance) {
+        if (format == null) {
+            return cell;
+        }
+        String iso = formats.iso(cell, format);
```
ClemDoum, line 234: There's something I don't understand here, we're calling date specific function `formats.iso` on a generic cell using a generic `format` property of the cell.

If `format` an arbitrary format expected for the string cell or its just here to parse dates stored as string values.

If format is just for dates, then we should probably rename everything related to this date parsing / filling more explicitly to dates: `PropertyMapping.dateFormat`, `readDateValue` (instead of `value`) etc...

Asked differently is there anything else than dates that we expect to validate / fill here ?

datashare-api/src/main/java/org/icij/datashare/tabular/MappingExecutor.java
```diff
@@ -0,0 +1,264 @@
...
+/**
+ * Turns a row into the statements a mapping asks for. One instance per run: rows are consumed on one
+ * thread, and nothing here reads or writes anything, so the caller owns the source and the store.
+ *
+ * <p>A row carries one entity's contribution, never the whole entity, so nothing here judges an
+ * entity against the model. A mapping that fills a required property on the first row of a group
+ * only would otherwise lose every other row of that group, and the entity a row contributes to is
+ * whole where its statements are regrouped, not here.
+ */
+public class MappingExecutor {
```
ClemDoum, line 31: This could be a `StatementParser`, I have the feeling it uses a mapping to extract statements from rows.

datashare-api/src/main/java/org/icij/datashare/tabular/MappingExecutor.java
```diff
@@ -0,0 +1,264 @@
...
+        mapping.entities().forEach(this::declare);
+        Stream.of(Skip.values()).forEach(reason -> skipped.put(reason, 0L));
+    }
+
+    /** What the run dropped, by reason, every reason present. */
+    public Map<Skip, Long> skipped() {
+        return Map.copyOf(skipped);
+    }
+
+    public List<Statement> statements(Row row) {
```
ClemDoum, line 68: Is the parser meant to be used on rows of the same document or on rows of several documents ? If it can be used on many documents isn't storing state in between ?

ClemDoum: I have the feeling the `sheet` is document specific (even narrower).

In this case we shouldn't really on the caller to make sure it calls this function on rows of the same sheet, we should enforce it in the API of this class. 

I think there are many ways to do it, one being to transform the API into `Stream<List<Statement>> rowStatements()` or `Stream<List<Statement>> parse()` so that people are not tempted to use this object on several sheets.

ClemDoum: This way we could also requirecolumns only once at the first column and not have to store the `checked` state

ClemDoum: The function could take some range as input if we want to parse a document partially.

datashare-api/src/main/java/org/icij/datashare/tabular/MappingExecutor.java
```diff
@@ -0,0 +1,264 @@
...
+        requireColumns(row);
+        Map<String, String> cells = new HashMap<>();
+        columns.forEach(column -> cells.put(column, cell(row, column)));
+        Map<String, String> ids = identify(cells, row.number());
+        Set<String> stored = stored(ids, cells);
```
ClemDoum, line 73: I struggled a bit to understand this part. 

If I understood right `identify` generates entity IDs for each expected entity ( what about `generateEntityIds()`). Cells will missing values empty values for any of the entity key colums are reported in `ENTITY_UNIDENTIFIED`

Then `stored` list columns of the cell/row which have a value (what about`getColumnsWithValue` or `columnsWithValues` or `getNonEmptyColumns`).

Then from we list all ID columns found on the row and check again in the cell's columns if we have a value for it and report `ENTITY_EMPTY` otherwise. What the difference with the first check in generateEntityIds ?

datashare-api/src/main/java/org/icij/datashare/tabular/MappingExecutor.java
```diff
@@ -0,0 +1,264 @@
...
+    /** What the run dropped, by reason, every reason present. */
+    public Map<Skip, Long> skipped() {
+        return Map.copyOf(skipped);
+    }
+
+    public List<Statement> statements(Row row) {
+        requireColumns(row);
+        Map<String, String> cells = new HashMap<>();
+        columns.forEach(column -> cells.put(column, cell(row, column)));
+        Map<String, String> ids = identify(cells, row.number());
```
ClemDoum, line 72: Question here, can an entity be optional, how is that reflected here ?
For instance someone we have `Person` and an optional `Employer` we can have missing employer ID and that's fine.

Do we take this into account here or report an error ?

datashare-api/src/main/java/org/icij/datashare/tabular/MappingExecutor.java
```diff
@@ -0,0 +1,264 @@
...
+            } else if (mapped.join() != null) {
+                filling.fill(property, mapped.columns().stream()
+                        .map(cells::get).filter(cell -> !cell.isEmpty())
+                        .collect(joining(mapped.join())), mapped.format(),
+                        provenance(row, String.join(",", mapped.columns())));
+            } else {
+                for (String column : mapped.columns()) {
+                    filling.fill(property, cells.get(column), mapped.format(), provenance(row, column));
+                }
+            }
```
ClemDoum, line 174: This part's branching is quite complex.

I think mentioned it in another PR but it would be simpler if `PropertyMapping` would have subtypes or would be an interface  so that we don't have to worry about the values of its different attributes and potential conflicts between them.

Each implem would have only the attributes it needs and wouldn't have to check in here what's `null` or not and we could just differ the filling logic to the implementations.

ClemDoum: The implems could turn a cell into statements

datashare-api/src/main/java/org/icij/datashare/tabular/MappingExecutor.java
```diff
@@ -0,0 +1,264 @@
...
+        String cell = row.values().getOrDefault(column, "");
+        if (cell.indexOf('\u0000') >= 0) {
+            count(Skip.CELL_UNREADABLE, column, row.number());
+            return "";
+        }
+        return Row.clean(cell);
+    }
+
+    private void count(Skip reason, String what, long rowNumber) {
+        skipped.merge(reason, 1L, Long::sum);
```
ClemDoum, line 261: I'm wondering about this "near silent" skips of errors.

I feel like for some error cases (probably not all of them) explicit failure at the first row would be  useful to stop processing early on. Say we have hundred of thousands of row and the user made a mistake in the mapping picking the wrong columns.

Or we're implementing the preview where it's important to provide complete feedback to the user so that it corrects a potentially wrong mapping.

Maybe the aggregation vs. explicit failure in case of error should be configurable (if that make sense)

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
```diff
@@ -32,20 +32,30 @@ public record ExtractionMapping(String id, String projectId, String userId, Stri
...
             Objects.requireNonNull(type, "type");
-            keys = List.copyOf(Objects.requireNonNull(keys, "keys"));
-            if (keys.isEmpty()) {
-                throw new InvalidEntityMapping(type);
-            }
+            // Keys name header columns, so they get the cleaning headers get: a key pasted with a
+            // non-breaking space still matches the header it names. No key at all is a valid
+            // mapping: each row is then its own record, identified by where it sits in the file.
+            keys = List.copyOf(Objects.requireNonNull(keys, "keys")).stream().map(Row::clean).toList();
             properties = Map.copyOf(Objects.requireNonNull(properties, "properties"));
```
caro3801, line 51: important `keyLiteral` is not cleaned against blank character like other keys. It can lead to quietly splitting one entity into two instead of merging them (similar to `PropertyMapping.literal()`).

```suggestion
            keys = List.copyOf(Objects.requireNonNull(keys, "keys")).stream().map(Row::clean).toList();
            keyLiteral = keyLiteral == null ? null : Row.clean(keyLiteral);
            properties = Map.copyOf(Objects.requireNonNull(properties, "properties"));
```

datashare-api/src/main/java/org/icij/datashare/tabular/MappingExecutor.java
```diff
@@ -0,0 +1,264 @@
...
+     *  options hold what was asked for and the id needs what was read: "1" and "Sheet1" name one
+     *  sheet and must hash alike, while two tables of one document must not. */
+    public MappingExecutor(ExtractionMapping mapping, String sheet) {
+        this.mapping = mapping;
+        this.documentId = mapping.documentId();
+        this.sheet = sheet;
+        mapping.requireValid();
+        mapping.entities().forEach(this::declare);
+        Stream.of(Skip.values()).forEach(reason -> skipped.put(reason, 0L));
+    }
```
caro3801, line 67: important: Could we add guard `sheet` against a NUL character here? Since this parameter is
deliberately a different string, the one the reader actually resolved, if that resolved value ever carries a NUL (a malformed workbook, a bad Tika-derived table label), nothing here catches it. 

```suggestion
    public MappingExecutor(ExtractionMapping mapping, String sheet) {
        this.mapping = mapping;
        this.documentId = mapping.documentId();
        if (sheet != null && sheet.indexOf('\u0000') >= 0) {
            throw new IllegalArgumentException("'sheet' contains a NUL character");
        }
        this.sheet = sheet;
        mapping.requireValid();
        mapping.entities().forEach(this::declare);
        Stream.of(Skip.values()).forEach(reason -> skipped.put(reason, 0L));
  }
```

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
```diff
@@ -60,25 +70,97 @@ public record PropertyMapping(List<String> columns, String join, String literal,
...
     public List<TargetModel.Violation> validate() {
         TargetModel target = TargetModelRegistry.get(model);
         List<TargetModel.Violation> violations = new ArrayList<>();
+        if (options.sheet() != null && holdsNul(options.sheet())) {
+            violations.add(new TargetModel.Violation("the sheet name holds a NUL character"));
+        }
+        if (holdsNul(documentId)) {
+            violations.add(new TargetModel.Violation("the document id holds a NUL character"));
+        }
+        DateFormats formats = new DateFormats();
```
caro3801, line 95: suggestion(perf): `validate()` builds its own `DateFormats` purely to prove every declared pattern compiles, then throws it away; `MappingExecutor` immediately builds a second one and recompiles the identical patterns. 
Since `ExtractionMapping` is a record meant to be reused across a run's many rows (and, per the stack's
description, potentially many documents once #2206 wires a real caller), this is a full `DateTimeFormatter`-compilation pass paid twice for the same mapping. 

Not urgent given the executor is still uncalled in production, but worth a note for when #2206 lands: could the mapping cache its compiled formats once, so `validate()` and the executor share them?

datashare-api/src/main/java/org/icij/datashare/tabular/DateFormats.java
```diff
@@ -0,0 +1,148 @@
...
+        } catch (IllegalArgumentException malformed) {
+            throw new UnusableDateFormat(pattern, "is not a date pattern: " + malformed.getMessage());
+        }
+        try {
+            date(format.parse(format.format(PROBE)));
+        } catch (DateTimeException undatable) {
+            throw new UnusableDateFormat(pattern, "cannot turn a cell into a date the model stores");
+        }
+        return format;
+    }
```
caro3801: hint: `date()` has a dedicated message for an offset pattern ("an offset is a precision the target model
does not store"), but `compile()`'s probe round-trip throws `UnsupportedTemporalTypeException` while
formatting the probe, before `date()` ever runs, so that specific message is unreachable for this input and callers just see the generic "cannot turn a cell into a date the model stores" instead. Small thing, but a mapping author debugging an offset pattern loses the more helpful explanation.

caro3801 (changes requested):
Nice piece of work and very well tested, the quality of the code is great! The two important comments are both about the same shape of bug: a string that skips the cleaning/validation its siblings get, so it can defeat the very invariant it exists for (key literal convergence, sheet-argument safety). Worth fixing before merge. The other two are follow-up-friendly and can wait, especially since
this executor has no production caller yet.

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
pirhoo: Fixed, with your suggestion: the key literal now gets the same cleaning as the keys.

It uncovered one more hole. Cleaning can turn a literal made only of invisible characters into an empty one, and an empty literal hashes exactly like no literal at all, so an author would think they had separated two aliases when they had not. A blank key literal is now refused at save, next to the blank property literal rule.

Two tests: an invisible character in a literal no longer splits an entity, and a blank literal fails validation.

datashare-api/src/main/java/org/icij/datashare/tabular/MappingExecutor.java
pirhoo: Fixed. The resolved sheet is the one string the mapping never saw, so validation cannot vouch for it, and a NUL in it would only surface deep inside the statement constructor, on the first row. It now fails at construction, with a test.

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
pirhoo: Agreed as a follow-up rather than here.

The double compile is per mapping, not per row: patterns are compiled once when they are declared and reused for every cell, so what is paid twice is a single pass over the mapping's own patterns at construction.

Caching them on the record would make a serialised value carry runtime state, and that cache has to be dropped whenever the ontology moves. I would rather make that call in #2206, once we know whether an executor is built per document or per run.

datashare-api/src/main/java/org/icij/datashare/tabular/DateFormats.java
pirhoo: Fixed by moving the check rather than the message. An offset is a property of the pattern, not of a parsed value, so it is now refused where the week-based fields are refused, and the message names the letter that caused it. The branch that could never run is gone.

The existing test now asserts the message instead of just the failure.

datashare-api/src/main/java/org/icij/datashare/tabular/DateFormats.java
pirhoo: Dates only, and only when the mapping asks for it.

The target model stores dates in ISO form, and a tabular file carries text, so someone has to say how that text is written. Nothing in a CSV or a JSON record declares it, so the mapping author does: the format sits on the property, next to the column it applies to, and it is optional. No format means the cell is stored exactly as it came.

Excel is the exception. Cells are typed there, so the reader already hands us an ISO date and no format is needed.

A cell the pattern cannot read is stored unchanged and counted, so one "n/a" in a date column does not cost a run, while a column that never converts shows up in the tally.

datashare-api/src/main/java/org/icij/datashare/tabular/MappingExecutor.java
pirhoo: Dates are the only thing formatted, today and in the plan: everything else is stored as text.

So yes, `dateFormat` it should be, with the filling method renamed to match. Nothing is stored in production yet, so the JSON name is free to change. Not in this push, it lands next.

datashare-api/src/main/java/org/icij/datashare/tabular/MappingExecutor.java
pirhoo: It does not parse anything. The readers do that, and this class is handed rows that are already parsed, plus a mapping, so "parser" would point at the wrong half of the pipeline.

If "executor" reads as vague, `RowMapper` or `StatementBuilder` are closer. No strong preference between those two.

datashare-api/src/main/java/org/icij/datashare/tabular/MappingExecutor.java
pirhoo: > Is the parser meant to be used on rows of the same document or on rows of several documents ? If it can be used on many documents isn't storing state in between ?

One instance per document section. The sheet is a constructor argument and never changes, and the document id comes from the mapping, so an instance cannot be pointed at a second document or a second sheet.

> I have the feeling the `sheet` is document specific (even narrower). In this case we shouldn't really on the caller to make sure it calls this function on rows of the same sheet, we should enforce it in the API of this class.

What stays open is the caller feeding it rows that came from somewhere else, and you are right that a stream signature closes that.

> This way we could also requirecolumns only once at the first column and not have to store the `checked` state

True, that flag exists only because rows arrive one at a time.

> The function could take some range as input if we want to parse a document partially.

The readers already hand back a stream of rows, so a caller can take a slice before they ever reach this class.

I would rather change the signature in #2206, when there is a real caller: it decides whether it wants a whole stream, a batch, progress or cancellation, and the shape should follow that rather than guess it now.

datashare-api/src/main/java/org/icij/datashare/tabular/MappingExecutor.java
pirhoo: Close, with one correction that is probably what made it hard to read: the second one is not about columns, it is about entities.

The first asks "can this row name the entity?", meaning every key column has a value so an id can be built. The second asks "does this row have anything to say about it?", meaning at least one mapped property actually holds something, directly or through an edge whose endpoint stores.

They fail differently, which is why there are two counters. A row with a passport and every other cell blank names an entity and stores nothing. A row full of values with a blank passport has plenty to store and nothing to address it by.

Renaming both along your lines: `entityIds` and `storing`.

datashare-api/src/main/java/org/icij/datashare/tabular/MappingExecutor.java
pirhoo: Supported, and not an error. A row where the employer key is blank produces no employer id, so no employer statement, and the person's employer property is simply left unfilled. The run continues.

It is counted under the unidentified tally, because nothing here can tell "legitimately absent" from "the author picked the wrong column". The tally is a report, not a failure.

datashare-api/src/main/java/org/icij/datashare/tabular/MappingExecutor.java
pirhoo: > This part's branching is quite complex. (...) it would be simpler if `PropertyMapping` would have subtypes or would be an interface so that we don't have to worry about the values of its different attributes and potential conflicts between them.

Agreed it would read better, and the conflicts are exactly what the compact constructor spends its lines on today.

The cost is that this record is stored as JSON in the mapping definition, so subtypes mean a type discriminator inside stored mappings. I would rather wait until the web API in #2206 fixes what the client actually posts, so the stored shape changes once.

> The implems could turn a cell into statements

That half I would keep out. Filling needs the model, the entity id, the date formats and the other entities of the row, and none of that belongs in a value we serialise. Subtype the data, leave the filling here.

datashare-api/src/main/java/org/icij/datashare/tabular/MappingExecutor.java
pirhoo: The case you describe already fails on the first row: if the mapping names a column the file does not have, the run stops there instead of importing every row as empty.

What is counted rather than thrown is per-row data loss: a blank key, a date cell that does not match its pattern. Those genuinely vary row by row, and one bad cell in a million-row file should not cost the run.

For the preview, the tally is the feedback channel: read the first rows, show what was dropped and why. If that turns out not to be enough, a strict mode sits easily on top.

datashare-api/src/main/java/org/icij/datashare/tabular/MappingExecutor.java
pirhoo: Done: the field is `dateFormat` now, and the method that reads a cell through it is `dateValue`. The violation message follows: "an unusable date format".

datashare-api/src/main/java/org/icij/datashare/tabular/MappingExecutor.java
pirhoo: Done: `entityIds` and `storing`.

datashare-api/src/main/java/org/icij/datashare/tabular/DateFormats.java
ClemDoum: Thanks for detailing !

datashare-api/src/main/java/org/icij/datashare/tabular/MappingExecutor.java
ClemDoum: Cool thank you for the update !

datashare-api/src/main/java/org/icij/datashare/tabular/MappingExecutor.java
ClemDoum: OK i'm fine with doing the update when implementing the caller !

datashare-api/src/main/java/org/icij/datashare/tabular/MappingExecutor.java
ClemDoum: I like both options maybe with a slight preference for `StatementBuilder`.

datashare-api/src/main/java/org/icij/datashare/tabular/MappingExecutor.java
ClemDoum: Thanks for the precision

datashare-api/src/main/java/org/icij/datashare/tabular/MappingExecutor.java
ClemDoum: Sounds good for me ! thank you

datashare-api/src/main/java/org/icij/datashare/tabular/MappingExecutor.java
ClemDoum: Yes sorry when I said wrong columns I meant wrong keys columns with potentially missing values.
If we think counting errors and making the diffs with the initial row count is enough for now  let start like this and you're right we can still add strict validation on the top of it

ClemDoum (approved):
Nice work thanks a lot !

datashare-api/src/main/java/org/icij/datashare/tabular/MappingExecutor.java
pirhoo: `StatementBuilder` it is, pushed: the class, its two tests and the PR body follow.

One caveat on the name, tell me if you read it differently: a `Builder` usually hands back the one thing it builds, and this hands back a list per row. `statements(row)` carries that on the method, so I think it reads fine, and `RowMapper` stays available if it bothers you later.

datashare-api/src/main/java/org/icij/datashare/tabular/StatementBuilder.java
```diff
@@ -0,0 +1,264 @@
...
+        String cell = row.values().getOrDefault(column, "");
+        if (cell.indexOf('\u0000') >= 0) {
+            count(Skip.CELL_UNREADABLE, column, row.number());
+            return "";
+        }
+        return Row.clean(cell);
+    }
+
+    private void count(Skip reason, String what, long rowNumber) {
+        skipped.merge(reason, 1L, Long::sum);
```
pirhoo: Agreed, let's start with the tally and keep a strict mode for when a caller asks for it.

One correction on the arithmetic, because it would bite the preview: the counters are per entity, not per row. A row counts one `ENTITY_UNIDENTIFIED` per entity it could not name, so a mapping declaring three entities can add three for a single bad row. Row count minus errors is therefore not the number of rows imported.

What it does give you is which reason dominates, and the debug log names the alias and the row number. A key column picked wrong shows up as an `ENTITY_UNIDENTIFIED` count near a multiple of the row count. If the preview needs a real "rows in, rows kept", I would rather count rows in #2206 than infer it from this.

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
```diff
@@ -32,20 +32,35 @@ public record ExtractionMapping(String id, String projectId, String userId, Stri
...
-    public record PropertyMapping(List<String> columns, String join, String literal, String entity, String format) {
+    public record PropertyMapping(List<String> columns, String join, String literal, String entity,
+                                  String dateFormat) {
         public PropertyMapping {
-            columns = List.copyOf(columns == null ? List.of() : columns);
+            // Column names get the cleaning headers get, and a literal gets the cleaning cells get:
+            // a mapping authored by copy-paste behaves like the file it was copied from. JSON is the
+            // exception: a record carries its own keys and JsonRowSource leaves them untouched, so a
+            // key holding an invisible character is named by no mapping and the read fails on row 1.
+            columns = List.copyOf(columns == null ? List.of() : columns).stream().map(Row::clean).toList();
```
caro3801, line 56: question: Columns are not deduped. Could that be an issue?
```suggestion
            columns = List.copyOf(columns == null ? List.of() : columns).stream().map(Row::clean).distinct().toList();
```

datashare-api/src/main/java/org/icij/datashare/tabular/StatementBuilder.java
```diff
@@ -0,0 +1,270 @@
...
+    public StatementBuilder(ExtractionMapping mapping, String sheet) {
+        this.mapping = mapping;
+        this.documentId = mapping.documentId();
+        // The one string here the mapping never saw, so validate() cannot vouch for it: a workbook
+        // naming a sheet with a NUL would otherwise reach Statement's constructor and abort the run
+        // on its first row.
+        if (sheet != null && sheet.indexOf('\u0000') >= 0) {
+            throw new IllegalArgumentException("the sheet name holds a NUL character");
+        }
+        this.sheet = sheet;
```
caro3801, line 61: important :  `sheet` only gets a NUL check instead of `Row.clean()` like every other strings in the statement. Cleaning it here would guarantee "same file mapped twice converges on the same entities" with the same hash id in case of non-breaking space or similar invisible character

```suggestion
        this.sheet = sheet == null ? null : Row.clean(sheet);
```

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
```diff
@@ -54,31 +69,107 @@ public record PropertyMapping(List<String> columns, String join, String literal,
...
+     *  any string a statement would carry (Statement's constructor aborts on one, so catching it
+     *  here is what keeps a bad mapping from killing a run mid-file). The compact constructor only
+     *  checks the model is known, so this is the only complete check, run on save and again by the
+     *  builder, so a mapping stored before the ontology moved underneath it still loads and fails
+     *  only when run. Aliases and property names are walked in sorted order, so the same mapping
+     *  always reports the same violations in the same order. */
     public List<TargetModel.Violation> validate() {
         TargetModel target = TargetModelRegistry.get(model);
         List<TargetModel.Violation> violations = new ArrayList<>();
+        if (options.sheet() != null && holdsNul(options.sheet())) {
```
caro3801, line 83: suggestion: there is a lot of near-identical blocks following two repeated shapes (`if (x != null && holdsNul(x)) violations.add(...)` and a blank-string check), and the wording has already started to drift between sites ("the sheet name holds a NUL character" vs "has a key literal holding a NUL character"). Could a couple of small helpers, to replace the repeated
blocks ?

datashare-api/src/main/java/org/icij/datashare/tabular/StatementBuilder.java
```diff
@@ -0,0 +1,270 @@
...
+            List<String> values = keys.stream().map(cells::get).toList();
+            if (values.stream().anyMatch(String::isEmpty)) {
+                count(Skip.ENTITY_UNIDENTIFIED, alias, rowNumber);
+            } else {
+                ids.put(alias, id(mapping.entities().get(alias).type(),
+                        mapping.entities().get(alias).keyLiteral(), values));
+            }
+        });
+        return ids;
+    }
```
caro3801, line 137: hint: `entityIds()` looks up `mapping.entities().get(alias)` twice in the keyed branch (once for
`.type()`, once for `.keyLiteral()`). Binding it to a local once would save a lookup per keyed entity
per row.

```suggestion
    private Map<String, String> entityIds(Map<String, String> cells, long rowNumber) {
        Map<String, String> ids = new TreeMap<>();
        keyColumns.forEach((alias, keys) -> {
            ExtractionMapping.EntityMapping entity = mapping.entities().get(alias);
            if (keys.isEmpty()) {
                ids.put(alias, rowId(alias, entity.type(), rowNumber));
                return;
            }
            List<String> values = keys.stream().map(cells::get).toList();
            if (values.stream().anyMatch(String::isEmpty)) {
                count(Skip.ENTITY_UNIDENTIFIED, alias, rowNumber);
            } else {
                ids.put(alias, id(entity.type(), entity.keyLiteral(), values));
            }
        });
        return ids;
    }
```

datashare-api/src/main/java/org/icij/datashare/tabular/StatementBuilder.java
```diff
@@ -0,0 +1,270 @@
...
+            }
+        }
+        return stored;
+    }
+
+    private boolean fills(String alias, Map<String, String> cells, Set<String> stored) {
+        return mapping.entities().get(alias).properties().values().stream().anyMatch(mapped ->
+                mapped.literal() != null || stored.contains(mapped.entity())
+                        || mapped.columns().stream().anyMatch(column -> !cells.get(column).isEmpty()));
+    }
```
caro3801, line 206: hint: `fills()` re-fetches `mapping.entities().get(alias)` and rebuilds a fresh properties stream on every pass of the `storing()` fixed-point loop, even though that per-alias list never changes across passes or rows. On a mapping with several interdependent entity-reference properties this could add up. Worth precomputing the per-alias property/column list once in `declare()` if this loop ever shows up in profiling, not urgent otherwise.

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
```diff
@@ -32,20 +32,35 @@ public record ExtractionMapping(String id, String projectId, String userId, Stri
...
+            // entity in two, which is the merge the literal exists to control.
+            keys = List.copyOf(Objects.requireNonNull(keys, "keys")).stream().map(Row::clean).toList();
+            keyLiteral = keyLiteral == null ? null : Row.clean(keyLiteral);
             properties = Map.copyOf(Objects.requireNonNull(properties, "properties"));
         }
     }
 
-    public record PropertyMapping(List<String> columns, String join, String literal, String entity, String format) {
+    public record PropertyMapping(List<String> columns, String join, String literal, String entity,
+                                  String dateFormat) {
```
caro3801: question: The date field was renamed from `format` to `dateFormat` with no `@JsonAlias` bridge, and
the mapper deserializes with `FAIL_ON_UNKNOWN_PROPERTIES=false`, so a mapping JSON still carrying the old key would silently lose its date format instead of failing loudly. Since this PR has no production
caller yet (wiring lands in #2206), is this worth a bridge now, or is it safe to defer since no
mapping has ever been persisted under the old name?

caro3801 (changes requested):
Two new important findings (question about duplicate join columns, sheet not going through Row.clean) that are both small, localized fixes worth doing before merge since they touch the determinism guarantee the PR is built around. Everything else here is cleanup or a question, nothing blocking. Thanks for the huge work and the fixes

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
pirhoo: Fixed, deduped. It bites in one place only: a join glues a cell to itself, so one column named twice writes its value twice. Without a join the repeat is invisible, the two statements carry the same id and collapse.

Cleaning now dedupes both lists, columns and keys, so a join listing one column twice is refused by the rule that already asks for more than one column, and the key dedup the builder did per mapping is gone. Two tests.

datashare-api/src/main/java/org/icij/datashare/tabular/StatementBuilder.java
pirhoo: Fixed, with your suggestion. The sheet is cleaned after the NUL check, so it gets what every other string in a statement gets, and a sheet named with a non-breaking space in one read no longer splits a keyless entity in two.

Test: a sheet written with a non-breaking space and one written plain land on the same id.

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
pirhoo: Fixed for the NUL shape: one helper, and all six messages now read the same way, "<what> holds a NUL character". The drift goes with them.

Left the blank checks inline: each carries its own explanation (no header can match, hashes like no literal at all, no row can store), so a helper would hide the if and keep the strings anyway.

datashare-api/src/main/java/org/icij/datashare/tabular/StatementBuilder.java
pirhoo: Applied, your suggestion as is.

datashare-api/src/main/java/org/icij/datashare/tabular/StatementBuilder.java
pirhoo: Agreed, not here. The per-alias list is worth precomputing, but where it lives depends on whether #2206 builds one of these per document or per run, so I would rather do it there, with a caller to measure.

datashare-api/src/main/java/org/icij/datashare/tabular/ExtractionMapping.java
pirhoo: Safe to defer, and nothing can have persisted the old key: the repository is only ever constructed by the db tests, there is no factory method and no binding, so no running Datashare has ever written a mapping. An alias now would bridge a format no stored row holds.

#2206 fixes the stored shape when it wires the web API.

caro3801 (approved):
Thanks for all the changes!

PR #2377, opened 2026-09-03
feat(model): keep the value as it was read alongside the formatted one

pirhoo:
A statement keeps the value as it was read when a format changed it, so `1970-03-01` can still show the `01/03/1970` the document carried. It sits outside the id hash, so the same fact keeps the same id either way.

Bottom of the stack that ends with #2204. Changeset 048 needs `mvn -pl datashare-db initialize liquibase:update`.

- feat(model): a statement carries the value as it was read alongside the formatted one
- feat(db): the raw value is stored and refreshed on re-save (changeset 048)

Write-only for now: nothing reads it back, so it stays out of `READ_FIELDS`, the one query that scans the whole statement table. Add it there when a reader needs it.


datashare-api/src/main/java/org/icij/datashare/model/Statement.java
```diff
@@ -5,12 +5,16 @@
...
 
 public record Statement(String id, String model, String entityId, String entityType,
-                        String property, String value, Provenance provenance) {
+                        String property, String value, String originalValue, Provenance provenance) {
     /** An entity id ends up in the (prj_id, entity_id) index, whose entries Postgres caps at 2704
      *  bytes and SQLite does not cap at all. Bounding it here is what keeps the two dialects from
      *  disagreeing on which mapping key a project accepts. */
     public static final int MAX_ENTITY_ID_LENGTH = 512;
 
+    /** The digest behind every tabular id, statement and entity alike, pinned by test: changing it
```
caro3801: question: the comment say it's "pinned by test", but I can't find it. Can you check ?

caro3801 (approved):
Clean and light PR thanks !

datashare-api/src/main/java/org/icij/datashare/model/Statement.java
pirhoo: Right, the comment was overstated. The only pin was a `length == 96` assertion inside `test_the_same_fact_gets_the_same_id`, which pins the digest width and not the algorithm. "entity alike" was also wrong: `DIGESTER` only builds statement ids, entity ids come in from the caller. Added `test_the_id_digest_is_pinned` with the expected id and reworded the comment to point at it.

pirhoo: Good catch, the comment overstates it. The only pin is `StatementTest.test_the_same_fact_gets_the_same_id`, which asserts the id is 96 chars, so it pins the digest width and not the algorithm. "entity alike" is also wrong: `DIGESTER` is only used for statement ids, entity ids come in from the caller. I'll add a golden-vector assertion for one known statement id and reword the comment to point at that test.

PR #2378, opened 2026-09-03
refactor(model): an entity has one type, not a set of them

pirhoo:
An entity carries one type instead of a set. #2204 hashes the type into the entity id, so two types under one id can only be corrupt or legacy data, and a multi-type entity becomes impossible by construction.

Stacked on #2377.

- refactor(model): an entity has one type, and folding a group that carries two is refused
- refactor(model): `TargetModel.validate` drops its cross-type reconciliation, FtM `serialize` refuses a type that is no schema of the model
- fix(index): the entities index names the model's type `entityType`, since the indexer already owns `type`
- fix(model): a rebuild skips only `UnrebuildableEntity`; the other refusals mean the statements were grouped wrongly, and now stop it rather than produce a short index
- fix(db): a fold that fails part way gives up its own group, not the rest of the stream

That last one was silent on the lazy Postgres path: an aborted fold left the iterator reading as exhausted, so every entity after it was dropped and the rebuild still reported success.


datashare-api/src/main/java/org/icij/datashare/model/ModelEntity.java
```diff
@@ -35,13 +36,13 @@ public record ModelEntity(String model, String id, Set<String> types, Set<String
...
         SortedSet<String> models = new TreeSet<>();
-        Set<String> types = new TreeSet<>();
+        String type = null;
         Set<String> documentIds = new TreeSet<>();
         Map<String, SortedSet<String>> values = new TreeMap<>();
         for (Statement statement : statements) {
             ids.add(statement.entityId());
             models.add(statement.model());
-            types.add(statement.entityType());
+            type = statement.entityType();
```
caro3801: question: it will only carry the last type found in the list (overriden each time). Is it ok ?

datashare-api/src/main/java/org/icij/datashare/text/StructuredEntity.java
```diff
@@ -12,9 +12,11 @@
 import java.util.Set;
 
 /** An entity rebuilt from the statement store, in the shape the "&lt;project&gt;.entities" index holds:
- *  the property keys are the namespaced wire form ("ftm_birthDate") the statements were stored under. */
+ *  the property keys are the namespaced wire form ("ftm_birthDate") the statements were stored under,
+ *  and the model's type is {@code entityType}, since the indexer stamps every document's {@code type}
+ *  field with its kind ("StructuredEntity"). */
 @IndexType("StructuredEntity")
-public record StructuredEntity(@IndexId String entityId, String model, Set<String> types, Set<String> modelVersions,
+public record StructuredEntity(@IndexId String entityId, String model, String entityType, Set<String> modelVersions,
```
caro3801: question: `@IndexId` is left on `entityId`, but `getId()` below overrides it to return
`model + "_" + entityId`. But `JsonObjectMapper`'s reflection-based id lookup reads the `@IndexId` field directly so it might create discrpencies. Should the annotation move (or a comment explain why it's deliberately left on the wrong field)?

caro3801 (approved):
Small, well-contained PR: collapsing the type set to a single field simplifies `TargetModel.validate` and the FtM serializer nicely. The comments are more about decisions than changes. Thanks for the comments throughout, they made this an easy review to follow!

datashare-api/src/main/java/org/icij/datashare/model/ModelEntity.java
pirhoo: Nice catch, and no, it is not ok. I dropped that check in the last commit of the branch, thinking the id recipe made it unreachable, and that reasoning was wrong: a group is cut on the entity id and the model, never on the type, so two types under one id land in the same group and the fold quietly keeps whichever row came last. The type would then depend on row order.

The check and its two tests are back in d13564d38, in the same shape as the neighbouring id and model refusals, and the repository test proves the case is reachable through the real store rather than only in theory. Thanks for pushing on it.

datashare-api/src/main/java/org/icij/datashare/text/StructuredEntity.java
pirhoo: Good instinct, and worth checking. Nothing is affected today: the indexer takes the document id from `Entity.getId()`, and the reflective reader you have in mind, `JsonObjectMapper.getId`, has no callers left in the repo, not even a test.

Still, you are right that it reads wrong, and `NamedEntity` and `SourcePath` both annotate exactly the field their `getId` returns. So rather than add a comment defending a field nothing reads, I dropped the annotation in 0b6954b223 and left `getId` as the single answer. Happy to go the other way and delete the unused reader instead if you would rather keep the marker as a convention.

ClemDoum (approved):

PR #2379, opened 2026-09-03
feat(tabular): one cleaning rule for headers and cells alike

pirhoo:
Headers and cells go through one cleaning rule: U+00A0 reads as a space, an invisible formatting character or stray BOM is removed, surrounding whitespace is not content. Without it an invisible character silently splits one key value into two entity ids in #2204.

Stacked on #2378.

- feat(tabular): one cleaning rule for headers and cells alike
- fix(tabular): a surplus cell holding only invisible characters no longer aborts the read
- refactor(tabular): the csv reader loses its own BOM strip, a narrower duplicate
- fix(tabular): the cleaning rule reaches the mapped cell, not only the header name
- fix(tabular): every reader calls a row of invisible filler blank, as `Row` does

### The characters

These have no visible shape of their own, so `Name` and `Na​me` look identical in a spreadsheet yet hash to different ids.

| | What it is | What the rule does |
|---|---|---|
| [U+00A0](https://www.compart.com/en/unicode/U+00A0), [U+2007](https://www.compart.com/en/unicode/U+2007), [U+202F](https://www.compart.com/en/unicode/U+202F) | [no-break spaces](https://en.wikipedia.org/wiki/Non-breaking_space): a space that keeps two words on the same line. Excel and Word write them without asking. | read as a plain space |
| [U+FEFF](https://en.wikipedia.org/wiki/Byte_order_mark) | the [BOM](https://en.wikipedia.org/wiki/Byte_order_mark), a marker some editors put at the very start of a file to flag its encoding | removed |
| [U+200B](https://en.wikipedia.org/wiki/Zero-width_space), [U+2060](https://www.compart.com/en/unicode/U+2060), [U+00AD](https://en.wikipedia.org/wiki/Soft_hyphen) | zero-width space, word joiner, soft hyphen: line-breaking hints with no shape | removed |
| [U+200E, U+200F](https://en.wikipedia.org/wiki/Implicit_directional_marks) | direction marks, used to lay out Arabic or Hebrew next to Latin text | removed |
| [U+200C, U+200D](https://en.wikipedia.org/wiki/Zero-width_joiner) | zero-width non-joiner and joiner | **kept** |

The rule names each character it removes instead of sweeping the zero-width range. U+200C and U+200D sit in that range and are not invisible at all: they spell words in Persian, Arabic and Hindi and hold a composed emoji together. Sweeping it rewrote values and, since the cleaned form is what gets hashed, merged two entities under one id.

Worth knowing: the rule can refuse a file it used to import. `Name` and `Name ` fold to one name, and a duplicate header fails the read. A cell reads the same way, so its surrounding whitespace no longer travels with the value, and a row holding nothing but this filler is skipped like an empty one.


datashare-api/src/main/java/org/icij/datashare/tabular/Row.java
caro3801, line 95: crucial: Looks like `clean()` never actually reaches the values a caller gets back. `headers()` cleans column names, and the 3-arg `values(headers, cells, number)` overload only calls `clean(cell)` to decide whether a *surplus* column is blank, but the map-building loop still stores the raw cell verbatim. That means a cell holding a BOM, zero-width character or non-breaking space keeps it in the returned value,
which is exactly the "one key value silently split into two entity ids" failure the class javadoc and this PR's title ("one cleaning rule for headers and cells alike") say is being fixed.

Could we clean the stored cell too?

```suggestion
    public static Map<String, String> values(List<String> headers, List<String> cells) {
        Map<String, String> values = new LinkedHashMap<>();
        for (int column = 0; column < headers.size(); column++) {
            if (headers.get(column) != null) {
                values.put(headers.get(column), column < cells.size() ? clean(cells.get(column)) : "");
            }
        }
        return Collections.unmodifiableMap(values);
    }
```

Worth a `RowTest` case asserting a mapped value (not just a header) gets cleaned, since nothing today covers this path.

datashare-api/src/test/java/org/icij/datashare/tabular/RowTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+    @Test
+    public void test_clean_removes_a_soft_hyphen() {
+        assertThat(Row.clean("AB\u00AD123")).isEqualTo("AB123");
+    }
+
+    @Test
+    public void test_a_surplus_cell_holding_only_a_non_breaking_space_carries_no_data() {
+        assertThat(Row.values(List.of("a"), List.of("1", "\u00A0"), 2L)).isEqualTo(Map.of("a", "1"));
+    }
```
caro3801, line 57: suggestion: Looks like every test here covers headers or the surplus-cell blank check, but nothing asserts that a normal mapped cell value comes back cleaned (previous comment is not caught here)

Could we add a case like this ?

```suggestion
    @Test
    public void test_a_surplus_cell_holding_only_a_non_breaking_space_carries_no_data() {
        assertThat(Row.values(List.of("a"), List.of("1", " "), 2L)).isEqualTo(Map.of("a", "1"));
    }

    @Test
    public void test_a_mapped_cell_carrying_a_non_breaking_space_reads_like_its_plain_form() {
        assertThat(Row.values(List.of("a"), List.of("1 "))).isEqualTo(Map.of("a", "1"));
    }
```

datashare-app/src/main/java/org/icij/datashare/tabular/DelimitedRowSource.java
caro3801: important: This blank-row check still uses `String::isEmpty`, so a row whose cells hold only a non-breaking space or other invisible filler is no longer treated as blank the way a truly empty row is, once `clean()` is the row's own definition of "no content" (see `Row.values(headers, cells, number)` which already uses `!clean(cell).isEmpty()` for the same purpose).

Could this use the same `clean()`-aware emptiness check, e.g. `cells.stream().allMatch(cell ->Row.clean(cell).isEmpty())`?


This is also present in 
* `datashare-app/src/main/java/org/icij/datashare/tabular/WorkbookRowSource.java:101` `values.stream().allMatch(String::isEmpty)` won't catch a row whose cells only carry invisible/NBSP filler, unlike `Row`'s own blank check.
* `datashare-app/src/main/java/org/icij/datashare/tabular/TikaTableRowSource.java:95`
 `values.stream().allMatch(String::isEmpty)` should probably line up with `Row.clean()`'s notion of blank so all three row sources treat an invisible-character-only row the same way `Row` does.

caro3801 (changes requested):
Nice, focused change overall, the `clean()` javadoc and comments do a lot of work explaining *why* each character is or isn't stripped, which makes the intent easy to follow. The one thing worth fixing
before merge is : right now the cleaning rule only reaches header names, not the cell values it's meant to protect, so the PR's stated goal isn't quite met yet for the data itself. This might require additionnal test cases that would have caught it. Once those are in, other comments are a quick follow-up to keep the three row
sources' blank-row detection consistent with the new rule.

datashare-api/src/main/java/org/icij/datashare/tabular/Row.java
pirhoo: Fixed. `values(headers, cells)` now stores `clean(cell)`, so a mapped value follows the same rule as a header name. Two `RowTest` cases cover it, one on a non-breaking space and one on a zero-width space.

datashare-api/src/test/java/org/icij/datashare/tabular/RowTest.java
pirhoo: Added, written against `Row.values(...)` directly: `test_a_mapped_cell_carrying_a_non_breaking_space_reads_like_its_plain_form` and `test_a_mapped_cell_carrying_a_zero_width_character_reads_like_its_plain_form`. The surplus-cell case you quoted was already there, so I left it as is.

datashare-app/src/main/java/org/icij/datashare/tabular/DelimitedRowSource.java
pirhoo: Fixed in all three. Rather than repeat the predicate, `Row.blank(String)` now holds the definition and the readers call `allMatch(Row::blank)`; the surplus-cell check in `Row` uses it too, so there is one notion of blank. A test per source covers a row of invisible filler: a csv record, a workbook row and an html row.

caro3801 (approved):

PR #2380, opened 2026-09-03
perf(db): leave an unchanged statement row untouched on re-save

pirhoo:
A no-op re-run rewrote every statement row and both its index entries, for values nothing reads: 35 to 70 to 94 MB over three identical saves. The upsert is now conditional, so only a row whose content actually moved is rewritten.

Stacked on #2362. Changeset 050 needs `mvn -pl datashare-db initialize liquibase:update`.

- perf(db): the statement upsert is conditional, so a no-op re-run rewrites nothing
- refactor(db): one `written_at` replaces `first_seen` and `last_seen`, refreshed only when the row is actually rewritten (changeset 050)

## Decisions

- `written_at` replaces `first_seen` and `last_seen`. Once an unchanged row is left alone, "last seen" stops being true and equals "first seen"; nothing read either column.
- `run_id` stays, but it is not a reference: nothing states a run id is a task id, and `JooqTaskRepository` deletes tasks with no foreign key. A date cannot be recovered by joining `task`, so it stays on the row.
- SQLite gets its `DROP COLUMN` written out. Liquibase emits none for that dialect and leaves the column in place silently, which then fails the insert on a `NOT NULL` column that should be gone.
- Numbered 050 while #2381 above carries 049. Liquibase runs changelog order, so the prefix is cosmetic and renumbering would rewrite a commit in the PR above for nothing.


datashare-db/src/main/java/org/icij/datashare/db/JooqStatementRepository.java
```diff
@@ -90,12 +90,22 @@ public int save(String projectId, String runId, Stream<Statement> statements) {
     // cannot drift, and the conflict branch reads the row being inserted through EXCLUDED rather than
     // binding its own values, which would not line up with the record's positional binds.
     private static int saveChunk(DSLContext create, Write write, List<Statement> chunk) {
+        // The update is conditional so a no-op re-run rewrites nothing: unconditional, every row and
+        // both index entries are rewritten on every re-run (measured 35->70->94 MB over three
```
ClemDoum: I'm not sure to understand this reference. Is it the the growth of the DB at each call ?

ClemDoum: If this is the result of some XPs, it could probably remain on the PR side rather than in code's comments (all the more that it refers to a state of the code that we can't see in the code, we do see the git diff on the PR).

pirhoo: Yes: it's the table plus its two indexes on disk, measured after three identical saves of the same statements. And you're right, it belongs in the PR body where it already is. I'm cutting the numbers from the comment and keeping only the reason the condition is there.

ClemDoum: Great thanks !

caro3801 (approved):

PR #2381, opened 2026-09-03
feat(db): retract a sheet's statements before rewriting them

pirhoo:
Re-extracting a sheet has to retract what it wrote last time, or a corrected value, a re-typed mapping or a shrunken file leaves stale statements behind.

Stacked on #2380. Nothing here has a caller yet; #2206 wires it. Changesets 049 and, from the PR below, 050: `mvn -pl datashare-db initialize liquibase:update`. Changeset 049 was edited in place, so a database that already ran it needs its `databasechangelog` row and the old `statement_prj_doc` index dropped first.

- feat(db): retraction deletes the statements one sheet of a document contributed, indexed on `(prj_id, doc_id)` (changeset 049)
- feat(model): `replace()` retracts a sheet and rewrites it, reporting both counts
- fix(db): a null sheet retracts the rows the write stored under the empty one, the way `Provenance` already reads it
- fix(db): the retraction rides the first chunk's transaction, so an extraction that fails on its first row leaves the sheet whole
- fix(model): `replace()` refuses a statement whose provenance is not the document and sheet being replaced
- refactor(model): `deleteByDocument` is scoped to a sheet, so it is now `deleteBySheet`
- test(db): null sheet, sheet scoping through `replace()`, empty stream, first-row failure, provenance refusal

## Decisions

- Sheet scope, not document scope. A mapping reads one sheet, and a multi-sheet workbook is several mappings on one document, so document scope deleted sheets 1 and 3 when sheet 2 was replaced and never rewrote them.
- Run scope would fail the other way: one entity legitimately aggregates statements from several documents.
- `replace()` is declared on the interface but has no default body: the retraction has to share the transaction that writes the first chunk, and only the implementation can open it. Later chunks commit on their own, so a failure part way through still leaves the sheet half rewritten.
- A stream that yields nothing retracts the sheet and writes nothing, which the two counts tell apart.
- `sheet` stays out of the index. `(prj_id, doc_id)` already narrows to the rows the delete has to visit, and an unbounded text column in a btree entry is what makes Postgres and SQLite disagree on what they accept.

Four things #2206 inherits:

- Changing which sheet a mapping reads strands what it wrote under the old one, since nothing then retracts it.
- Two mappings on one document and sheet retract each other, since `statement` carries no mapping id.
- `RowSource` never hands back the sheet it resolved, so a caller cannot yet name the key the write stored. The provenance check turns that mismatch into a refusal rather than a silent wipe.
- Every row `replace()` writes is new, so #2380's conditional upsert cannot spare any of them. A caller wanting the cheap no-op re-run calls `save()`.


datashare-api/src/main/java/org/icij/datashare/model/StatementRepository.java
```diff
@@ -8,19 +8,39 @@
...
+     *  nothing then retracts it. */
+    int deleteBySheet(String projectId, String documentId, String sheet);
+
+    /** Rewrites what one sheet of a document contributed. The retraction rides the first chunk's
+     *  transaction, so an extraction that fails on its first row leaves the sheet as it was rather
+     *  than empty; later chunks commit on their own, so a failure part way through does leave the
+     *  sheet half rewritten. Refuses a statement whose provenance is not the document and sheet being
+     *  replaced, since the delete keys on the arguments and the rows key on themselves. The rows
+     *  written here are new, so the conditional upsert a plain save leans on cannot spare any. */
+    Replaced replace(String projectId, String runId, String documentId, String sheet,
```
caro3801: choice: The PR description says `replace()` lives on the implementation, not on the interface, but this  diff actually adds `replace()` (and `deleteBySheet()`) as abstract methods on `StatementRepository` itself.
Every implementer, including the fake in `EntitiesIndexRebuilderTest`, now has to implement them, which is a different contract than what's documented. 
So it's either possible to move these two methods down to `JooqStatementRepository` to match the stated design, or update the PR description to reflect that they are now part of the interface.

pirhoo: Good catch, the description is stale. `replace()` started as a default method with its body on the interface, and the last commit moved the body down to the jOOQ implementation so the retraction can run inside the transaction that writes the first chunk. Only the body moved, the declaration stays on the interface, otherwise the caller in #2206 would have to depend on the jOOQ class directly.

I have reworded that bullet.

caro3801 (approved):

PR #2413, opened 2026-09-18
feat(cli): add --nextStage to target the ENQUEUEIDX output stage

pirhoo:
The ENQUEUEIDX task always sent documents to the next stage in the chain, defaulting to NLP, so targeting another stage meant chaining it with `--stages ENQUEUEIDX,ARTIFACT`. A new option names the target directly:

```
datashare stage run --stages ENQUEUEIDX --nextStage ARTIFACT
```

Without the option, nothing changes: the stages chain still decides.

- feat: add `--nextStage` on both CLI surfaces (JOpt and picocli), with no default so a settings file is never overwritten
- feat: reject an unknown stage name, and any stage running before ENQUEUEIDX since those consume file paths rather than document ids
- test: cover the override, both rejections, and the option propagation from the command line
- docs: add a `--nextStage` example to the `stage run` help


datashare-app/src/main/java/org/icij/datashare/tasks/EnqueueFromIndexTask.java
```diff
@@ -93,4 +94,24 @@ public Long call() throws Exception {
...
+     *  queue would poison it rather than fail. */
+    private static Stage parseNextStage(String value) {
+        Stage stage = Stage.parse(value).orElseThrow(
+                () -> new IllegalArgumentException("unknown --nextStage value \"%s\"".formatted(value)));
+        if (Stage.comparator.compare(stage, Stage.ENQUEUEIDX) <= 0) {
+            throw new IllegalArgumentException(
+                    "--nextStage %s runs before %s in the pipeline".formatted(stage, Stage.ENQUEUEIDX));
+        }
+        return stage;
+    }
```
caro3801, line 129: suggestion: `parseNextStage` only rejects stages by ordinal position, but two later stages
(`CREATENLPBATCHESFROMIDX` and `BATCHNLP`) never actually drain a stage-named queue. Neither extends `PipelineTask`, so `--nextStage CREATENLPBATCHESFROMIDX` (or `BATCHNLP`, which also has no entry in `CliApp.TASK_CLASSES`) would enqueue documents that sit there forever, and the task still logs success so the failure is silent. This is CLI-only and likely used by people who know the pipeline, so it's not blocking, but the fix is a one-line whitelist and it protects against a confusing debugging session.

Validating against the set of stages that actually have a `PipelineTask` consumer (effectively
`CATEGORIZE`, `NLP`, `ARTIFACT`) instead of an ordinal comparison could fix this + An allow list would make this future proof if non-queue-driven stage is inserted between `ENQUEUEIDX` and `ARTIFACT`.

```suggestion
    private static final java.util.Set<Stage> QUEUE_CONSUMING_STAGES =
            java.util.EnumSet.of(Stage.CATEGORIZE, Stage.NLP, Stage.ARTIFACT);

    private static Stage parseNextStage(String value) {
        Stage stage = Stage.parse(value).orElseThrow(
                () -> new IllegalArgumentException("unknown --nextStage value \"%s\"".formatted(value)));
        if (!QUEUE_CONSUMING_STAGES.contains(stage)) {
            throw new IllegalArgumentException(
                    "--nextStage %s has no queue consumer downstream of %s".formatted(stage, Stage.ENQUEUEIDX));
        }
        return stage;
    }

File: datashare-app/src/main/java/org/icij/datashare/tasks/EnqueueFromIndexTask.java
Comment (caro3801) [inline datashare-app/src/main/java/org/icij/datashare/tasks/EnqueueFromIndexTask.java:59]: important: Since `nextStage()` is now resolved lazily in `call()` (line 67), the `outputQueue` field
built by `PipelineTask`'s constructor (via `super(...)` at line 53) still resolves the stages-chain
default and constructs an unused `DocumentQueue` for that default stage every time this task runs. 

 For example, running  `--nextStage ARTIFACT` will resolves the stages-chain *default* next stage for `ENQUEUEIDX`, which is `NLP`, and opens a real `DocumentQueue` for `test:queue:nlp`. Then `call()` separately resolves `nextStage()` to `ARTIFACT` and  opens a second queue, `test:queue:artifact` (line 86), which is the one actually used. 

The result still targets `ARTIFACT` as expected, so it's not a correctness bug, but every run now silently allocates a queue connection for the wrong stage that nothing ever reads from or writes to.
It's dead work today, and it gets confusing for anyone instrumenting queue creation once `--nextStage` no longer matches what was actually built at construction time.

Would it make sense to make that base-class queue construction lazy, or have `EnqueueFromIndexTask`
avoid triggering it since it always resolves its own queue name? This spans `PipelineTask.java`
too.

File: datashare-app/src/main/java/org/icij/datashare/tasks/EnqueueFromIndexTask.java [hunk 2]
```diff
@@ -93,4 +94,24 @@ public Long call() throws Exception {
...
+                () -> new PipelineHelper(propertiesProvider).getNextStage(Stage.ENQUEUEIDX));
+    }
+
+    /** A stage running before ENQUEUEIDX consumes file paths, not document ids, so enqueuing into its
+     *  queue would poison it rather than fail. */
+    private static Stage parseNextStage(String value) {
+        Stage stage = Stage.parse(value).orElseThrow(
+                () -> new IllegalArgumentException("unknown --nextStage value \"%s\"".formatted(value)));
+        if (Stage.comparator.compare(stage, Stage.ENQUEUEIDX) <= 0) {
+            throw new IllegalArgumentException(
```
caro3801, line 124: hint(minor): The message reads "--nextStage ENQUEUEIDX runs before ENQUEUEIDX in the pipeline" when the rejected value is ENQUEUEIDX itself, which is a bit confusing to read while debugging. 

```suggestion
            throw new IllegalArgumentException(
                    stage == Stage.ENQUEUEIDX
                            ? "--nextStage cannot target ENQUEUEIDX itself"
                            : "--nextStage %s runs before %s in the pipeline".formatted(stage, Stage.ENQUEUEIDX));
```

datashare-app/src/main/java/org/icij/datashare/tasks/EnqueueFromIndexTask.java
```diff
@@ -83,7 +83,8 @@ public Long call() throws Exception {
         logger.info("enqueuing doc ids for index {} targeting {}{} with {} scroll and size of {} : {} documents found",
                     projectName, nextStage, pipelineInfo, scrollDuration, scrollSize, totalHits);
 
-        try (DocumentQueue<String> outputQueue = factory.createQueue(getOutputQueueName(), String.class)) {
+        String outputQueueName = new PipelineHelper(propertiesProvider).getQueueNameFor(nextStage);
```
caro3801: suggestion: `nextStage()` and the `outputQueueName` line each construct their own `PipelineHelper`,
which reparses and re-sorts the `stages` property from scratch on every call. It's cheap for small
stage lists today, but could we hoist a single `PipelineHelper` (built once, like a local shared
between `nextStage()` and the queue-name lookup) to avoid the repeat parsing?

caro3801 (approved):
Nice, this will be very helpful to decouples `EnqueueFromIndexTask` from the fixed default chain, thanks! 
The CLI wiring and tests for `--nextStage` are clean and read well alongside the existing test suite. None of the findings here block the merge: the[ `parseNextStage` comment ](https://github.com/ICIJ/datashare/pull/2413#discussion_r4045206842) is worth a follow-up since `CREATENLPBATCHESFROMIDX`/`BATCHNLP` silently orphan enqueued documents today, but it's a CLI-only option aimed at people who know the pipeline, and the fix is cheap whenever it lands. The rest are smaller cleanup opportunities.

datashare-app/src/main/java/org/icij/datashare/tasks/EnqueueFromIndexTask.java
pirhoo: Done. Replaced the ordinal check with an allow list of the stages that actually have a task draining their queue: CATEGORIZE, NLP, ARTIFACT. Confirmed CreateNlpBatchesFromIndex and BatchNlpTask extend DefaultTask and read the index, so neither ever drains a stage queue. Added a test for the CREATENLPBATCHESFROMIDX rejection.

datashare-app/src/main/java/org/icij/datashare/tasks/EnqueueFromIndexTask.java
pirhoo: The allow list removes this message entirely. A rejected value now gets "--nextStage X has no task draining its queue, expected one of [CATEGORIZE, NLP, ARTIFACT]", so ENQUEUEIDX itself reads clearly too.

datashare-app/src/main/java/org/icij/datashare/tasks/EnqueueFromIndexTask.java
pirhoo: Fixed without touching PipelineTask: EnqueueFromIndexTask now overrides getOutputQueueName() to return null, so the base class builds no output queue at construction. Nothing in this class reads the inherited outputQueue, it always resolves its own in call(). New test asserts no test:queue:nlp is created when --nextStage ARTIFACT is used.

datashare-app/src/main/java/org/icij/datashare/tasks/EnqueueFromIndexTask.java
pirhoo: Done. One PipelineHelper is built at the top of call() and reused by both the next-stage lookup and the queue name.

